PR #27, opened 2024-06-06
change v4 to v1

aratz-lasa:
Bento breaks with Benthos v4 version and starts at 1, so we need to remove the v4 suffix.

> If the module is released at major version 2 or higher, the module path must end with a [major version suffix](https://go.dev/ref/mod#major-version-suffixes) like /v2. This may or may not be part of the subdirectory name. For example, the module with path [golang.org/x/repo/sub/v2](http://golang.org/x/repo/sub/v2) could be in the /sub or /sub/v2 subdirectory of the repository [golang.org/x/repo](http://golang.org/x/repo).

[https://go.dev/ref/mod](https://go.dev/ref/mod)


gregfurman (commented):
LGTM -- maybe commits should be squashed though.

PR #37, opened 2024-06-09
Sync upstream changes into bento

gregfurman:
Pulling in upstream changes. 

Changes:
- Add richer CLI customisation options
- ~Move `strip_html` to `internal/impl` and update docs (which generated weirdly and with an `unknown` param)~
- Include upstream `CHANGELOG` information into `CHANGELOG.old.md`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,8 +20,7 @@ VER_PATCH := $(shell echo $(VER_CUT) | cut -f3 -d.)
 VER_RC    := $(shell echo $(VER_PATCH) | cut -f2 -d-)
 DATE      := $(shell date +"%Y-%m-%dT%H:%M:%SZ")
 
-VER_FLAGS = -X github.com/warpstreamlabs/bento/internal/cli.Version=$(VERSION) \
-	-X github.com/warpstreamlabs/bento/internal/cli.DateBuilt=$(DATE)
+VER_FLAGS = -X main.Version=$(VERSION) -X main.DateBuilt=$(DATE)
 
 LD_FLAGS   ?= -w -s
 GO_FLAGS   ?=
--- a/internal/cli/run.go
+++ b/internal/cli/run.go
@@ -72,7 +72,7 @@ func traverseHelp(cmd *cli.Command, pieces []string) []pluginHelp {
 	pieces = append(pieces, cmd.Name)
 	var args []string
 	for _, a := range cmd.Flags {
-		args = append(args, a.Names()[0])
+		args = append(args, "--"+a.Names()[0])
 	}
 	help := []pluginHelp{{
 		Path:  strings.Join(pieces, "_"),
--- a/internal/cli/run.go
+++ b/internal/cli/run.go
@@ -1,9 +1,11 @@
 package cli
 
 import (
+	"encoding/json"
 	"fmt"
 	"os"
 	"runtime/debug"
+	"strings"
 
 	"github.com/urfave/cli/v2"
 	"gopkg.in/yaml.v3"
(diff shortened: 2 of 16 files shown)


internal/impl/html/bloblang.go
jem-davies, line 1: Don't think these changes have been made on the core engine repo.

jem-davies (commented):
will do a proper review later if not picked up - but just clocked one thing

internal/impl/html/bloblang.go
gregfurman: They are on the Apache V2 repo. Perhaps it's worth referencing where each change comes from in the commit messages when doing this upstream syncing. i.e `benthos <commit sha>: <commit description>`

internal/cli/run.go
```diff
@@ -111,68 +123,113 @@ func App(opts *common.CLIOpts) *cli.App {
...
 	}
+	flags = append(flags, runFlags(opts)...)
 
 	app := &cli.App{
-		Name:  "Bento",
-		Usage: "A stream processor for mundane tasks - https://warpstreamlabs.github.io/bento",
-		Description: `
-Either run Bento as a stream processor or choose a command:
+		Name:  opts.BinaryName,
+		Usage: opts.ExecTemplate("A stream processor for mundane tasks - {{.DocumentationURL}}"),
```
jem-davies, line 212:
```suggestion
		Usage: opts.ExecTemplate("A stream processor for mundane tasks - {{.DocumentationURL}}/about"),
```

That link is broken - need to append /about to the end.

internal/impl/html/bloblang_test.go
jem-davies, line 1: potentially take this out -

public/components/pure/extended/package.go
```diff
@@ -11,6 +11,7 @@ package extended
 import (
 	// Import pure but larger packages.
 	_ "github.com/warpstreamlabs/bento/internal/impl/awk"
+	_ "github.com/warpstreamlabs/bento/internal/impl/html"
```
jem-davies, line 14: potentially take this out

website/docs/guides/bloblang/methods.md
```diff
@@ -594,7 +594,7 @@ Attempts to remove all HTML tags from a target string.
 
 #### Parameters
 
-**`preserve`** &lt;(optional) array&gt; An optional array of element types to preserve in the output.  
+**`preserve`** &lt;(optional) unknown&gt; An optional array of element types to preserve in the output.
```
jem-davies, line 597: hmm maybe if the html changes are taken out, this might go back to an array? not sure, not checked

internal/bloblang/query/methods_test.go
jem-davies, line 1: maybe this stays in depending

internal/bloblang/query/methods_strings.go
jem-davies, line 1: maybe this stays in depending

jem-davies (commented):
Looks good, just a few minor comments: 

- not sure why html has been moved 
- A link to the docs when running ./bento -h - was dead

Makefile
```diff
@@ -20,8 +20,7 @@ VER_PATCH := $(shell echo $(VER_CUT) | cut -f3 -d.)
 VER_RC    := $(shell echo $(VER_PATCH) | cut -f2 -d-)
 DATE      := $(shell date +"%Y-%m-%dT%H:%M:%SZ")
 
-VER_FLAGS = -X github.com/warpstreamlabs/bento/internal/cli.Version=$(VERSION) \
-	-X github.com/warpstreamlabs/bento/internal/cli.DateBuilt=$(DATE)
+VER_FLAGS = -X main.Version=$(VERSION) -X main.DateBuilt=$(DATE)
```
jem-davies, line 23: this line has come from the connect repo -

gregfurman: Think this makes sense to add because these variables are now being set in the `main` package:

```go
var (
	// Version version set at compile time.
	Version string
	// DateBuilt date built set at compile time.
	DateBuilt string
	// BinaryName binary name.
	BinaryName string = "bento"
)
```

internal/cli/run.go
gregfurman: Good catch. Perhaps it would be worthwhile to redirect `/docs` to `/docs/about`

go.mod
```diff
@@ -83,7 +83,7 @@ require (
 	github.com/lib/pq v1.10.9
 	github.com/linkedin/goavro/v2 v2.12.0
 	github.com/matoous/go-nanoid/v2 v2.0.0
-	github.com/microcosm-cc/bluemonday v1.0.25
+	github.com/microcosm-cc/bluemonday v1.0.23
```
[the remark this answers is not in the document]
gregfurman: Think this just happened when I ran `go mod tidy` -- will revert.

public/components/pure/extended/package.go
gregfurman: This was an upstream change alongside the moving of html: https://github.com/redpanda-data/connect/blob/main/public/components/pure/extended/package.go#L16

website/docs/guides/bloblang/methods.md
gregfurman: I noticed there were actually a few cases of this `<unknown>` param. Probably worth checking out _why_ this is happening regardless

internal/bloblang/query/methods_test.go
gregfurman: I suppose I can remove it for now. This component has been moved out of the MIT version and into connect so there will be some more drift here that we should make note of.

PR #47, opened 2024-06-16
Add deploy-docsite workflow, Add build docsite to test workflow

jem-davies:
### YOU WILL NEED TO CHANGE A SETTING

In the github repo UI: 

 - click 'settings' 
 - left side bar, under 'code and automation' > Click 'Pages'
 - Under heading 'Build and deployment' > dropdown 'Source' > Change to 'Github Actions'


jem-davies:
Changes:
 - Added a 'job' to .github/workflows/deploy-docsite.yml workflow, that will build the doc site when a PR is created.
 - Added a new .github/workflows/deploy-docsite.yml that will redeploy the site on push to main.

I have made this [repo](https://github.com/jem-davies/jem-davies.github.io) with a change to the favicon to test -[ link to site](https://jem-davies.github.io/).

.github/workflows/deploy-docsite.yml
```diff
@@ -0,0 +1,46 @@
...
+
+# Added Stage 'Build-Docsite'
+# Changed 'Upload-Artifact' path
+
+name: Deploy Doc Site
+
+on:
+  workflow_dispatch:
+  push:
+    branches: [main]
```
gregfurman, line 11: Should this also happen on tag creation/release events?

jem-davies: I think that you could do it either: on pushes to main OR a release event, I think if you did both you would be rebuilding the same site on a release event. 

If we decide to only re-build the docsite on release then we can still trigger the build manually in the github UI for small fixes, i.e. favicon changes. 

@richardartoul what do you reckon?

.github/workflows/test.yml
```diff
@@ -56,6 +56,18 @@ jobs:
...
+
+    steps:
+    - uses: actions/checkout@v4
+    - name: Build Doc Site
+      uses: actions/setup-node@v4
+      with:
+        node-version: 20.x
+    - run: yarn --cwd ./website install && yarn --cwd ./website build
+
     # TODO: Remove this after https://github.com/golangci/golangci-lint/issues/3094 is addressed
```
jem-davies, line 71: Raising issue for this: https://github.com/warpstreamlabs/bento/issues/53

.github/workflows/test.yml
```diff
@@ -56,6 +56,18 @@ jobs:
...
     # - name: Get govulncheck
```
jem-davies, line 72:
```suggestion
    # TODO: Remove this after https://github.com/golangci/golangci-lint/issues/3094 is addressed
    # Issue: https://github.com/warpstreamlabs/bento/issues/53
    # - name: Get govulncheck
```

.github/workflows/deploy-docsite.yml
gregfurman: Yeah sure I see what you're saying. Perhaps just on main is fine since calling `go install` with `latest` will return HEAD of main in any case.

PR #66, opened 2024-07-05
Component: Add input for etcd key or prefix watcher

gregfurman:
### Changes
- Create new etcd `Watch` input, that will watch for new `PUT` or `DELETE` operations on a specific key or prefix (see [Watch API docs](https://etcd.io/docs/v3.5/learning/api/#watch-api)).
-  Add integration tests for watching a **single key** and **all keys** for changes.
- Include extensive configuration options for watch stream operations.
- Updated website with new docs.

### TODO
- Determine version this will be released in -- currently set as `1.x.x`
- Include a better description for the etcd component
- Improve the leasing capabilities
- Create an output sink for operating on etcd (i.e PUTs, DELETEs, new lease creation)


internal/impl/etcd/integration_test.go
```diff
@@ -0,0 +1,203 @@
...
+	require.NoError(t, err)
+	pool.MaxWait = time.Second * 60
+
+	resource, err := pool.RunWithOptions(&dockertest.RunOptions{
+		Repository:   "bitnami/etcd",
+		Tag:          "latest",
+		ExposedPorts: []string{"2379", "2380"},
+		Env: []string{
+			"ALLOW_NONE_AUTHENTICATION=yes",
+			"ETCD_ADVERTISE_CLIENT_URLS=http://localhost:2379",
```
jem-davies, line 40:
```suggestion
		Env: []string{
			"ALLOW_NONE_AUTHENTICATION=yes",
			"ETCD_ADVERTISE_CLIENT_URLS=http://localhost:2379",

		PortBindings: map[docker.Port][]docker.PortBinding{
			docker.Port("2379" + "/tcp"): {
				docker.PortBinding{HostIP: "", HostPort: "2379"},
			},
			docker.Port("2380" + "/tcp"): {
				docker.PortBinding{HostIP: "", HostPort: "2380"},
			},
		},
	})			
```

internal/impl/etcd/integration_test.go
```diff
@@ -0,0 +1,203 @@
+package etcd
+
+import (
+	"context"
+	"fmt"
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/ory/dockertest/v3"
```
jem-davies, line 10:
```suggestion
	"github.com/ory/dockertest/v3"
	"github.com/ory/dockertest/v3/docker"
```

internal/impl/etcd/input.go
```diff
@@ -0,0 +1,226 @@
...
+
+` + "```json" + `
+{
+  "key": "<string or []byte>",
+  "value": "<string or []byte>",
+  "type": "<'PUT' or 'DELETE'>",
+  "version": "<int64>",
+  "mod_revision": "<int64>",
+  "create_revision": "<int64>",
+  "lease": "<int64>"
```
jem-davies, line 64: nit: - think this would be a bit confusing for me, I would remove the double quotes. I just think it would imply that it was a json string of a number.

internal/impl/etcd/util.go
```diff
@@ -0,0 +1,35 @@
...
+
+		if utf8.Valid(e.Kv.Value) {
+			event["value"] = string(e.Kv.Value)
+		}
+
+		eventsMap = append(eventsMap, event)
+	}
+
+	return eventsMap
+}
```
jem-davies, line 35:
```suggestion
package etcd

import (
	"unicode/utf8"

	mvccpb "go.etcd.io/etcd/api/v3/mvccpb"
	clientv3 "go.etcd.io/etcd/client/v3"
)

func etcdEventsToMap(events []*clientv3.Event) []map[string]any {

	eventsMap := make([]map[string]any, len(events))

	for i, e := range events {
		event := map[string]any{
			"key":             e.Kv.Key,
			"value":           e.Kv.Value,
			"type":            mvccpb.Event_EventType_name[int32(e.Type)],
			"version":         e.Kv.Version,
			"mod_revision":    e.Kv.ModRevision,
			"create_revision": e.Kv.CreateRevision,
			"lease":           e.Kv.Lease,
		}

		if utf8.Valid(e.Kv.Key) {
			event["key"] = string(e.Kv.Key)
		}

		if utf8.Valid(e.Kv.Value) {
			event["value"] = string(e.Kv.Value)
		}

		eventsMap[i] = event
	}

	return eventsMap
}
```

Couple of things changed: 

- don't append because you will get [null, {...}] 
- the value of the type needs to be resolved

internal/impl/etcd/input.go
```diff
@@ -0,0 +1,226 @@
...
+	clientv3 "go.etcd.io/etcd/client/v3"
+
+	"github.com/warpstreamlabs/bento/public/service"
+)
+
+func etcdWatchFields() []*service.ConfigField {
+	return []*service.ConfigField{
+		service.NewStringField(etcdKeyField).
+			Description("The key or prefix being watched.").
+			Default(""),
```
jem-davies, line 15: not sure about the default value "". 

So in the test for the multiple keys, you would have: 

```yaml
    key: ""
    options:
      with_prefix: true
```

so if the key is default value - you have to set `options.with_prefix` true - which is not the default?

internal/impl/etcd/integration_test.go
```diff
@@ -0,0 +1,203 @@
...
+		}
+
+	}
+
+	if _, err := etcdClient.Delete(context.Background(), "/foo", clientv3.WithPrefix()); err != nil {
+		t.Error(err)
+	}
+
+	var tmpSize int
+	assert.Eventually(t, func() bool {
```
jem-davies, line 192: think it would be good to have another test that checks the value returned is expected

jem-davies (changes requested):
some changes required I think

internal/impl/etcd/util.go
jem-davies: also not sure about the need to put them into an array - in what cases would you get more than 1 event in an array? (not familar with etcd)

gregfurman: >  - don't append because you will get [null, {...}]

Oh wow that was a massive oversight on my end. Nice catch.

> - the value of the type needs to be resolved

The underlying event type has a `String()` method defined so I'm not sure the extra step is needed:
```golang
func (x Event_EventType) String() string {
	return proto.EnumName(Event_EventType_name, int32(x))
}
```

> also not sure about the need to put them into an array - in what cases would you get more than 1 event in an array? (not familar with etcd)

If I were watching a prefix `/foo` which contained `/baz` `/buz` and `/foobar`, and I deleted `/foo` (thereby also removing `/baz` `/buz` and `/foobar`) I believe this would register as multiple delete events for each key. Let me extend the integration tests to see if this actually is the case.

internal/impl/etcd/input.go
gregfurman: Probably a better idea to make this a required value as opposed to allowing this default.

internal/impl/etcd/integration_test.go
gregfurman: I'm going to re-work how I do these integration tests since they can only currently be run sequentially.

danriedl:
hey just checking in, 
to let you know i created a PR for etcd cache/input in the `old` benthos-dev/benthos project in april:
https://github.com/redpanda-data/connect/pull/2555

gregfurman:
Hey @danriedl! Wow that looks awesome. Since you have context on etcd, Go, and delcarative pipelines, would love it if you gave these changes a look? Else, if you want to try combine our solutions I think that'd be awesome.

Happy to chat on [Discord](https://console.warpstream.com/socials/discord) or [Slack](https://console.warpstream.com/socials/slack) too :)

jem-davies (approved):

PR #67, opened 2024-07-08
docs: Add cookbook for Kafka topic mirroring

gregfurman:
We have recieved a couple of questions relating to topic mirroring. Hopefully this guide can better help users get up-and-running with this faster.

### Changes
- Create cookbook showing how to mirror Kafka topics while preserving original partition mapping
- Added in some processing steps to better illustrate dynamic output routing with string interpolation and metadata

### DIFF
--- a/website/cookbooks/kafka_topic_mirroring.md
+++ b/website/cookbooks/kafka_topic_mirroring.md
@@ -1,0 +1,133 @@
+---
+slug: kafka-mirroring
+title: Kafka Topic Mirroring
+description: Learn how to mirror Kafka topics while preserving partition mapping.
+---
+
+Kafka-flavoured Bento (カフカ風弁当; Kafuka-fū Bentō), a favourite here at WarpStream Labs, is a quick-and-easy recipe you can whip up in minutes. This cookbook will illustrate how to use Bento for consuming and publishing events to Kafka, with the goal of **mirroring Kafka topics while preserving partition mappings**.
+
+For example, the diagram below shows *partition preservation* of some process where `bento` consumes an event from `Partition 2` of `Topic A` and maps it to `Partition 2` of `Topic B`:
+```
+  Topic A                                                                  Topic B                       
++-----------------------------+                                          +-----------------------------+
+|                             |                                          |                             |
+|    +--------------------+   |                                          |   +--------------------+    |
+| P1 |                    |   |                                          |   |                    | P1 |
+|    +--------------------+   |                                          |   +--------------------+    |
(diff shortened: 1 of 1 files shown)


website/cookbooks/kafka_topic_mirroring.md
```diff
@@ -0,0 +1,176 @@
...
+  "metadata": {
+    "output_topic": "<'woody' | 'tinny' | 'pvc'>",
+    ...
+  }
+}
+```
+
+## Consuming Events
+
+Let's jump in, and start consuming data using the [`kafka input`][kafka.input] component. Here, we will reading in all new events from the `words` topic.
```
richardartoul, line 56: reading -> read

website/cookbooks/kafka_topic_mirroring.md
```diff
@@ -0,0 +1,176 @@
...
+
+```yaml
+input:
+  kafka:
+    consumer_group: bento_bridge_consumer
+    addresses: [ TODO ]
```
richardartoul, line 62: Probably replace with like $BOOTSTRAP_BROKER instead of TODO? or is that the convention in these docs?

website/cookbooks/kafka_topic_mirroring.md
```diff
@@ -0,0 +1,176 @@
...
+          meta output_topic = "woody"
+        } else if ["newspaper", "antelope", "recidivist"].contains(this.word) {
+          meta output_topic = "tinny"
+        } else if this.word == "leap" {
+          meta output_topic = "pvc"
+        } else {
+          root = deleted()
+        }
+
+output:
```
richardartoul, line 161: Can you add something at the end that is an example of input --> output, without the processor, and with the output topic configured to be the input topic using the templating? Also, does the kafka input support regexes? If so, can you add one more example of that?

Finally, can you use franz_kafka everywhere instead of kafka?

richardartoul (approved):

website/cookbooks/kafka_topic_mirroring.md
gregfurman: Wow that is embarassing. Will change.

website/cookbooks/kafka_topic_mirroring.md
gregfurman: Docs convention is `TODO`. See https://warpstreamlabs.github.io/bento/cookbooks/joining-streams#caching-articles

richardartoul (approved):

Issue #70, opened 2024-07-10
Connection status incorrect postgre sql_insert


emmanuelbertho:
Hello,

I noticed a bug when using sql_insert with postgres as an output in streams mode.
When I put fake connection parameters for postgres the /stats endpoint for my stream displays that the connection is up tough it is not.

Logs : 

`2024-07-10 09:49:57 {"@service":"bento","label":"","level":"warning","msg":"Failed to execute init_statement: dial tcp: lookup myhost on 127.0.0.11:53: no such host","path":"root.output","stream":"Emmanuel"}
2024-07-10 09:49:57 {"@service":"bento","label":"","level":"info","msg":"Output type sql_insert is now active","path":"root.output","stream":"Emmanuel"}`

Stream : 

`{
  "active": true,
  "uptime": 6.814057709,
  "uptime_str": "6.814057788s",
  "config": {
    "buffer": {
      "none": { }
    },
    "input": {
      "redis_streams": {
        "body_key": "data",
        "streams": [
          "value-change-Emmanuel"
        ],
        "url": "redis://redis:6379"
      }
    },
    "output": {
      "sql_insert": {
        "args_mapping": "[ this.timestamp, this.data.string() ]",
        "columns": [
          "timestamp",
          "data"
        ],
        "driver": "postgres",
        "dsn": "postgres://fakeuser:fakepasswd@myhost:1234/?database=fakedb&sslmode=disable",
        "init_statement": "CREATE TABLE IF NOT EXISTS aa (\n   timestamp TIMESTAMPTZ NOT NULL,\n   data jsonb NOT NULL);\nSELECT create_hypertable('aa', by_range('timestamp'));",
        "table": "aa"
      }
    },
    "pipeline": {
      "processors": [
        {
          "mapping": "\r\n                root.data = this\r\n                root.data.\"\" = deleted()\r\n                root.timestamp = now()"
        }
      ]
    }
  }
}`

Stats : 

`{
  "input_connection_failed{label=\"\",path=\"root.input\",stream=\"Emmanuel\"}": 0,
  "input_connection_lost{label=\"\",path=\"root.input\",stream=\"Emmanuel\"}": 0,
  "input_connection_up{label=\"\",path=\"root.input\",stream=\"Emmanuel\"}": 1,
  "input_latency_ns{label=\"\",path=\"root.input\",stream=\"Emmanuel\"}": {
    "p50": 0,
    "p90": 0,
    "p99": 0
  },
[...]

PR #71, opened 2024-07-12
sql: Add optional verification ping on database client initialisation

gregfurman:
### Summary

- Add a flag that, when true, pings a SQL database. If the ping fails the component is set to an error state.
- From #70 

### Changes

- Extended `sqlOpenWithReworks` to include a `shouldPing` boolean which will run a `db.PingContext` if true. 
- Small refactor of `sqlOpenWithReworks`, placing the `rework` functionality into a seperate `reworkDSN` function.
- Added new flag `init_verify_conn` to the SQL component's config and included this in the `connSettings` struct.
- Updated docs.

### DIFF
--- a/internal/impl/sql/conn_fields.go
+++ b/internal/impl/sql/conn_fields.go
@@ -78,6 +78,12 @@ CREATE TABLE IF NOT EXISTS some_table (
 			Optional().
 			Advanced().
 			Version("1.0.0"),
+		service.NewBoolField("init_verify_conn").
+			Description("Whether to verify the database connection on startup by performing a simple ping, by default this is disabled.").
+			Default(false).
+			Optional().
+			Advanced().
+			Version("1.2.0"),
 		service.NewDurationField("conn_max_idle_time").
 			Description("An optional maximum amount of time a connection may be idle. Expired connections may be closed lazily before reuse. If `value <= 0`, connections are not closed due to a connections idle time.").
 			Optional().
@@ -125,6 +131,7 @@ type connSettings struct {
 	initOnce           sync.Once
 	initFileStatements [][2]string // (path,statement)
 	initStatement      string
+	initVerifyConn     bool
 }
 
 func (c *connSettings) apply(ctx context.Context, db *sql.DB, log *service.Logger) {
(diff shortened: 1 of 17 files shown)


internal/impl/sql/conn_fields.go
```diff
@@ -78,6 +78,12 @@ CREATE TABLE IF NOT EXISTS some_table (
 			Optional().
 			Advanced().
 			Version("1.0.0"),
+		service.NewBoolField("init_verify_conn").
+			Description("Whether to verify the database connection on startup by performing a simple ping. Disabling this allows for certain components to lazily connect.").
```
jem-davies, line 82:
```suggestion
			Description("Whether to verify the database connection on startup by performing a simple ping, by default this is disabled.").
```

nit: Think that this is clearer that the default is set to false?

internal/impl/sql/conn_fields.go
```diff
@@ -78,6 +78,12 @@ CREATE TABLE IF NOT EXISTS some_table (
...
+			Default(false).
+			Optional().
+			Advanced().
+			Version("1.0.3"),
```
jem-davies, line 86: Will this version number need updating now?

gregfurman: Good catch! Will change

internal/impl/sql/conn_fields.go
gregfurman: Yeah that’s fair. You think it’s worth mentioning lazy loading?

jem-davies: not sure - if just me or I don't know exactly what is meant lazy loading?

jem-davies: just think if it is worded so clearer to explain what the default behaviour is

internal/impl/sql/conn_fields.go
gregfurman: Changed to `1.2.0` 🚀

internal/impl/sql/conn_fields.go
gregfurman: Added your suggestion 👍

PR #97, opened 2024-08-17
remove references to depreciated function  meta() in docs

jem-davies:


gregfurman (approved):
Nice catch! That this is deprecated isn't apparent on the docs though: https://warpstreamlabs.github.io/bento/docs/guides/bloblang/functions/#meta

Maybe we should consider fixing the deprecated labelling too?

Also, I found quite a few docs where `meta` is still referenced (a few of them are ref'd below):

```
http_server 
sql_raw 
system_window 
aws_dynamodb 
aws_s3 
azure_blob_storage 
cache 
gcp_cloud_storage 
nats 
nats_jetstream 
redis_hash 
snowflake_put 
sql 
sql_insert 
sql_raw 
aws_dynamodb_partiql 
azure_cosmosdb 
cached 
dedupe 
group_by 
group_by_value 
javascript 
log 
metric 
nats_request_reply 
parquet_decode 
redis 
redis_script 
sql 
sql_insert 
sql_raw 
sql_select 
```

Should we do this all at once or you wanna make a follow-up PR?

jem-davies:
Think to do a follow up PR because I did originally see them but needed to check because some of them have type casting because the old function only worked with Strings as the metadata value.

PR #118, opened 2024-09-17
add output cypher component

jem-davies:


cmd/bento/config.yaml
gregfurman, line 1: Think this should be removed

internal/impl/cypher/output_cypher.go
```diff
@@ -0,0 +1,155 @@
...
+		Fields(
+			service.NewStringField("database").
+				Description("The name of the database to connect to.").
+				Example("neo4j"),
+			service.NewStringField("uri").
+				Description("The URI of the database.").
+				Example("bolt://localhost:7687"),
+			service.NewBoolField("noAuth").
+				Description("No Authentication currently implemented, defaults to true.").
+				Default(true),
```
gregfurman, line 21: Maybe this should be removed for now since it doesn't look like it's being used?

internal/impl/cypher/output_cypher.go
```diff
@@ -0,0 +1,155 @@
...
+
+	values := make(map[string]any)
+
+	for _, msgPart := range batch {
+
+		for k, v := range cyp.values {
+			values[k] = v.String(msgPart)
+		}
+
+		_, err := neo4j.ExecuteQuery(ctx, cyp.driver,
```
gregfurman, line 134: Should this output be a `query` or an `INSERT` statement of some kind? I'd imagine a`SELECT` (or neo4j equivalent) shouldn't be processed as output

internal/impl/cypher/output_cypher.go
```diff
@@ -0,0 +1,155 @@
...
+	}
+	uri, err := conf.FieldString("uri")
+	if err != nil {
+		return nil, err
+	}
+	noAuth, err := conf.FieldBool("noAuth")
+	if err != nil {
+		return nil, err
+	}
+	query, err := conf.FieldString("query")
```
gregfurman, line 100: Just a thought but I wonder if this should be sanitised in some way to prevent some query injection

gregfurman (changes requested):

cmd/bento/config.yaml
jem-davies: taken it out

internal/impl/cypher/output_cypher.go
jem-davies: added a field for basic_auth so it is being used now

internal/impl/cypher/output_cypher.go
jem-davies: I think there is a few different verbs that can create data - so with this approach, yes you could add the equivalent of a select but I think that is more on the user at that point.

internal/impl/cypher/output_cypher.go
jem-davies: I think that should be ok - as is.

gregfurman (changes requested):
Looking good. Have left some comments and nits. Haven't validated the tests on my local though but will do so soon.

internal/impl/cypher/output_cypher_test.go
```diff
@@ -0,0 +1,270 @@
...
+	resource, err := pool.RunWithOptions(&dockertest.RunOptions{
+		Repository: "neo4j",
+		Tag:        "latest",
+		Env:        env,
+	})
+	if err != nil {
+		return "", err
+	}
+
+	neo4jDockerAddress := fmt.Sprintf("bolt://localhost:%s", resource.GetPort("7687/tcp"))
```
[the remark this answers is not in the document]
jem-davies: refactored the tests so they are similar to etcd 👍  -- also I think that the docker engine will only assign to a free port anyway.

internal/impl/cypher/output_cypher.go
```diff
@@ -0,0 +1,227 @@
...
+	if !noAuth {
+		basicAuthMap, _ := conf.FieldStringMap("basic_auth")
+		basicAuth := CypherBasicAuth{user: basicAuthMap["user"], password: basicAuthMap["password"], realm: basicAuthMap["realm"]}
+		return &CypherOutput{database: database, url: url, noAuth: noAuth, basicAuth: basicAuth, query: query, values: values}, nil
+	}
+
+	return &CypherOutput{database: database, url: url, noAuth: noAuth, query: query, values: values}, nil
+}
+
+func validateQueryAndValues(query string, values map[string]*service.InterpolatedString) error {
```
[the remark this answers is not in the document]
jem-davies: There are TestCypherOutputMissingValue and TestCypherOutputMissingParam in output_cypher_test.go - have you seen them / they not what we will want?

jem-davies:
@gregfurman thanks for review - updated the code / responded to your comment...

internal/impl/cypher/output_cypher.go
```diff
@@ -0,0 +1,263 @@
...
+	// Basic Auth
+	cypherUser     = "user"
+	cypherPassword = "password"
+	cypherRealm    = "realm"
+)
+
+var cypherOutputDescription string = `
+## Executes a Cypher Query
+
+The ` + "`" + `query` + "`" + ` field is expected to be a valid cypher query with 0 or more parameters with the $ syntax:
```
gregfurman, line 32:
```suggestion
The ` + "`" + `query` + "`" + ` field is expected to be a valid cypher query with 0 or more parameters with the ` + "`" + `$` + "`" + `syntax:
```

internal/impl/cypher/output_cypher.go
```diff
@@ -0,0 +1,263 @@
...
+			`
+output:
+  cypher:
+    database: neo4j
+    uri: bolt://localhost:7687
+    basic_auth:
+      user: neo4j
+      password: password
+    query: |
+      CREATE (p:Person {name: $name}) RETURN p
```
gregfurman, line 91: Should we not just allow bloblang mappings directly in here instead of that extra values field?

internal/impl/cypher/output_cypher_test.go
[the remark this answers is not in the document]
gregfurman: I get you. However, if we ran 2 tests concurrently and tried to call `resource.GetPort("7687/tcp")` when both containers were running (each internally assigned the same port) then we would've run into a similar issue as with etcd. 

I see you addressed this though 😄

internal/impl/cypher/output_cypher_test.go
```diff
@@ -0,0 +1,259 @@
...
+		map[string]any{},
+		neo4j.EagerResultTransformer,
+		neo4j.ExecuteQueryWithDatabase("neo4j"),
+	)
+
+	var listOfNamesFromDB []string
+
+	for _, record := range results.Records {
+		dictionary := record.AsMap()
+		x := dictionary["n"].(dbtype.Node)
```
gregfurman, line 118: nit: Maybe don't use `x` lol

internal/impl/cypher/output_cypher_test.go
```diff
@@ -0,0 +1,259 @@
...
+    name: ${! json("name") }
+  batching:
+    count: 0
+  `, localURL)
+
+	s, err := createCypherOutputFromYaml(template)
+	require.NoError(t, err)
+
+	// send the messages
+	sendMessages(&testing.T{}, s)
```
gregfurman, line 160: You need to pass the `testing.T` fixture to these sub-tests, otherwise we'll have issues with test context
```suggestion
	sendMessages(t, s)
```

internal/impl/cypher/output_cypher_test.go
```diff
@@ -0,0 +1,259 @@
...
```
gregfurman, line 190:
```suggestion
	sendMessages(t, s)
```

internal/impl/cypher/output_cypher_test.go
```diff
@@ -0,0 +1,259 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/warpstreamlabs/bento/internal/component/output"
+	"github.com/warpstreamlabs/bento/internal/component/testutil"
+	"github.com/warpstreamlabs/bento/internal/manager/mock"
+	"github.com/warpstreamlabs/bento/internal/message"
+	"github.com/warpstreamlabs/bento/public/service/integration"
+)
+
+func setupNeo4j(env []string) (string, error) {
```
gregfurman, line 21: If you add a `t.Testing` fixture up here, you can ensure that the resource is purged when the test is done, as opposed to relying on that `resource.Expire()` call.

```
	require.NoError(t, err)
	t.Cleanup(func() {
		assert.NoError(t, pool.Purge(resource))
	})
```

Could also do assertions up here with `require.NoError(t, err)` and just return a `string` instead of worrying about returning an error type and checking if its not nil.

internal/impl/cypher/output_cypher.go
```diff
@@ -0,0 +1,263 @@
...
+	session   neo4j.SessionWithContext
+}
+
+type CypherBasicAuth struct {
+	user     string
+	password string
+	realm    string
+}
+
+func NewCypherOutputFromConfig(conf *service.ParsedConfig, mgr *service.Resources) (*CypherOutput, error) {
```
gregfurman, line 153: Assuming we're going to write an `insert` as well, we should consider adding `metadata` when streaming in input

internal/impl/cypher/output_cypher_test.go
```diff
@@ -0,0 +1,259 @@
...
+cypher:
+  database: neo4j
+  uri: %s
+  no_auth: true
+  query: |
+    CREATE (p:Person {name: $name}) RETURN p
+  values:
+    name: ${! json("name") }
+  batching:
+    count: 0
```
gregfurman, line 153: Any chance we can just have a single test with batching?

gregfurman (commented):
This is looking great, dude! Just some comments regarding tests (setup + teardown, properly passing around the `testing.T` and adding a batching test).

Also ran your tests and my local and all looks good 👍

internal/impl/cypher/output_cypher.go
jem-davies: Think it's like this because this is how the func [ExecuteQuery](https://pkg.go.dev/github.com/neo4j/neo4j-go-driver/v5/neo4j#ExecuteQuery) is in the neo4j package + it is similar to how the sql_insert is done with the args_mapping field.

internal/impl/cypher/output_cypher.go
jem-davies: unsure what you mean? (did you mean `input` instead of `insert`) 

I think yeah we def. could add some metadata fields as done with a lot of the components, think that some follow on PRs will come after this for an input / more functionality, maybe we add that in later on?

internal/impl/cypher/output_cypher_test.go
jem-davies: added in

internal/impl/cypher/output_cypher_test.go
jem-davies: updated them

jem-davies:
@gregfurman addressed the last batch of feedback, thanks

internal/impl/cypher/output_cypher.go
gregfurman: I meant like assuming we are going to write a `cypher_insert` similar to `sql_insert` since this component is equivalent to `sql_raw`. But yeah let's get this bad-boy in.

gregfurman (approved):
LGTM! Sorry for taking so long on this one 😄

PR #123, opened 2024-09-30
Add spanner support as an SQL output

rvilim:
This adds Google Cloud Spanner (https://cloud.google.com/spanner?hl=en) support for sql_raw and sql_insert. 

Since spanner conforms to the standards sql library this is mostly just an import. Quoting in spanner sql is slightly different than normal sql (it doesn't like double quotes) so some conditional logic had to be added, particularly in the integration tests.

The integration tests are done in the same fashion as all the other ones, with a docker container that runs the spanner emulator (dockerfile + startup script included here)

## Note
This is a resubmission of #122 which I am closing. #122 removes some uneeded db.Close() calls, we decided that's a bit scary. It was honestly eaiser just to make a new branch opting in to the changes I need rather than undoing all the changes I don't need.

Here we re-instantiate the DB object in the tests a bunch because spanner will not let you reuse one once you've closed it.


internal/impl/sql/integration_test.go
```diff
@@ -718,6 +758,78 @@ func TestIntegrationPostgres(t *testing.T) {
...
+  baz string(50) not null,
+		) primary key (foo)`, name))
+		return name, err
+	}
+
+	require.NoError(t, pool.Retry(func() error {
+		var db *sql.DB
+		db, err = sql.Open("spanner", dsn)
+		if err != nil {
+			fmt.Println(`open error`, err)
```
jem-davies, line 815:
```suggestion
			t.Logf(`open error: %s`, err)
```

internal/impl/sql/integration_test.go
```diff
@@ -718,6 +758,78 @@ func TestIntegrationPostgres(t *testing.T) {
...
+			return err
+		}
+		if err = db.Ping(); err != nil {
+			fmt.Println(`ping error`, err)
```
jem-davies, line 819:
```suggestion
			t.Logf(`ping error: %s`, err)
```

internal/impl/sql/integration_test.go
```diff
@@ -718,6 +758,78 @@ func TestIntegrationPostgres(t *testing.T) {
...
+			db.Close()
+			return err
+		}
+		if _, err := createTable("footable"); err != nil {
+			fmt.Println(`create table error`, err)
```
jem-davies, line 824:
```suggestion
			t.Logf(`create table error: %s`, err)
```

internal/impl/sql/integration_test.go
```diff
@@ -718,6 +758,78 @@ func TestIntegrationPostgres(t *testing.T) {
...
 }
 
+func TestIntegrationSpanner(t *testing.T) {
+	integration.CheckSkip(t)
+
+	pool, err := dockertest.NewPool("")
+	if err != nil {
+		t.Skipf("Could not connect to docker: %s", err)
+	}
+	pool.MaxWait = 10 * time.Second
```
jem-davies, line 768:
```suggestion
	pool.MaxWait = 3 * time.Minute
```

just so it's the same as the other tests in this file

internal/impl/sql/integration_test.go
```diff
@@ -18,6 +18,8 @@ import (
 	"github.com/warpstreamlabs/bento/public/service"
 	"github.com/warpstreamlabs/bento/public/service/integration"
 
+	_ "github.com/googleapis/go-sql-spanner"
+
```
jem-davies, line 22:
```suggestion

```

jem-davies: not needed?

internal/impl/sql/integration_test.go
```diff
@@ -718,6 +758,78 @@ func TestIntegrationPostgres(t *testing.T) {
...
+	// This needs to be set so that the Spanner connector will pick up that we are using the emulator.
+	t.Setenv("SPANNER_EMULATOR_HOST", emulatorHost)
+
+	project := "test-project"
+	instance := "test-instance"
+	database := "test-database"
+
+	dsn := fmt.Sprintf("projects/%s/instances/%s/databases/%s", project, instance, database)
+
+	t.Cleanup(func() {
```
jem-davies, line 793: should we add db.Close() to this cleanup function?

jem-davies (commented):
just a few small things - looks good 👍  thanks a lot for the contribution

internal/impl/sql/integration_test.go
```diff
@@ -718,6 +758,78 @@ func TestIntegrationPostgres(t *testing.T) {
...
+
+	resource, err := pool.BuildAndRunWithBuildOptions(&dockertest.BuildOptions{
+		ContextDir: "./resources",
+		Dockerfile: "Dockerfile_spanner",
+	}, &dockertest.RunOptions{
+		Name:         "spannertest",
+		ExposedPorts: []string{"9010/tcp", "9020/tcp"},
+	})
+	if err != nil {
+		fmt.Printf("Could not start resource: %s", err)
```
jem-davies, line 778:
```suggestion
		t.Logf("Could not start resource: %s", err)
```

internal/impl/sql/integration_test.go
rvilim: oh oops! thanks

internal/impl/sql/integration_test.go
rvilim: I'm not sure we can? There is no db object in scope there. Also, at that point the entire docker container is getting killed so to a certain extent it doesn't matter.

rvilim: There isn't a db object in scope in `Cleanup` atm. Or do you mean adding


```go
var db *sql.DB
db, err = sql.Open("spanner", dsn)
db.Close()
```

at the start of cleanup? Happy to do that if you like.

rvilim:
> just a few small things - looks good 👍 thanks a lot for the contribution

Thanks for catching my silliness! I had one q about what you were looking for in the cleanup function. Happy to go either way, I'm not sure it matters a ton since we are killing the docker container and everything is going out of scope anyways.

internal/impl/sql/integration_test.go
jem-davies: yeah would need to do that and alter the line below to := ... Can see how it's done in the other tests.

PR #125, opened 2024-10-03
Fix "zombie" http output websocket connections 

buger:
At the moment http_server websocket implementation just writes to webscocket, and do not have a way to detect client disconnect proactively. More over, if client disconnects, and it fails write to it, it does not stop the handler, and continues to be inside the endless for loop, untill stream is stopped `for !h.shutSig.IsSoftStopSignalled() {`.

It cause issues when you have multiple clients connected/disconnected, because essentially it fanout using round robin to them, but if you have closed "zombie" connections, they will also consume events.

It fixes the issue by using websocket native mechanisms like "ping/pong" methods. Also it adds ephemeral "Read", just to immidiately detect the client shutdown.

### DIFF
--- a/internal/impl/io/output_http_server.go
+++ b/internal/impl/io/output_http_server.go
@@ -493,7 +493,7 @@ func (h *httpServerOutput) wsHandler(w http.ResponseWriter, r *http.Request) {
 			_ = ts.Ack(ctx, nil)
 		case <-ticker.C:
 			// Send a ping message to the client
-			// nolint:errcheck // this function does not actually return an error
+			//nolint:errcheck // this function does not actually return an error
 			ws.SetWriteDeadline(time.Now().Add(h.conf.WriteWait))
 			if err := ws.WriteMessage(websocket.PingMessage, nil); err != nil {
 				h.log.Warn("WebSocket ping error: %v", err)
--- a/internal/impl/io/output_http_server.go
+++ b/internal/impl/io/output_http_server.go
@@ -478,7 +478,6 @@ func (h *httpServerOutput) wsHandler(w http.ResponseWriter, r *http.Request) {
 			// Write messages to the client
 			var writeErr error
 			for _, msg := range message.GetAllBytes(ts.Payload) {
-				//nolint:errcheck // this function does not actually return an error
 				_ = ws.SetWriteDeadline(time.Now().Add(h.conf.WriteWait))
 				if writeErr = ws.WriteMessage(websocket.BinaryMessage, msg); writeErr != nil {
 					break
--- a/internal/impl/io/output_http_server.go
+++ b/internal/impl/io/output_http_server.go
@@ -416,10 +416,12 @@ func (h *httpServerOutput) wsHandler(w http.ResponseWriter, r *http.Request) {
     )
 
     ws.SetReadLimit(512)
-    ws.SetReadDeadline(time.Now().Add(pongWait))
+    if err := ws.SetReadDeadline(time.Now().Add(pongWait)); err != nil {
+        h.log.Warn("Failed to set read deadline: %v", err)
+        return
+    }
     ws.SetPongHandler(func(string) error {
-        ws.SetReadDeadline(time.Now().Add(pongWait))
-        return nil
+        return ws.SetReadDeadline(time.Now().Add(pongWait))
     })
 
     // Start a goroutine to read messages (to process control frames)
@@ -455,7 +457,10 @@ func (h *httpServerOutput) wsHandler(w http.ResponseWriter, r *http.Request) {
             // Write messages to the client
             var writeErr error
             for _, msg := range message.GetAllBytes(ts.Payload) {
-                ws.SetWriteDeadline(time.Now().Add(writeWait))
+                if err := ws.SetWriteDeadline(time.Now().Add(writeWait)); err != nil {
+                    writeErr = err
+                    break
+                }
                 if writeErr = ws.WriteMessage(websocket.BinaryMessage, msg); writeErr != nil {
                     break
                 }
@@ -470,7 +475,10 @@ func (h *httpServerOutput) wsHandler(w http.ResponseWriter, r *http.Request) {
             _ = ts.Ack(ctx, nil)
         case <-ticker.C:
             // Send a ping message to the client
-            ws.SetWriteDeadline(time.Now().Add(writeWait))
+            if err := ws.SetWriteDeadline(time.Now().Add(writeWait)); err != nil {
+                h.log.Warn("Failed to set write deadline for ping: %v", err)
+                return
+            }
             if err := ws.WriteMessage(websocket.PingMessage, nil); err != nil {
                 h.log.Warn("WebSocket ping error: %v", err)
                 return
--- a/internal/impl/io/output_http_server.go
+++ b/internal/impl/io/output_http_server.go
@@ -389,56 +389,103 @@ func (h *httpServerOutput) streamHandler(w http.ResponseWriter, r *http.Request)
 }
 
 func (h *httpServerOutput) wsHandler(w http.ResponseWriter, r *http.Request) {
-	var err error
-	defer func() {
-		if err != nil {
-			http.Error(w, "Bad request", http.StatusBadRequest)
-			h.log.Warn("Websocket request failed: %v\n", err)
-			return
-		}
-	}()
+    var err error
+    defer func() {
+        if err != nil {
+            http.Error(w, "Bad request", http.StatusBadRequest)
+            h.log.Warn("WebSocket request failed: %v", err)
+            return
+        }
+    }()
 
-	upgrader := websocket.Upgrader{}
+    upgrader := websocket.Upgrader{}
 
-	var ws *websocket.Conn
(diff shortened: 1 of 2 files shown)


internal/impl/io/output_http_server.go
```diff
@@ -389,57 +389,104 @@ func (h *httpServerOutput) streamHandler(w http.ResponseWriter, r *http.Request)
...
+        h.log.Warn("WebSocket upgrade failed: %v", err)
+        return
+    }
+    defer ws.Close()
+
+    // Set up ping/pong handlers and deadlines
+    const (
+        writeWait  = 10 * time.Second
+        pongWait   = 60 * time.Second
+        pingPeriod = (pongWait * 9) / 10
```
buger, line 415: Those who would read it may find it strange, so here is why:

By setting `pingPeriod` to 90% of `pongWait` (i.e., `(pongWait * 9) / 10`), we ensure that:

1.  **The server sends a ping before the `pongWait` deadline**:

    -   If `pongWait` is 60 seconds, `pingPeriod` becomes 54 seconds.
    -   This means the server sends a ping every 54 seconds.
2.  **There's a buffer period for the client to respond**:

    -   The client has 6 seconds (the remaining 10% of `pongWait`) to respond with a pong before the read deadline (`pongWait`) expires.
3.  **Avoiding premature timeouts**:

    -   If the `pingPeriod` were equal to `pongWait`, the read deadline might expire before the client has a chance to respond to the ping.
    -   By setting `pingPeriod` slightly less than `pongWait`, we avoid this race condition.

internal/impl/io/output_http_server.go
```diff
@@ -389,57 +389,104 @@ func (h *httpServerOutput) streamHandler(w http.ResponseWriter, r *http.Request)
...
+    )
+
+    ws.SetReadLimit(512)
```
buger, line 418: On the question, on why we need both ping/pong and Read goroutine:

-   **Necessity of Ping/Pong with a Read Loop**:

    -   **Idle Clients**: The read loop alone cannot detect if an idle client has disconnected unexpectedly.
    -   **Timely Detection**: Ping/pong can detect unresponsive clients within a predictable timeframe (`pongWait`), regardless of message traffic.
    -   **Protocol Compliance**: Utilizing ping/pong aligns with WebSocket protocol standards and best practices.-   **Ping/Pong vs. Read Loop**:

    -   **Complementary Mechanisms**: The read loop and ping/pong serve different purposes and complement each other.
    -   **Comprehensive Health Check**: Together, they ensure active communication channels are maintained and resources are freed promptly when clients disconnect.

buger:
^ fixed linter

internal/impl/io/output_http_server.go
```diff
@@ -389,57 +389,112 @@ func (h *httpServerOutput) streamHandler(w http.ResponseWriter, r *http.Request)
...
+    ws.SetPongHandler(func(string) error {
+        return ws.SetReadDeadline(time.Now().Add(pongWait))
+    })
+
+    // Start a goroutine to read messages (to process control frames)
+    done := make(chan struct{})
+    go func() {
+        defer close(done)
+        for {
+            _, _, err := ws.ReadMessage()
```
gregfurman, line 432: Not 100% clued up on websockets so excuse the ignorance but will doing a `ws.ReadMessage()` clear all bytes in the buffer after reading? Else, does a `WriteMessage` fan-out to all consumers on the connection?

I'm concerned about a race condition where the payload we send using the below `WriteMessage` ends up being consumed in this goroutine instead of to the client where intending to send this.

internal/impl/io/output_http_server.go
gregfurman: This all makes sense! A small nit would be we should consider adding these as configuration variables with the `advanced()` flag. Think these are sensible defaults in any-case.

gregfurman (commented):
This is great! Thanks so much for the contribution and fix 🍱😄 Just a couple of questions/suggestions.

Mostly concerned that the `ReadMessage` in that go-routine could cause a race-condition when sending a `Ping` to the same socket we write the message bytes to.

Otherwise, run `make fmt && make lint` on your local to sort out these linting issues 🙂

internal/impl/io/output_http_server.go
buger: There's no risk of a race condition like the one you're describing. 

In the context of WebSockets, the connection between the client and the server is full-duplex, meaning both ends can send and receive messages independently of each other. This feature allows for simultaneous communication without interference.

- **Separate Channels**: The `ws.ReadMessage()` method reads messages that the **client** has sent to the server. The `ws.WriteMessage()` method sends messages from the server to the **client**. They operate on separate channels within the WebSocket protocol and don't interfere with each other.

- **No Buffer Overlap**: When you call `ws.WriteMessage()`, it doesn't add messages to a buffer that `ws.ReadMessage()` would read from. Similarly, `ws.ReadMessage()` doesn't consume messages that the server has sent—only those sent by the client.

- **Required Read Loop**: Implementing a read loop as done, is recommended in WebSocket servers. It ensures that control frames (like pings and pongs) are properly handled, which is essential for maintaining the connection. Not reading from the WebSocket can lead to the connection being closed due to unhandled control frames or buffer overflows.

internal/impl/io/output_http_server.go
```diff
@@ -389,57 +389,112 @@ func (h *httpServerOutput) streamHandler(w http.ResponseWriter, r *http.Request)
...
+    )
+
+    ws.SetReadLimit(512)
+    if err := ws.SetReadDeadline(time.Now().Add(pongWait)); err != nil {
+        h.log.Warn("Failed to set read deadline: %v", err)
+        return
+    }
+    ws.SetPongHandler(func(string) error {
+        return ws.SetReadDeadline(time.Now().Add(pongWait))
+    })
```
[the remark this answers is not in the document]
buger: Yes, both calls are necessary:
1. Initial `SetReadDeadline` sets the first timeout for the connection to prevent indefinite blocking
2. `PongHandler`'s `SetReadDeadline` extends the timeout each time a pong is received, keeping the connection alive only while the client is responsive

Think of it like resetting a timer - you need both the initial timer setup and the mechanism to reset it when activity is detected.

buger:
@gregfurman @jem-davies just a gentle bump in case if you missed the last message :)

internal/impl/io/output_http_server.go
```diff
@@ -393,50 +420,94 @@ func (h *httpServerOutput) wsHandler(w http.ResponseWriter, r *http.Request) {
...
 			if !open {
+				// If the transactions channel is closed, trigger server shutdown
 				go h.TriggerCloseNow()
 				return
 			}
-		case <-r.Context().Done():
+			// Write messages to the client
+			var writeErr error
+			for _, msg := range message.GetAllBytes(ts.Payload) {
+				//nolint:errcheck // this function does not actually return an error
```
jem-davies, line 481:
```suggestion
```

internal/impl/io/output_http_server.go
```diff
@@ -393,50 +420,94 @@ func (h *httpServerOutput) wsHandler(w http.ResponseWriter, r *http.Request) {
...
+			}
+			if writeErr != nil {
+				h.mWSError.Incr(1)
+				_ = ts.Ack(ctx, writeErr)
+				return // Exit the loop on write error
+			}
+			_ = ts.Ack(ctx, nil)
+		case <-ticker.C:
+			// Send a ping message to the client
+			// nolint:errcheck // this function does not actually return an error
```
jem-davies, line 497:
```suggestion
			//nolint:errcheck // this function does not actually return an error
```

buger:
Applied.

internal/impl/io/output_http_server.go
```diff
@@ -145,6 +160,18 @@ Please note, messages are considered delivered as soon as the data is written to
...
+				Default("10s").
+				Advanced(),
+			service.NewDurationField(hsoFieldPongWait).
+				Description("The time allowed to read the next pong message from the client.").
+				Default("60s").
+				Advanced(),
+			service.NewDurationField(hsoFieldPingPeriod).
+				Description("Send pings to client with this period. Must be less than pong wait.").
+				Default("54s").
+				Advanced(),
```
jem-davies, line 174:
```suggestion
			service.NewDurationField(hsoFieldWriteWait).
				Description("The time allowed to write a message to the websocket.").
				Default("10s").
				AtVersion("1.4.0").
				Advanced(),
			service.NewDurationField(hsoFieldPongWait).
				Description("The time allowed to read the next pong message from the client.").
				Default("60s").
				AtVersion("1.4.0").
				Advanced(),
			service.NewDurationField(hsoFieldPingPeriod).
				Description("Send pings to client with this period. Must be less than pong wait.").
				Default("54s").
				AtVersion("1.4.0").
				Advanced(),
```

jem-davies: at this AtVersion in as well

jem-davies (dismissed):
am happy to merge if @gregfurman is -- thanks a lot !

gregfurman (approved):
LGTM! Thanks @buger for addressing the comments and @jem-davies for the extra 👀

PR #130, opened 2024-10-11
 kafka-franz: Add more configurable fields for kafka-franz performance tuning

gregfurman:
Adds the following fields to `input.kafka_franz`:
- A list of `group_balancers` (default `cooperative_sticky`)
- `metadata_max_age` (default 5m)
- `fetch_max_bytes` (default 50MiB)
- `fetch_max_partition_bytes` (default 1MiB)
- `fetch_max_wait` (default 5s)
- `preferring_lag` (optional, no default)

Adds the following fields to `output.kafka_franz`:
- `partitioner` has a `uniform_bytes` enum
- `uniform_bytes_options` (default `{ "bytes" : "1MiB",  "adaptive" : false, "keys" : false }`)
  - Note: we do not yet allow specifying a `PartitionerHasher`. This approach should probably be reworked using bloblang with some lint rule.
- `max_buffered_records` (default 10K)
- `metadata_max_age` (default 5m)

A new integration test `TestIntegrationKafkaWarpstreamDefaults` has also been added with the performance tuning defaults found [here](https://docs.warpstream.com/warpstream/byoc/configure-kafka-client/tuning-for-performance#producer-configuration).

### DIFF
--- a/internal/impl/couchbase/integration_test.go
+++ b/internal/impl/couchbase/integration_test.go
@@ -14,6 +14,7 @@ import (
 	"github.com/ory/dockertest/v3/docker"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+
 	"github.com/warpstreamlabs/bento/public/service"
 	"github.com/warpstreamlabs/bento/public/service/integration"
 )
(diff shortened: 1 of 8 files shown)


richardartoul (approved):
Add `ProducerBatchMaxBytes` then we can merge it

gregfurman:
@richardartoul The component is already there just named incorrectly from the looks of it:


https://github.com/warpstreamlabs/bento/blob/acb7a046307d079cdd89bc1448087c737d5f45ed/internal/impl/kafka/output_kafka_franz.go#L333

https://github.com/warpstreamlabs/bento/blob/acb7a046307d079cdd89bc1448087c737d5f45ed/internal/impl/kafka/output_kafka_franz.go#L205-L216

PR #132, opened 2024-10-16
Gcp big query interpolate table

jem-davies:


richardartoul (approved):
Stamping to unblock cause it looks right to me, you can merge once the unit tests are updated

PR #135, opened 2024-10-19
Enable using AWS Secrets from Secret Manager for SQL DSN URLs

jem-davies:


internal/impl/sql/conn_fields.go
```diff
@@ -102,7 +110,16 @@ CREATE TABLE IF NOT EXISTS some_table (
 			Description("An optional maximum number of open connections to the database. If conn_max_idle is greater than 0 and the new conn_max_open is less than conn_max_idle, then conn_max_idle will be reduced to match the new conn_max_open limit. If `value <= 0`, then there is no limit on the number of open connections. The default is 0 (unlimited).").
 			Optional().
 			Advanced(),
+		service.NewStringField("secret_name").
+			Description("An optional field that can be used to get the user + password from AWS Secrets Manager").
```
jem-davies, line 114: update with all the fields that can be set

internal/impl/sql/conn_fields.go
```diff
@@ -257,7 +281,65 @@ func reworkDSN(driver, dsn string) (string, error) {
...
+			log.Fatal("Failed to parse DSN:", err)
+		}
+
+		username := parsedDSN.User.Username()
+		password, _ := parsedDSN.User.Password()
+		host := parsedDSN.Hostname()
+		port := parsedDSN.Port()
+		dbName := parsedDSN.Path[1:]
+
+		cfg, err := aws_config.LoadDefaultConfig(context.TODO())
```
jem-davies, line 298: Need to change to use the session from ./internal/impl/aws

internal/impl/sql/conn_fields.go
jem-davies: only allowing setting username + password

website/docs/guides/cloud/aws.md
```diff
@@ -50,7 +50,7 @@ If you are using short term credentials then you will also need to set the field
 credentials:
   id: foo     # aws_access_key_id
   secret: bar # aws_secret_access_key
-  token: baz  # aws_session_token
+  token: baz  # awsConfion_token
```
richardartoul, line 53: is this a typo?

website/cookbooks/redshift.md
```diff
@@ -0,0 +1,218 @@
...
+    region: eu-west-1
+    credentials:
+      id: "$AWS_ACCESS_KEY_ID"
+      secret: "$AWS_SECRET_ACCESS_KEY"
+```
+
+ - `driver`: Redshift is compatible with PostgreSQL, so we set the driver field to `postgres`
+ - `dsn`: We configure this to connect to Redshift, but the username + password we have them set to `from_secret`, Bento will fetch the secret and update the DSN to use them.
+ - `secret_name`: Here we give the secret_name of the secret Bento will use to update the DSN. 
+ - `region` + `credentials`: This is one way to give the Bento Config access to the Secret in AWS.
```
richardartoul, line 54: Why are you interleaving the documentation for region and credentials, they're separate no?

website/cookbooks/redshift.md
```diff
@@ -0,0 +1,218 @@
...
+
+The other fields are can be looked up on the [sql_insert][credentials] page.
+
+_______________________
```
richardartoul, line 58: You never explicitly explain how to format the secret value in the secret manager. You should add a paragraph here saying: "The secret should be stored in the AWS secret manager with the name you specified in the configuration above. The value of the secretion should be a JSON documentation with a username/password: $EXAMPLE."

Then:

"Below is a sample standalone terraform module that creates a redshift cluster, username/password, and stores the username/password in AWS Secret manager in the right format for Bento to read it."

Just so there is some explanation of the terraform

website/cookbooks/redshift.md
```diff
@@ -0,0 +1,218 @@
...
+
+So you are going to need: 
+  - A Redshift Cluster
+  - A Secret in SecretManager, with username + password keys as the secret value
+  - Some AWS Access Keys Bento can use to get the Secret
+
+
+```
+redshift_endpoint = "my-redshift-cluster.[...].eu-west-1.redshift.amazonaws.com:5439"
+redshift_master_password_secret_name = "redshift-password"
```
richardartoul, line 21: This should be `REDSHIFT_SECRET_NAME` I think since you reference it as `$REDSHIFT_SECRET_NAME` later on

internal/impl/sql/conn_fields.go
```diff
@@ -257,7 +279,67 @@ func reworkDSN(driver, dsn string) (string, error) {
...
+
+	return *result.SecretString, nil
+}
+
+func BuildAwsDsn(dsn string, driver string, secretName string, awsConf aws.Config, getSecretFunc func(secretName string, awsConf aws.Config) (awsSecret string, err error)) (awsSecretDsn string, err error) {
+	if secretName != "" && driver == "postgres" {
+
+		parsedDSN, err := url.Parse(dsn)
+		if err != nil {
+			return "", err
```
richardartoul, line 301: wrap each of these return errors like;

```
return fmt.Errorf("error parsing DNS URL: %w")
```

to provide more context

internal/impl/sql/conn_fields.go
```diff
@@ -257,7 +279,67 @@ func reworkDSN(driver, dsn string) (string, error) {
...
+		username := parsedDSN.User.Username()
+		password, _ := parsedDSN.User.Password()
+		host := parsedDSN.Hostname()
+		port := parsedDSN.Port()
+		path := parsedDSN.Path
+		rawQuery := parsedDSN.RawQuery
+
+		secretString, err := getSecretFunc(secretName, awsConf)
+		if err != nil {
+			return "", err
```
richardartoul, line 313: wrap

internal/impl/sql/conn_fields.go
```diff
@@ -257,7 +279,67 @@ func reworkDSN(driver, dsn string) (string, error) {
...
+		}
+
+		var secrets map[string]interface{}
+		if err := json.Unmarshal([]byte(secretString), &secrets); err != nil {
+			return "", err
```
richardartoul, line 318: wrap

internal/impl/sql/conn_fields.go
```diff
@@ -257,7 +279,67 @@ func reworkDSN(driver, dsn string) (string, error) {
...
+		}
+
+		if val, ok := secrets["username"].(string); ok && val != "" {
+			username = val
+		}
```
richardartoul, line 323: Should it just be an error if username is missing? It won't work right?

internal/impl/sql/conn_fields.go
```diff
@@ -257,7 +279,67 @@ func reworkDSN(driver, dsn string) (string, error) {
...
+		if val, ok := secrets["password"].(string); ok && val != "" {
+			password = val
```
richardartoul, line 325: same, should we just error if its not defined?

website/cookbooks/redshift.md
```diff
@@ -0,0 +1,218 @@
...
+
+## Bento Config
+
+You will need to create an AWS Access Key that can retrieve the aforementioned secret, in the below example we have used `credentials` in the config, but you can configure this a number of ways, to find out what options you have you can [read the guide][credentials].
+
+```yaml
+output:
+  sql_insert:
+    driver: postgres 
+    dsn: postgresql://from_secret:from_secret@"$REDSHIFT_ENDPOINT"/dev
```
richardartoul, line 34: After reading the code a bit, it doesn't actually matter at all what this string is right? if so I'd recommend making it:

```
postgresql://username_from_secret:password_from_secret@"$REDSHIFT_ENDPOINT"/dev
```

website/cookbooks/redshift.md
```diff
@@ -0,0 +1,218 @@
...
+      CREATE TABLE test (name varchar(255), age int);
+    secret_name: "$REDSHIFT_SECRET_NAME"
+    region: eu-west-1
+    credentials:
+      id: "$AWS_ACCESS_KEY_ID"
+      secret: "$AWS_SECRET_ACCESS_KEY"
+```
+
+ - `driver`: Redshift is compatible with PostgreSQL, so we set the driver field to `postgres`
+ - `dsn`: We configure this to connect to Redshift, but the username + password we have them set to `from_secret`, Bento will fetch the secret and update the DSN to use them.
```
richardartoul, line 52: Add another sentenec: "Note that the username/password stored in the DNS are ignored completely if we're reading them from a secret, however, we still need the DNS to have a valid "shape" so that we can parse it in Bento and replace the username/password, that's why we use the "username_from_secret" and "password_from_secret" placeholders. The $REDSHIFT_ENDPOINT is not a placeholder and must be manually subtituted with the real endpoint for your Redshift cluster.

richardartoul (approved):
Left some comments, but looks really good to me.

website/docs/guides/cloud/aws.md
jem-davies: fixed

website/cookbooks/redshift.md
jem-davies: removed region, from this specific line - kind of not relevant anyway I think

website/cookbooks/redshift.md
jem-davies: 👍

website/cookbooks/redshift.md
jem-davies: added a note that links to the interpolation page.

internal/impl/sql/conn_fields.go
jem-davies: 👍 DSN but yeah good spot

internal/impl/sql/conn_fields.go
jem-davies: 👍

internal/impl/sql/conn_fields.go
jem-davies: so I parse the username from the DSN , and if the username is extracted from the secret value I update the DSN otherwise continue with what was in the original DSN.

internal/impl/sql/conn_fields.go
jem-davies: ☝️

website/cookbooks/redshift.md
jem-davies: 👍

jem-davies: The link to the interpolation field should explain that the $REDSHIFT_ENDPOINT could actually be how the config might look.

PR #137, opened 2024-10-20
add: GCP Storage Write API compatible output sink

gregfurman:
### Summary
- Adds a new GCP sink that uses the[ BigQuery Storage Write API](https://cloud.google.com/bigquery/docs/write-api).
- Uses the [goccy/bigquery-emulator](https://github.com/goccy/bigquery-emulator) BigQuery emulator for testing
- This component currently does NOT support:
  - create/write dispositions 
  - CSV entries, where data should be read-in as JSON
  - Stream Types other than `DEFAULT` (see other options in [GCP docs](https://cloud.google.com/bigquery/docs/write-api#application-created_streams))


internal/impl/gcp/output_bigquery_storage.go
```diff
@@ -0,0 +1,341 @@
...
+	if err != nil {
+		return
+	}
+
+	switch streamType {
+	case string(managedwriter.DefaultStream):
+		conf.streamType = managedwriter.DefaultStream
+		// TODO Future versions of Bento will extend support for different stream types.
+	case string(managedwriter.CommittedStream), string(managedwriter.BufferedStream), string(managedwriter.PendingStream):
+		conf.streamType = managedwriter.DefaultStream
```
richardartoul, line 107: Why not just error in the meantime?

internal/impl/gcp/output_bigquery_storage.go
```diff
@@ -0,0 +1,341 @@
...
+		return nil, fmt.Errorf("failed to create new ManagedStream type [%s] for table [%s]: %w", destTable, string(bq.conf.streamType), err)
+	}
+
+	managedStreamPair := &streamWithDescriptor{
+		stream:            stream,
+		messageDescriptor: messageDescriptor,
+	}
+
+	bq.streamCacheLock.Lock()
+	bq.streams[destTable] = managedStreamPair
```
richardartoul, line 337: You check if this existed with an RLock, then released it, created it, and are writing it conditionally. This means you can end up overwriting a stream that was created while you weren't holding the lock. I assume thats fine and doesn't matter? If so, can you leave a comment saying why thats fine.

Also you could just avoid this problem entirely by holding the lock for the entire duration with a giant

lock
defer unlock()

at the top of the function. That will cause a bit of stalling when writing to many different tables but once the cache is filled shouldnt matter.

I'm cool either way though.

richardartoul (approved):

internal/impl/gcp/output_bigquery_storage.go
gregfurman: The linter will catch this so it shouldn't cause an issue. Misleading to keep it though I suppose. Will remove.

PR #143, opened 2024-10-29
add: log all errors wrapper for Logger component

gregfurman:
Adds an optional wrapped logger that will promote any log with an `error` type to the `ERROR` level.

This does not include ading an error in a logger's `With` or `WithField`.

For example, if you set:

```yaml
logger:
    log_all_errors: true
```

Then any subsequent `logger` calling `Warn`, `Info`,`Debug`, `Trace`, or with an `error` type in the params will promote the log to an `Error` type.

In addition, a field `@log_all_errors=true` is added to all logs generated from the strict logger:

```go
level=error msg="Trace message root module: fail"  @log_all_errors=true @service=bento_service
```


internal/log/strict.go
```diff
@@ -0,0 +1,77 @@
...
+	return wrappedLogger.With("@strict", true)
+}
+
+func (s *strictLogger) promoteOnError(promotedFrom, format string, v ...any) bool {
+	for _, arg := range v {
+		if arg == nil {
+			continue
+		}
+
+		switch arg.(type) {
```
richardartoul, line 21: I think you probably want something like this:

```golang
_, isErr := arg.(error)
```

instead to see if it implements the error interface, not if its literally an error no?

internal/log/config.go
```diff
@@ -14,13 +14,15 @@ const (
 	fieldFilePath         = "path"
 	fieldFileRotate       = "rotate"
 	fieldFileRotateMaxAge = "rotate_max_age_days"
+	fieldStrictLogging    = "strict"
```
richardartoul, line 17: `log_all_errors` IMO instead of "strict" since thats way more clear

richardartoul: Update all the variables / go code to match the name as well

richardartoul (approved):
LGTM with a few comments

internal/log/strict.go
gregfurman: Wont the switch case also check if the value `arg` implements the `error` interface? 

Or are you suggesting we remove the check for `*error` as well since that's _technically_ not an `error` type.

internal/log/config.go
gregfurman: Done!

internal/log/strict.go
richardartoul: I guess if your tests pass its fine

richardartoul (approved):

internal/log/strict.go
```diff
@@ -0,0 +1,77 @@
+package log
```
richardartoul, line 1: rename this file

internal/log/strict_test.go
```diff
@@ -0,0 +1,97 @@
+package log
```
richardartoul, line 1: rename this file

internal/log/strict_test.go
```diff
@@ -0,0 +1,97 @@
...
+}
+
+func (t testErr) Error() string {
+	return "custom error"
+}
+func (t testErr) Foo() string {
+	return "bar"
+}
+
+func TestStrictLogger(t *testing.T) {
```
richardartoul, line 25: rename this test

richardartoul (approved):

jem-davies:
I acknowledge the request for review - will give it a good look soon 👍

website/docs/components/logger/about.md
```diff
@@ -138,3 +138,12 @@ The maximum number of days to retain old log files based on the timestamp encode
...
 Default: `0`  
 
+### `log_all_errors`
+
+Experimental: When set, the logger will promote any log containing an `error` type to the `ERROR` level.
+
+
+Type: `bool`  
+Default: `false`  
+Requires version 1.3.0 or newer
```
jem-davies, line 148:
```suggestion
Requires version 1.4.0 or newer  
```

internal/log/docs.go
```diff
@@ -28,6 +28,7 @@ func Spec() docs.FieldSpecs {
 			docs.FieldBool(fieldFileRotate, "Whether to rotate log files automatically.").HasDefault(false),
 			docs.FieldInt(fieldFileRotateMaxAge, "The maximum number of days to retain old log files based on the timestamp encoded in their filename, after which they are deleted. Setting to zero disables this mechanism.").HasDefault(0),
 		),
+		docs.FieldBool(fieldLogAllErrors, "Experimental: When set, the logger will promote any log containing an `error` type to the `ERROR` level.").HasDefault(false).Advanced().AtVersion("1.3.0"),
```
jem-davies, line 31:
```suggestion
		docs.FieldBool(fieldLogAllErrors, "Experimental: When set, the logger will promote any log containing an `error` type to the `ERROR` level.").HasDefault(false).Advanced().AtVersion("1.4.0"),
```

I forgot to check on the cypher output 🤕

internal/log/config.go
```diff
@@ -14,13 +14,15 @@ const (
 	fieldFilePath         = "path"
 	fieldFileRotate       = "rotate"
 	fieldFileRotateMaxAge = "rotate_max_age_days"
+	fieldLogAllErrors     = "log_all_errors"
```
jem-davies, line 17: I think that something like `promote_errors` is a better name, because that is more descriptive of what this field does? Taken the word promote from the description you have written.

jem-davies: I might think that like `log_all_errors` takes the message contents and logs any/all errors in the message data?

jem-davies (changes requested):
LGTM - just the version number

jem-davies (approved):

PR #147, opened 2024-10-30
Add support for int8/16 to parquet encode

richardartoul:


internal/impl/parquet/convert.go
```diff
@@ -99,44 +197,65 @@ func setField(field reflect.Value, value any) error {
...
 			field.SetFloat(float64(v))
-			return nil
+		case int8:
+			field.SetFloat(float64(v))
+		case int16:
+			field.SetFloat(float64(v))
+		case int32:
+			field.SetFloat(float64(v))
 		case int64:
 			field.SetFloat(float64(v))
```
arjunnair1997, line 254: is int64 to float 32 lossy?

internal/impl/parquet/convert.go
```diff
@@ -99,44 +197,65 @@ func setField(field reflect.Value, value any) error {
...
-			return nil
+		case float32:
+			field.SetFloat(float64(v))
+		case float64:
+			field.SetFloat(v)
```
arjunnair1997, line 258: Is this lossy? Or is the float32 field just stored as float64

arjunnair1997 (approved):

internal/impl/parquet/convert.go
richardartoul: definitely lossy, but wont underflow/overflow

internal/impl/parquet/convert.go
richardartoul: I think so, but would have to be a humongous number. We can fix this later if it comes up, thats how it worked before

PR #150, opened 2024-11-03
add: Jitter duration field to batching policy

gregfurman:
## Changes
This PR introduces a new field to Bento's `batching` policy that allows for jittered flushes of new batches. The actual jittered delay is calculated as a random duration between `0-jitter*period`.

A lot of docs are updated as a result of this which have been added in a seperate commit.

### DIFF
--- a/internal/batch/policy/policy_test.go
+++ b/internal/batch/policy/policy_test.go
@@ -2,6 +2,7 @@ package policy_test
 
 import (
 	"context"
+	"math/rand"
 	"reflect"
 	"testing"
 	"time"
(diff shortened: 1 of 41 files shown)


internal/batch/policy/policy_test.go
```diff
@@ -127,6 +128,50 @@ func TestPolicyPeriod(t *testing.T) {
...
+
+	tCtx, done := context.WithTimeout(context.Background(), time.Second*30)
+	t.Cleanup(func() {
+		require.NoError(t, pol.Close(tCtx))
+		done()
+	})
+
+	// Set RNG seed to ensure deterministic test runs
+	//nolint:staticcheck
+	rand.Seed(1)
```
jem-davies, line 149:
```suggestion
	rand.New(rand.NewSource(1))
```

don't think it's a big deal

jem-davies (changes requested):
~~LGTM , just added 1 comment to get rid of the depreciated function~~

jem-davies (approved):

PR #158, opened 2024-11-10
GitHub actions updates and fixes

jem-davies:
- use free up disk space job github action - rather than bash script to delete more stuff
 - fix a few unit tests
 - update go versions in cross build 1.20,1.21 -> 1.21,1.22
 - update go.mod to go version 1.22
 - add version field to .goreleaser.yaml config 
 - add ldflags that were missing
 
Example workflow:  https://github.com/warpstreamlabs/bento/actions/runs/11766198856


.goreleaser.yml
```diff
@@ -16,6 +17,8 @@ builds:
       -s -w
       -X github.com/warpstreamlabs/bento/internal/cli.Version={{.Version}}
       -X github.com/warpstreamlabs/bento/internal/cli.DateBuilt={{.Date}}
+      -X main.Version={{.Version}}
+      -X main.DateBuilt={{.Date}}
```
gregfurman, line 21: Great! Can we download the binary that is created and check if this works?

.github/workflows/cross_build.yml
```diff
@@ -23,15 +23,18 @@ jobs:
       with:
         fetch-depth: 0
 
-    - name: Free up some disk space on ubuntu
+    - name: Free Disk Space (Ubuntu)
```
gregfurman, line 26: Does this have a silent mode? Currently the cleanup takes ~10 mins which is not great.

See https://github.com/warpstreamlabs/bento/actions/runs/11801043486/job/32873407799#step:3:1

gregfurman (approved):
LGTM! Just some considerations for the improved CI.

.goreleaser.yml
gregfurman: Jem manually showed me this works 🫡

internal/impl/protobuf/processor_protobuf_test.go
```diff
@@ -235,7 +235,7 @@ func TestProtobufErrors(t *testing.T) {
 			message:    "testing.Person",
 			importPath: "../../../config/test/protobuf/schema",
 			input:      `{"firstName":5,"lastName":"quaye","email":"caleb@myspace.com"}`,
-			output:     "invalid value for string type: 5",
+			output:     "invalid value for string field firstName: 5",
```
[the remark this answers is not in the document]
jem-davies: Yeah I think so

.github/workflows/cross_build.yml
jem-davies: Don't think it has a silent mode, config options are at the top of [this file](https://github.com/jlumbroso/free-disk-space/blob/main/action.yml) ... Maybe less things could be deleted such that it doesn't take as long, but this change is there because it fixes the job that was there before: that wasn't deleting enough stuff. S reluctant to change it. The previous job has a runtime ~3 mins.

Issue #162, opened 2024-11-13
Opensearch output fails when deleting non-existing document


arnitolog:
I'm creating a pipeline that will read data from kafka topic and ingest/delete it in Opensearch. Everything works fine, except the case when bento is trying to delete a non-existing document. In this case it fails with the error below:
```
time="2024-11-13T23:04:09Z" level=info msg="Running main config from specified file" bento_version=v1.3.0 path=/bento.yaml service=bento
time="2024-11-13T23:04:09Z" level=info msg="Listening for HTTP requests at: http://0.0.0.0:4195" service=bento
time="2024-11-13T23:04:09Z" level=info msg="Launching a Bento instance, use CTRL+C to close" service=bento
time="2024-11-13T23:04:09Z" level=info msg="Output type stdout is now active" label="" path=root.output.fallback.1 service=bento
time="2024-11-13T23:04:09Z" level=info msg="Output type opensearch is now active" label="" path=root.output.fallback.0.reject_errored service=bento
time="2024-11-13T23:04:09Z" level=info msg="Input type kafka is now active" label=kafka_topic path=root.input service=bento
time="2024-11-13T23:04:09Z" level=debug msg="Starting consumer group" label=kafka_topic path=root.input service=bento
time="2024-11-13T23:04:12Z" level=debug msg="Consuming messages from topic 'ap-test' partition '0'" label=kafka_topic path=root.input service=bento
time="2024-11-13T23:04:13Z" level=debug msg="Successfully dispatched [%!s(uint64=2)] documents in 334ms (%!s(int64=5) docs/sec)" label="" path=root.output.fallback.0.reject_errored service=bento
time="2024-11-13T23:06:51Z" level=debug msg="Successfully dispatched [%!s(uint64=1)] documents in 25ms (%!s(int64=40) docs/sec)" label="" path=root.output.fallback.0.reject_errored service=bento
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x1b11fdc]

goroutine 105 [running]:
github.com/warpstreamlabs/bento/internal/impl/opensearch.(*Output).buildBulkableRequest.func1({_, _}, {{0xc002410420, 0x11}, {0x59a3ca0, 0x6}, {0xc001bbaa28, 0x3}, 0x0, 0x0, ...}, ...)
[...]

PR #163, opened 2024-11-14
properly handle biri.Error

arnitolog:
fixes https://github.com/warpstreamlabs/bento/issues/162

### DIFF
--- a/internal/impl/opensearch/integration_test.go
+++ b/internal/impl/opensearch/integration_test.go
@@ -476,8 +476,7 @@ action: delete
 	}, nil)
 	require.NoError(t, err)
 	if get.IsError() {
-		var respCode int
-		respCode = get.StatusCode
+		respCode := get.StatusCode
 		if respCode == http.StatusNotFound {
 			// Document was not found, as expected
 		} else {
--- a/internal/impl/opensearch/integration_test.go
+++ b/internal/impl/opensearch/integration_test.go
@@ -451,6 +451,41 @@ action: ${! @elastic_action }
 			resEqualsJSON(t, get, string(testMsg[i]))
 		}
 	}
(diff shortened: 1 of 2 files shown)


gregfurman (commented):
Hey! Thanks for the contribution and quick fix for this. Any chance you have capacity to add an integration test to the `opensearch` suite for this case?

There are 2 deletion tests but none test this case you've encountered of trying to delete a non-existent document AFAIK. See [opensearch/integration_test.go](https://github.com/warpstreamlabs/bento/blob/main/internal/impl/opensearch/integration_test.go#L432).

arnitolog:
@gregfurman, I've added integration test for this case. Please take a look

internal/impl/opensearch/integration_test.go
```diff
@@ -451,6 +451,41 @@ action: ${! @elastic_action }
...
+
+	// Verify the document was not found
+	get, err := client.Do(ctx, osapi.DocumentGetReq{
+		Index:      "test_conn_index",
+		DocumentID: "non-existing-id",
+	}, nil)
+	require.NoError(t, err)
+	if get.IsError() {
+		var respCode int
+		respCode = get.StatusCode
```
gregfurman, line 480: Linter doesn't like this one.
```suggestion
		respCode := get.StatusCode
```

gregfurman:
Awesome! Thanks for adding that test. 

Your golangci-lint step is failing so just run `make fmt && make lint` and happy to approve 😄 

I added a suggestion that you can also accept that should fix it.

arnitolog:
@gregfurman , fixed

gregfurman (approved):
LGTM! Thanks for this 👍

PR #170, opened 2024-11-21
update old reference to a codec -> scanner aws_s3 input

jem-davies:


gregfurman:
If your make docs fail that'll be tragic

jem-davies:
> ~If your make docs fail that'll be tragic~ We good 😎

I 'member'd for a change 😆

PR #172, opened 2024-11-21
add sqs large message compatible client to input aws_sqs

jem-davies:


gregfurman:
This is looking great! Can you add a test to the sqs integration test suite (with this large bucket option enabled) testing against LocalStack?

I'm thinking maybe pre-populate the SQS queue with some data that is more than 256Kb and let's see if this bad-boy works 🤠

jem-davies:
closing this PR because functionality will be provided on https://github.com/warpstreamlabs/bento/pull/180 with the example written in the docs

PR #174, opened 2024-11-23
add: Error Handling Environment Bundle to Bento configuration

gregfurman:
## Motivation
We need global config to allow for a processing step to return and log an error, with the subsequent processors not being run.

Related to https://github.com/warpstreamlabs/bento/pull/166 -- where we opted for nacking at the failed `processor` instead of at the output layer.

Can be used as follows:
```yaml
pipeline:
  processors:
    - mapping: |
        root = throw(this.value)
    - mapping: |
        root.message = "I'm never reached..."

error_handling:
  strategy: reject
```

## Changes
- Creates config for a new `error_handling` in the top-level that can be one of `reject` or `none`
- Adds an interface that iterates through all messages from a processor and checks if any contain an error.
- Failures will be nacked at the input layer without needing to reach the `output`.
- Fixes the `generate` input component always enabling `auto_replay_nacks`.

## TODO
- ~Add more tests.~
- ~Populate this config at the manager level to allow for nested `input.processors` and `output.processors` to also have their results checked for errors.~
- Look into adding backoff for retries and shutting off the `input` to stop new data being read in.
- ~Add documentation for this feature.~

### DIFF
--- a/internal/cli/common/manager.go
+++ b/internal/cli/common/manager.go
@@ -16,6 +16,7 @@ import (
 
 	"github.com/warpstreamlabs/bento/internal/api"
 	"github.com/warpstreamlabs/bento/internal/bundle"
+	"github.com/warpstreamlabs/bento/internal/bundle/strict"
 	"github.com/warpstreamlabs/bento/internal/component/metrics"
 	"github.com/warpstreamlabs/bento/internal/config"
 	"github.com/warpstreamlabs/bento/internal/docs"
@@ -103,6 +104,11 @@ func CreateManager(
 		manager.OptSetTracer(trac),
 		manager.OptSetStreamsMode(streamsMode),
 	}, mgrOpts...)
+
+	// Initialise processors with global error handling strategy
+	if conf.ErrorHandling.Strategy == "reject" {
+		mgrOpts = append(mgrOpts, manager.OptSetEnvironment(strict.StrictBundle(bundle.GlobalEnvironment)))
+	}
 
 	// Create resource manager.
 	var mgr *manager.Type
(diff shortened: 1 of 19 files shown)


jem-davies (commented):
I would just change the key `strict_mode` to something like `on_error` with the values being an enum of, `continue` (default) / `strict_mode`. 

So that we can add new modes later on if required, or to add a new mode to change the retry behaviour later? i.e. a new `on_error: exponential_backoff`. 

I think that adding this field to the manager level also makes sense when running in streams mode, it will be ever stream that is created will have the same strict_mode / on_error mode - will simplify understanding?

internal/cli/common/manager.go
```diff
@@ -104,6 +105,11 @@ func CreateManager(
 		manager.OptSetStreamsMode(streamsMode),
 	}, mgrOpts...)
 
+	// Initialise processors with global error handling strategy
+	if conf.ErrorHandling.Strategy == "reject" {
+		mgrOpts = append(mgrOpts, manager.OptSetEnvironment(strict.StrictBundle(bundle.GlobalEnvironment)))
+	}
+
```
gregfurman, line 112: Here is where the bundler gets applied

internal/errorhandling/config.go
```diff
@@ -0,0 +1,30 @@
+package errorhandling
+
+import (
+	"github.com/warpstreamlabs/bento/internal/docs"
+	// "github.com/warpstreamlabs/bento/internal/retries"
```
jem-davies, line 5:
```suggestion
```

jem-davies (approved):

jem-davies (approved):

PR #182, opened 2024-12-05
Add new rate limiting functionality to rate-limit bytes

gregfurman:
## Changes
- New `MessageAwareRateLimit` interface with an `Add()` method.
- `Add()` is passed a slice of internal messages or message-parts where the total bytes are summated
- `Access()` will increment the counter and block if the rate-limiting is reached
- Made some changes with airgapped* implementations since we need to allow `RateLimit` and `ratelimit.V1` to be backwards compatible since its in the public API
- The `local` rate limiter has these changes implemented whereas the `redis` one has not.

### DIFF
--- a/internal/impl/pure/rate_limit_local.go
+++ b/internal/impl/pure/rate_limit_local.go
@@ -93,6 +93,15 @@ func (r *localRatelimit) Access(ctx context.Context) (time.Duration, error) {
 	r.mut.Lock()
 	defer r.mut.Unlock()
 
+	// Rate limiting count is enabled
+	if r.size > 0 {
+		r.bucket--
+	}
+
+	if r.bucket < 0 {
+		r.exceededLimit = true
+	}
+
 	if !r.exceededLimit {
 		return 0, nil
 	}
@@ -109,7 +118,7 @@ func (r *localRatelimit) Access(ctx context.Context) (time.Duration, error) {
 
 // Add increments the observation or message-bytes counter. Returns true if this
 // triggers the conditions of the rate-limiter.
-func (r *localRatelimit) Add(ctx context.Context, part *message.Part) bool {
+func (r *localRatelimit) Add(ctx context.Context, parts ...*message.Part) bool {
 	r.mut.Lock()
 	defer r.mut.Unlock()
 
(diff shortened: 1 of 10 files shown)


internal/impl/pure/rate_limit_local.go
```diff
@@ -14,8 +15,11 @@ func localRatelimitConfig() *service.ConfigSpec {
 		Stable().
 		Summary(`The local rate limit is a simple X every Y type rate limit that can be shared across any number of components within the pipeline but does not support distributed rate limits across multiple running instances of Bento.`).
 		Field(service.NewIntField("count").
-			Description("The maximum number of requests to allow for a given period of time.").
-			Default(1000)).
+			Description("The maximum number of requests to allow for a given period of time. If `0` disables count based rate-limiting.").
+			Default(1000).LintRule(`root = if this < 0 { [ "count cannot be less than zero" ] }`)).
+		Field(service.NewIntField("byte_size").
+			Description("The maximum number of bytes to allow for a given period of time. If `0` disables size based rate-limiting.").
```
jem-davies, line 21: Is this strictly true? - isn't it possible that the byte_size limit can be exceeded?

Also nit: `...disable size based...` -> `...disable byte_size based...`

internal/impl/pure/rate_limit_local.go
```diff
@@ -39,55 +43,116 @@ func newLocalRatelimitFromConfig(conf *service.ParsedConfig) (*localRatelimit, e
...
+	if time.Since(r.lastRefresh) >= r.period {
+		r.refresh()
+	}
+
+	if r.exceededLimit {
+		return true
+	}
+
+	// Rate limiting bytes is enabled
+	if r.byteSize > 0 && len(parts) != 0 {
```
jem-davies, line 134:
```suggestion
	if r.byteSize > 0 {
```

nit: is that `len(parts) != 0` required if it is then subsequently in the for loop condition?

internal/impl/pure/rate_limit_local.go
```diff
@@ -39,55 +43,116 @@ func newLocalRatelimitFromConfig(conf *service.ParsedConfig) (*localRatelimit, e
...
+		for i := 0; i < len(parts) && r.byteBucket >= 0; i++ {
+			if parts[i] != nil {
+				r.byteBucket -= len(parts[i].AsBytes())
+			}
+		}
+	}
+
+	if r.byteBucket < 0 {
```
jem-davies, line 142: could move this check into the loop and return early?

jem-davies (commented):
looks good, just small small comments

internal/impl/pure/rate_limit_local.go
gregfurman: Yeah this isn't necesary. Will change

internal/impl/pure/rate_limit_local.go
gregfurman: The loop will break early due to that `&& r.byteBucket >= 0;` condition. Guess this is a bit confusing though so happy to rather return early

internal/impl/pure/rate_limit_local.go
gregfurman: I mean, not really. If you exceed the byte_size and then try access the resource -- things should block

> Also nit: ...disable size based... -> ...disable byte_size based...
Nice catch. Changed

internal/impl/pure/rate_limit_local.go
jem-davies: ahh yeah makes sense , guess having the return in the loop would make it a bit clearer

gregfurman: done

jem-davies (approved):
LGTM

PR #187, opened 2024-12-11
Add kafka rate limit

jem-davies:
### Changes

- Add new rate limit field to Kafka input in order, which pauses polling when rate is exceeded.

### DIFF
--- a/internal/impl/kafka/input_kafka_franz.go
+++ b/internal/impl/kafka/input_kafka_franz.go
@@ -126,6 +126,10 @@ With this option, you can return topic order and per-topic partition ordering. T
 		Field(service.NewBatchPolicyField("batching").
 			Description("Allows you to configure a [batching policy](/docs/configuration/batching) that applies to individual topic partitions in order to batch messages together before flushing them for processing. Batching can be beneficial for performance as well as useful for windowed processing, and doing so this way preserves the ordering of topic partitions.").
 			Advanced()).
+		Field(service.NewStringField("rate_limit").
+			Description("An optional [`rate_limit`](/docs/components/rate_limits/about) to throttle invocations by.").
+			Default("").
+			Advanced()).
 		LintRule(`
 let has_topic_partitions = this.topics.any(t -> t.contains(":"))
 root = if $has_topic_partitions {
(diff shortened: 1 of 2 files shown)


internal/impl/kafka/input_kafka_franz.go
```diff
@@ -490,6 +528,8 @@ func (p *partitionTracker) loop() {
 			}()
 
 			if len(sendBatch) > 0 {
+				_ = p.rateLimitFn(closeAtLeisureCtx, sendBatch)
```
gregfurman, line 531: I think we need to be rate-limiting at this step:

```golang
			pauseTopicPartitions := map[string][]int32{}
			iter := fetches.RecordIter()
			for !iter.Done() {
				record := iter.Next()
				if checkpoints.addRecord(closeCtx, f.recordToMessage(record), f.checkpointLimit) {
					pauseTopicPartitions[record.Topic] = append(pauseTopicPartitions[record.Topic], record.Partition)
				}
			}
```

Where we actually read in the KAfka data + then convert to bento messages.

internal/impl/kafka/input_kafka_franz.go
```diff
@@ -819,8 +859,10 @@ func (f *franzKafkaReader) Connect(ctx context.Context) error {
 			iter := fetches.RecordIter()
 			for !iter.Done() {
 				record := iter.Next()
-				if checkpoints.addRecord(closeCtx, f.recordToMessage(record), f.checkpointLimit) {
+				if batch, pause := checkpoints.addRecord(closeCtx, f.recordToMessage(record), f.checkpointLimit); pause {
 					pauseTopicPartitions[record.Topic] = append(pauseTopicPartitions[record.Topic], record.Partition)
+				} else {
+					f.waitForAccess(ctx, batch)
```
gregfurman, line 865: This loop reads a single record at a time. We now extract a batch of records to-be-sent from `addRecord` and rate-limit if the batch exceeds the resource amount.

cc @jem-davies

jem-davies (commented):
@gregfurman LGTM - can merge if you happy

PR #189, opened 2024-12-16
Add backoff and error payload sampling to error handling bundle

gregfurman:
## Changes

- Adds an `ErrorSampleBundle` that will randomly select a proportion of failed events and log them -- allowing us to see the payload of those events that failed processing.
- Adds a `RetryBundle` that will retry failed processing steps with exponential backoff -- closing the input layer while retrying.

### Errored Payload Sampling
The below will log a proportion of all failed payloads:
```yaml
error_handling:
  error_sample_rate: 0.5
```

### Retry with backoff

```yaml
error_handling:
  strategy: backoff
  max_retries: 0 # default is 0 which means we will retry forever.
```

## TODO
- [x] Add configuration tests for the new `error_handling` fields.
- [x] Add mocked tests for `error_sample_rate`.
- [x] Validations for `max_retries` field.


jem-davies (commented):
will do another look at the processors but seems ok to me

internal/bundle/strict/bundle.go
```diff
@@ -1,28 +1,118 @@
...
+	return errorSampleEnv
+}
+
+// RetryBundle returns a modified environment to reprocess any batches containing any message-level errors
+// through the entire pipeline. During retries, the input-layer will pause new message ingestion.
+func RetryBundle(ctx context.Context, b *bundle.Environment, maxRetries int) *bundle.Environment {
+	env := b.Clone()
+
+	for _, spec := range b.InputDocs() {
+		_ = env.InputAdd(func(conf input.Config, nm bundle.NewManagement) (input.Streamed, error) {
```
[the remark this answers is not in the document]
gregfurman: I was following the lead of the other `tracing` bundle -- where this is done.

Thinking to leave as anonymous function since it's very specialised to the initialisation/overloading and can't really imagine it being used elsewhere.

internal/cli/common/manager.go
```diff
@@ -106,8 +106,10 @@ func CreateManager(
 	}, mgrOpts...)
 
 	// Initialise processors with global error handling strategy
-	if conf.ErrorHandling.Strategy == "reject" {
-		mgrOpts = append(mgrOpts, manager.OptSetEnvironment(strict.StrictBundle(bundle.GlobalEnvironment)))
+	if conf.ErrorHandling.Strategy != "none" || conf.ErrorHandling.ErrorSampleRate > 0 {
```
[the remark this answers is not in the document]
jem-davies: Yeah I think checking it like this is fine

gregfurman, line 109: OK sick. FYI, I made this `"none"` because that was used by `buffer` and some other components to indicate a disabled type. Changing this though yo `""`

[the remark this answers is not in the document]
gregfurman: OK sick. FYI, I made this `"none"` because that was used by `buffer` and some other components to indicate a disabled type. Changing this to `""` though :)

internal/bundle/strict/processor_sample_error.go
```diff
@@ -0,0 +1,94 @@
...
+				if parseErr != nil {
+					fields["parse_payload_error"] = parseErr.Error()
+				} else {
+					// Convert to string representation
+					if jsonBytes, err := json.Marshal(sobj); err == nil {
+						fields["structured_payload"] = string(jsonBytes)
+					} else {
+						fields["structured_payload_error"] = "failed to marshal structured payload"
+					}
+				}
```
jem-davies, line 77: I think we could just return the `raw_payload` field and not worry about marshalling it to json then returning either a `structured_payload` / `structured_payload_error` field, just because it's an assumption to make, that it will be a json message, it might not be, (could be a csv row) and marshalling every message part in a loop could be resource intensive. 

It probably makes sense to add a field to enable or disable these `structured_payload` / `structured_payload_error` fields because it sounds contextual to specific application of Bento to me?

internal/errorhandling/docs.go
```diff
@@ -4,6 +4,10 @@ import "github.com/warpstreamlabs/bento/internal/docs"
 
 func Spec() docs.FieldSpecs {
 	return docs.FieldSpecs{
-		docs.FieldString(fieldStrategy, "The error handling strategy.").HasOptions("none", "reject").HasDefault("none"),
+		docs.FieldString(fieldStrategy, "The error handling strategy.").HasOptions("reject", "backoff").HasDefault(""),
+		docs.FieldFloat(fieldErrorSampleRate, "Proportion of failed message payloads to randomly sample from and log.").HasDefault(0).
+			LinterBlobl(`root = if this < 0 || this > 1 { "error_sample_rate should be between 0 and 1." }`),
+		docs.FieldInt(fieldMaxRetries, "Maximum retries before dropping a message for a `backoff` strategy. `0` indicates indefinite retries.").HasDefault(0).
```
jem-davies, line 10: "Maximum retries before dropping a message for a `backoff` strategy." This took me a while to understand after reading the source for processor_retry.go. 

I think that it would be better if we could look at making it so that a config like:

```yaml
error_handling:
  strategy: reject
  max_retries: 2
```

Would fail initialising because the `max_retries` is only for `strategy: backoff`? 

Also where it says: "retries before _dropping_ a message" - the messages don't get dropped when using this strategy, they get retried X times, then they continue with errors so a config like: 

```yaml
error_handling:
  strategy: backoff
  max_retries: 2

pipeline:
  processors:
    - mapping: |
        root = throw("error")
    - mapping: | 
        root.message = "I'm never reached"
```

Would do this: 

```
{}
ERRO failed assignment (line 1): error             @service=bento label="" path=root.pipeline.processors.0
ERRO failed assignment (line 1): error             @service=bento label="" path=root.pipeline.processors.0
ERRO failed assignment (line 1): error             @service=bento label="" path=root.pipeline.processors.0
{"message":"I'm never reached"}
```

jem-davies (changes requested):
I think there is an issue with the description of the `backoff` / `max_retries` - about the dropping of messages but it doesn't drop messages like the `reject` strategy? - it will continue along the pipeline after doing the max_retries?

I think it would be better as well if the content on [/components/processors/about#error-handling](https://warpstreamlabs.github.io/bento/docs/components/processors/about#error-handling) was moved to [/docs/configuration/about](https://warpstreamlabs.github.io/bento/docs/configuration/about) because this is where the other top level config options such as `shutdown_delay` are described? 

I noticed that there is a relative link to `error-handling` [here](http://localhost:3000/bento/docs/components/processors/about#error-handling) when it should probably be [error-handling-1](http://localhost:3000/bento/docs/components/processors/about#error-handling-1) it's a bit confusing because there are two headings both 'error-handling'? 

I think as well that there isn't enough explanation in the docs currently about the backoff strategy - about how that will block input and the default options for the:

```go
backoffCtor := func() backoff.BackOff {
		boff := backoff.NewExponentialBackOff()
		boff.InitialInterval = time.Millisecond * 100
		boff.MaxInterval = time.Second * 60
		boff.MaxElapsedTime = 0
		return boff
	}
```

internal/errorhandling/docs.go
gregfurman: > Would fail initialising because the `max_retries` is only for `strategy: backoff`?

Yeah happy to add a linting check to this.

> Also where it says: "retries before dropping a message" - the messages don't get dropped ...

Oh yeah that is correct. Definitely an oversight on my end. Should we drop messages or do you think continuing as it currently does is sufficient?

internal/bundle/strict/processor_sample_error.go
gregfurman: I don't think we should worry about overhead tbh. If resource intensivity is an issue, a user can just sample a smaller proportion of failed payloads.

If I read in CSV data, don't we convert this to a JSON friendly format anyway when constructing messages? So the structured payload will still correctly show CSV data but just in JSON/structured form

jem-davies (approved):

website/docs/configuration/about.md
```diff
@@ -296,6 +296,62 @@ The `shutdown_timeout` option sets a hard deadline for Bento process to graceful
...
+```
+
+### Strategy
+
+If any message within a batch contains an error, the `error_handling.strategy` determines how we will deal with the entire batch. The currently supported strategies are as follows:
+
+| Strategy | Description |
+|----------|-------------|
+| `reject` | Immediately fails entire batch when any message has an error. Propagates negative acknowledgment (nack) to the input. |
+| `backoff` | Retries batches containing failed messages using exponential backoff. Starting at `100ms`, the delay doubles after each retry up to a maximum of `60s` between attempts. When `max_retries` is `0` (default), retries continue indefinitely. Otherwise, messages are dropped from the batch after the maximum retries are exhausted. |
```
jem-davies, line 325:
```suggestion
| Strategy | Version | Description |
|----------|-------------|-------------|
| `reject` | v1.4.0 |Immediately fails entire batch when any message has an error. Propagates negative acknowledgment (nack) to the input. |
| `backoff` | v1.5.0 |Retries batches containing failed messages using exponential backoff. Starting at `100ms`, the delay doubles after each retry up to a maximum of `60s` between attempts. When `max_retries` is `0` (default), retries continue indefinitely. Otherwise, messages are dropped from the batch after the maximum retries are exhausted. |
```

Think would be good to add a version column for the strategies.

gregfurman:
Closed in favour of https://github.com/warpstreamlabs/bento/pull/221

PR #190, opened 2024-12-17
add unsafe_dynamic_query to gcp_bigquery_select

jem-davies:
Will enable things like: 

```yaml
# {"table": "test.people", "where": "city IN (?,?,?)", "columns": ["name", "age", "city"], "args": ["London", "Paris", "Dublin"]}
pipeline:
  processors:
    - gcp_bigquery_select:
        project: ${GGP_PROJECT}
        table: ${! this.table } # test.people
        columns_mapping: root = this.columns #["name", "age", "city"]
        where:  ${! "city IN ("+this.args.join(",").re_replace_all("\\b\\w+\\b","?")+")" } # city IN (?,?,?)
        args_mapping: root = this.args # ["London", "Paris", "Dublin"]
        unsafe_dynamic_query: true
```

So that there is greater flexibility with this processor.

### DIFF
--- a/internal/impl/gcp/processor_bigquery_select.go
+++ b/internal/impl/gcp/processor_bigquery_select.go
@@ -164,7 +164,7 @@ An example to show the use of the unsafe_dynamic_query field:`,
 pipeline:
   processors:
     - gcp_bigquery_select:
-        project: ${GGP_PROJECT}
+        project: ${GCP_PROJECT}
         table: ${! this.table } # test.people
         columns_mapping: root = this.columns #["name", "age", "city"]
         where:  ${! "city IN ("+this.args.join(",").re_replace_all("\\b\\w+\\b","?")+")" } # city IN (?,?,?)
--- a/internal/impl/gcp/processor_bigquery_select.go
+++ b/internal/impl/gcp/processor_bigquery_select.go
@@ -141,7 +141,7 @@ func newBigQuerySelectProcessorConfig() *service.ConfigSpec {
 			Description("An optional suffix to append to the select query.").
 			Optional()).
 		Field(service.NewBoolField("unsafe_dynamic_query").
-			Description("Whether to enable [interpolation functions](/docs/configuration/interpolation/#bloblang-queries) in the columns & where fields. Great care should be made to ensure your queries are defended against injection attacks.").
+			Description("Whether to enable [interpolation functions](/docs/configuration/interpolation/#bloblang-queries) in the columns_mapping, table & where fields. When `unsafe_dynamic_query` is set to true, you should provide a bloblang mapping via the `columns_mapping` config field, and not `columns`. Great care should be made to ensure your queries are defended against injection attacks.").
 			Advanced().
 			Default(false).
 			Version("1.5.0").
--- a/internal/impl/gcp/processor_bigquery_select.go
+++ b/internal/impl/gcp/processor_bigquery_select.go
@@ -92,16 +92,16 @@ func bigQuerySelectProcessorConfigFromParsed(inConf *service.ParsedConfig) (conf
 	}
 
 	// check config fields are being used appropriately
-	if !inConf.Contains("columns_mapping") && conf.unsafeDyn {
+	if conf.queryParts.columnsMapping == nil && conf.unsafeDyn {
 		err = errors.New("invalid gcp_bigquery_select config: unsafe_dynamic_query set to true but no columns_mapping provided")
+		return
+	}
+	if inConf.Contains("columns_mapping") && !conf.unsafeDyn {
+		err = errors.New("invalid gcp_bigquery_select config: unsafe_dynamic_query set to false but columns_mapping provided")
 		return
 	}
 	if conf.queryParts.columnsMapping != nil && len(conf.queryParts.columns) != 0 {
 		err = errors.New("invalid gcp_bigquery_select config: cannot set both columns_mapping and columns field")
-		return
-	}
-	if len(conf.queryParts.columns) != 0 && conf.unsafeDyn {
-		err = errors.New("invalide gcp_bigquery_select config: unsafe_dynamic_query set to true but columns field provided")
 		return
 	}
 
--- a/internal/impl/gcp/processor_bigquery_select.go
+++ b/internal/impl/gcp/processor_bigquery_select.go
@@ -160,7 +160,7 @@ pipeline:
 			`
 An example to show the use of the unsafe_dynamic_query field:`,
 			`
-# {"table": "test.people", "where": "city IN (?,?,?)", "columns": ["name", "age", "city"], "args": ["London", "Paris", "Dublin"]}
+# {"table": "test.people", "columns": ["name", "age", "city"], "args": ["London", "Paris", "Dublin"]}
 pipeline:
   processors:
     - gcp_bigquery_select:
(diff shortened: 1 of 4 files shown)


internal/impl/gcp/processor_bigquery_select.go
```diff
@@ -118,6 +154,22 @@ pipeline:
...
+		).
+		Example("Unsafe Dynamic Query",
+			`
+An example to show the use of the unsafe_dynamic_query field:`,
+			`
+# {"table": "test.people", "where": "city IN (?,?,?)", "columns": ["name", "age", "city"], "args": ["London", "Paris", "Dublin"]}
+pipeline:
+  processors:
+    - gcp_bigquery_select:
+        project: ${GGP_PROJECT}
```
jem-davies, line 167:
```suggestion
        project: ${GCP_PROJECT}
```

internal/impl/gcp/processor_bigquery_select.go
```diff
@@ -40,16 +46,33 @@ func bigQuerySelectProcessorConfigFromParsed(inConf *service.ParsedConfig) (conf
...
+	if conf.unsafeDyn {
+		if queryParts.tableDyn, err = inConf.FieldInterpolatedString("table"); err != nil {
+			return
+		}
+		if inConf.Contains("where") {
+			if queryParts.whereDyn, err = inConf.FieldInterpolatedString("where"); err != nil {
+				return
+			}
+		}
+		if inConf.Contains("columns_mapping") {
```
jem-davies, line 58: probably need to add a linter such that only 1 columns / columns_mapping field is provided

internal/impl/gcp/processor_bigquery_select.go
```diff
@@ -77,8 +100,15 @@ func newBigQuerySelectProcessorConfig() *service.ConfigSpec {
...
 		Summary("Executes a `SELECT` query against BigQuery and replaces messages with the rows returned.").
 		Field(service.NewStringField("project").Description("GCP project where the query job will execute.")).
-		Field(service.NewStringField("table").Description("Fully-qualified BigQuery table name to query.").Example("bigquery-public-data.samples.shakespeare")).
-		Field(service.NewStringListField("columns").Description("A list of columns to query.")).
+		Field(service.NewInterpolatedStringField("table").Description("Fully-qualified BigQuery table name to query.").Example("bigquery-public-data.samples.shakespeare")).
+		Field(service.NewStringListField("columns").
+			Description("A list of columns to query.").
+			Optional()).
+		Field(service.NewBloblangField("columns_mapping").
+			Description("An optional [Bloblang mapping](/docs/guides/bloblang/about) which should evaluate to an array of column names to query.").
```
jem-davies, line 108: add better description - think about how we need to set `unsafe_dynamic_query` for this field to be evaluated.

internal/impl/gcp/processor_bigquery_select.go
```diff
@@ -118,6 +154,22 @@ pipeline:
               args_mapping: root = [ this.term ]
         result_map: |
           root.count = this.get("0.total_count")
+`,
+		).
+		Example("Unsafe Dynamic Query",
+			`
+An example to show the use of the unsafe_dynamic_query field:`,
+			`
+# {"table": "test.people", "where": "city IN (?,?,?)", "columns": ["name", "age", "city"], "args": ["London", "Paris", "Dublin"]}
```
jem-davies, line 163:
```suggestion
# {"table": "test.people", "columns": ["name", "age", "city"], "args": ["London", "Paris", "Dublin"]}
```

gregfurman (commented):
Nice one here. Some questions/feedback.

Also, can we write a test for this? Just a unit test with this new interpolation functionality would be fine.

Stretch goal though: am thinking it could be nice to add an integration test using that bigquery emulator. Check our the `setupBigQueryEmulator` function and the `TestGCPBigQueryStorageOutputWriteOK` test for reference.

internal/impl/gcp/processor_bigquery_select.go
```diff
@@ -165,34 +217,77 @@ func newBigQuerySelectProcessor(inConf *service.ParsedConfig, options *bigQueryP
...
+			}
+			cols, ok := icols.([]any)
+			if !ok {
+				msg.SetError(fmt.Errorf("col mapping returned non-array result: %T", icols))
+				continue
+			}
+
+			proc.config.queryParts.columns, err = toStringSlice(cols)
+			if err != nil {
+				msg.SetError(fmt.Errorf("%w", err))
```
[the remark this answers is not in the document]
jem-davies: taken out 😳

internal/impl/gcp/processor_bigquery_select.go
```diff
@@ -96,6 +126,12 @@ func newBigQuerySelectProcessorConfig() *service.ConfigSpec {
 		Field(service.NewStringField("suffix").
 			Description("An optional suffix to append to the select query.").
 			Optional()).
+		Field(service.NewBoolField("unsafe_dynamic_query").
+			Description("Whether to enable [interpolation functions](/docs/configuration/interpolation/#bloblang-queries) in the columns & where fields. Great care should be made to ensure your queries are defended against injection attacks.").
```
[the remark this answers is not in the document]
jem-davies: yeah - I guess my thinking around this is that I would go with these changes without this admonition, because the whole processor is marked as experimental anyway. What we should do is a new gcp_bigquery_select_v2 where we can take a more considered approach to the config fields / api. Then after that is in, add an admonition that this processor will be deprecated on a major version release.

internal/impl/gcp/processor_bigquery_select.go
jem-davies: added some checks to func bigQuerySelectProcessorConfigFromParsed()

internal/impl/gcp/processor_bigquery_select.go
jem-davies: I think that we have to use the `service.NewBloblangField` so that we are able to write a bloblang mapping to return an array of strings in a new field because we can't change the existing `columns` field to a `service.NewInterpolatedStringListField` because I think that would expect the config to look like: 

```yaml
columns: 
  - ${! this.column_1}
  - ${! this.column_2}
```

Can't find any examples that use the `service.NewInterpolatedStringListField` but it fails the yaml linter: expecting a yaml array. So it's unclear how you would handle a variable amount of columns?

gregfurman: Oh weird. That `NewInterpolatedStringListField` definitely seems like something we'd want to include. Maybe in a followup

gregfurman (approved):
LGTM! Thanks for addressing my comments. Let's look at the dynamic column names in a followup. Think it would be a useful addition.

internal/impl/gcp/processor_bigquery_select.go
jem-davies: yeah just not sure how that would work though

internal/impl/gcp/processor_bigquery_select.go
```diff
@@ -165,34 +217,77 @@ func newBigQuerySelectProcessor(inConf *service.ParsedConfig, options *bigQueryP
...
+
+	var columnsMappingExecutor *service.MessageBatchBloblangExecutor
+	if columnsMapping != nil {
+		columnsMappingExecutor = batch.BloblangExecutor(columnsMapping)
 	}
 
 	for i, msg := range batch {
 		outBatch = append(outBatch, msg)
 
+		if proc.config.unsafeDyn {
```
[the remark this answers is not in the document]
jem-davies: refactored this.

PR #203, opened 2025-01-02
feat: added interpolation to the prefix property of gcp_cloud_storage.

paulosimao-ardanlabs:
This PR adds interpolation to the prefix property of gcp_cloud_storage. Although there are no messages in context at this stage (so we have only bloblang expressions), it seems to be more ergonomic to deal with dynamic file filtering as input.

### DIFF
--- a/internal/impl/gcp/input_cloud_storage.go
+++ b/internal/impl/gcp/input_cloud_storage.go
@@ -75,7 +75,7 @@ By default Bento will use a shared credentials file when connecting to GCP servi
 			service.NewStringField(csiFieldBucket).
 				Description("The name of the bucket from which to download objects."),
 			service.NewInterpolatedStringField(csiFieldPrefix).
-				Description("An optional path prefix, if set only objects with the prefix are consumed - supports interpolation.").
+				Description("An optional path prefix, if set only objects with the prefix are consumed.").
 				Default(""),
 		).
 		Fields(codec.DeprecatedCodecFields("to_the_end")...).
--- a/internal/impl/gcp/input_cloud_storage.go
+++ b/internal/impl/gcp/input_cloud_storage.go
@@ -24,7 +24,7 @@ const (
 
 type csiConfig struct {
 	Bucket        string
-	Prefix        string
+	Prefix        *service.InterpolatedString
 	DeleteObjects bool
 	Codec         codec.DeprecatedFallbackCodec
 }
@@ -33,7 +33,7 @@ func csiConfigFromParsed(pConf *service.ParsedConfig) (conf csiConfig, err error
 	if conf.Bucket, err = pConf.FieldString(csiFieldBucket); err != nil {
 		return
 	}
-	if conf.Prefix, err = pConf.FieldString(csiFieldPrefix); err != nil {
+	if conf.Prefix, err = pConf.FieldInterpolatedString(csiFieldPrefix); err != nil {
 		return
 	}
 	if conf.Codec, err = codec.DeprecatedCodecFromParsed(pConf); err != nil {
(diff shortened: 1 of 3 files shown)


internal/impl/gcp/input_cloud_storage_test.go
```diff
@@ -0,0 +1,41 @@
...
+	conf, err := spec.ParseYAML(yamlStr, service.GlobalEnvironment())
+	require.NoError(t, err)
+
+	ret, err := csiConfigFromParsed(conf)
+	require.NoError(t, err)
+
+	return ret
+}
+
+func Test_csiConfigFromParsed(t *testing.T) {
```
jem-davies, line 31: I think that this test is misleading because here we are executing the interpolation against a populated message `{"a":1,"b":"asd"}` but in `input_cloud_storage.go` we are calling `TryString` with a zero value []byte{} for the message. So it is only possible in the input to use bloblang interpolations that don't reference message data?

internal/impl/gcp/input_cloud_storage.go
```diff
@@ -173,7 +173,13 @@ func newGCPCloudStorageTargetReader(
 		conf:   conf,
 	}
 
-	it := bucket.Objects(ctx, &storage.Query{Prefix: conf.Prefix})
+	interpolatedPrefix, err := conf.Prefix.TryString(service.NewMessage([]byte{}))
```
jem-davies, line 176: As far as I know - we don't have any input component that has a field that allows for bloblang interpolation. This is because we need a message to execute the bloblang query against, but don't have one yet because the input component is what creates them. So to get around this you have created an empty message to execute against. This means that any bloblang query cannot reference any fields of the message - that isn't a problem in of itself, but this PR doesn't make that clear in the updated documentation imho.

jem-davies (changes requested):
Think will need to have a think about allowing a bloblang interpolation on an input component, because of the issue about not having a message to execute it against.

internal/impl/gcp/input_cloud_storage.go
jem-davies: https://warpstreamlabs.github.io/bento/docs/components/inputs/azure_queue_storage#queue_name

jem-davies: so there is a precedent for this and that same approach is taken there

jem-davies:
Thanks for the PR - question: Are you able to give an example config where you are using this / what use-case does this change solve?

internal/impl/gcp/input_cloud_storage.go
```diff
@@ -74,8 +74,8 @@ By default Bento will use a shared credentials file when connecting to GCP servi
 		Fields(
 			service.NewStringField(csiFieldBucket).
 				Description("The name of the bucket from which to download objects."),
-			service.NewStringField(csiFieldPrefix).
-				Description("An optional path prefix, if set only objects with the prefix are consumed.").
+			service.NewInterpolatedStringField(csiFieldPrefix).
+				Description("An optional path prefix, if set only objects with the prefix are consumed - supports interpolation.").
```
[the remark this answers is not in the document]
paulosimao-ardanlabs: Thanks - fixed.

internal/impl/gcp/input_cloud_storage_test.go
paulosimao-ardanlabs: Thats a good point you raised. Thanks for that. I adapted the test case to use a pure bloblang expression without relying on the message. 

> So it is only possible in the input to use bloblang interpolations that don't reference message data?

This is an input component that will be used to feed the pipeline -  it sounds ok to me that we don't have messages to interpolate at this stage - If this was an intermediary processor or output, maybe this would be different.

Is there something else we should have considered (or done differently) here?

paulosimao-ardanlabs:
> Thanks for the PR - question: Are you able to give an example config where you are using this / what use-case does this change solve?

Most of the time is used for dealing with dates in the path, like this: /path/date=${!(timestamp_unix()-(2*60*60)).format_timestamp_strftime("%Y-%m-%d-%H")}

Or setting prefix based on other conditions - this could have been addressed with env vars - but the team finds it way more ergonomic, and we already have a good amount of pipelines like this.

jem-davies (approved):

PR #208, opened 2025-01-06
feat: proposing new impl of sftp input and output for evaluation. Dep…

paulosimao-ardanlabs:
…ending on feedback, it may lead to next steps.


paulosimao-ardanlabs:
Wrongly created, discarding.

PR #214, opened 2025-01-14
add an example go module with an example input plugin

jem-davies:
~~draft - going to add more types of plugins:~~

~~- processor~~
~~- output~~
~~- bloblang method / function~~

EDIT  - Just including the one example - this resource can be used as an example go module that uses plugins - the examples included [here](https://pkg.go.dev/github.com/warpstreamlabs/bento@v1.4.1/public/service#pkg-examples) are linked on the readme, not worth duplicating code + effort I think


gregfurman:
Maybe just import from the pure package? Not sure we need all plugins

jem-davies:
> Maybe just import from the pure package? Not sure we need all plugins

We just got the io + pure at the moment, maybe can get rid of io but that when has stdout for now.

resources/plugin_example/main.go
```diff
@@ -0,0 +1,27 @@
...
+
+import (
+	"context"
+
+	// import all components with:
+	// _ "github.com/warpstreamlabs/bento/public/components/all"
+
+	// or you can import select components with:
+	// _ "github.com/warpstreamlabs/bento/public/components/aws"
+	// for example.
```
gregfurman, line 11: Nice!

gregfurman (approved):
LGTM. Nice one!

PR #215, opened 2025-01-15
fix for nats_kv to be consistent in passing meta values along

kmpm:
The `nats_kv` processor removed all meta values from message on succes. That is not consistent behaviour with other operations.

- The operations `get` and `get_revision` now copy any pre existing meta values on success.
- Test `get` and `get_revision` as well as one other operation for consistency.

### DIFF
--- a/internal/impl/nats/integration_kv_test.go
+++ b/internal/impl/nats/integration_kv_test.go
@@ -151,6 +151,7 @@ cache_resources:
 			require.NoError(t, err)
 
 			m := service.NewMessage([]byte("hello"))
+			m.MetaSetMut("inherited", "world")
 			return p.Process(context.Background(), m)
 		}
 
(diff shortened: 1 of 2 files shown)


internal/impl/nats/metadata.go
```diff
@@ -26,3 +26,19 @@ func newMessageFromKVEntry(entry jetstream.KeyValueEntry) *service.Message {
...
 	return msg
 }
+
+func newMessageWithMetaFromKVEntry(entry jetstream.KeyValueEntry, src *service.Message) *service.Message {
+	msg := service.NewMessage(entry.Value())
+	//nolint:errcheck
+	src.MetaWalkMut(func(key string, value any) error {
+		msg.MetaSetMut(key, value)
+		return nil
+	})
```
gregfurman, line 36: Would probably be easier to just set the bytes of the original message to the new content. That way we can retain all the original metadata without needing to construct a new message from the KV entry.

```suggestion
	msg := src.Copy()
	msg.SetBytes(entry.Value())
```

This way would also retain the original message's context -- and therefore its tracing ID.

gregfurman (commented):
Thanks for this PR! Have just added a small suggestion.

Think for completeness we should copy the original message instead of creating a new one -- preseving the original message's metadata and context. Open to suggestions here though.

gregfurman (approved):
LGTM! Will merge when CI passes 😄

PR #217, opened 2025-01-17
chore: enable building with GOOS=windows

kmpm:
Building for windows should output an .exe file. This will enable cross compiling for windows and still using make by doing `GOOS=windows make`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,6 +11,7 @@ PATHINSTTOOLS      = $(DEST_DIR)/tools
 PATHINSTSERVERLESS = $(DEST_DIR)/serverless
 PATHINSTDOCKER     = $(DEST_DIR)/docker
 DOCKER_IMAGE       ?= ghcr.io/warpstreamlabs/bento
+GOEXE              ?= $(shell go env GOEXE)
 
 VERSION   := $(shell git describe --tags || echo "v0.0.0")
 VER_CUT   := $(shell echo $(VERSION) | cut -c2-)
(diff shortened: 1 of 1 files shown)


gregfurman:
Hey. Perhaps I am missing something but I wasn't aware cross compilation to windows was not working. Can you clarify your current setup (Go version and current OS and Arch) and what happens if you currently try to cross compile to windows? 

Also, have you tried setting the os-arch combination like `make GOOS=windows GOARCH=amd64 bento` and seeing if a `.exe` is output?

If you check out the [releases](https://github.com/warpstreamlabs/bento/releases/tag/v1.4.1) page there is also a windows bento distro.

kmpm:
Cross compiling on windows works, but not using the Makefile since the tasks expects a binary name without any extension.

This is just a thing that makes life easier when developing and want to create a windows output quick and easy.
If it's not of interest, just close the PR.

gregfurman:
> Cross compiling on windows works, but not using the Makefile since the tasks expects a binary name without any extension.

Ahh I see. I misunderstood the initial issue. I wasn't aware compiling/cross-compiling to windows adds a `.exe` by default. 

Makes sense to me to add a `.exe` when compiling for windows. We could probably simplify the changes and allow the Go compiler to just dynamically set this extension by passing the path to the target folder in the `-o` flag . i.e

```makefile
$(PATHINSTBIN)/%: $(SOURCE_FILES)
	@go build $(GO_FLAGS) -tags "$(TAGS)" -ldflags "$(LD_FLAGS) $(VER_FLAGS)" -o $@/ ./cmd/$*
```

> This is just a thing that makes life easier when developing and want to create a windows output quick and easy.
If it's not of interest, just close the PR.

Definitely appreciate your trying to improve DevEx! Most of the maintainers/contributors aren't on windows so this type of input is valuable. Also, feel free to open an issue in future if you have suggestions like this -- much nicer for discussions imo 🙂

kmpm:
I am developing on linux, not windows, but from time to time have to do tests with unreleased versions on windows machines.
What I would like to do is...
```shell
# create linux binary in `target/bin/bento`
$ make

# create windows binary in `target/bin/bento.exe`
$ GOOS=windows make
```

The thing is that the Makefile target looks for `$(PATHINSTBIN)/%` where `%` equals `bento`.
Building with your suggestion works, if `target/bin` is empty, but the resulting binary would be `target/bin/bento/bento.exe` not `target/bin/bento.exe` and the target pattern would not match the binary but the folder. 

If first building using `$ make` this will create a file called `bento` in the folder `target/bin/bento/`.
When then building for windows with `$ GOOS=windows make` it will do nothing since the target pattern will match the folder, instead of `target/bin/bento/bento.exe` and since that already exists and is up to date nothing will be done.
If one would delete all files in `target/bin/bento` but keep the folders things would still not be built for any OS, if running make again, since the target matches the folder and not the final output.
You could run `make clean` in between but it's not correct.

My solution is unfortunately ugly, that I can agree upon, but I haven't found anything better that lets the make target match the actual output.

Since I am in a quite small minority it's ok to close the PR. It's no big deal. I can work around it.

PR #219, opened 2025-01-19
feat: add go native zeromq as `zmq4n`

### LINKED ISSUE #218: ZeroMQ without cgo?
I would like to use zeromq from time to time, but I don't like having to use CGO.

Would there be any interest if I made a new set of components `zmq4n` (instead of `zmq4`) using native go and the library https://github.com/go-zeromq/zmq4 ?

kmpm:
- Uses the github.com/go-zeromq/zmq4 bindings
- Closes #218


internal/impl/zmq4n/input_zmq4n.go
```diff
@@ -0,0 +1,219 @@
...
+	case "PULL":
+		return zmq4.Pull, nil
+	}
+	return zmq4.Pull, errors.New("invalid ZMQ socket type")
+}
+
+func (z *zmqInput) Connect(ignored context.Context) (err error) {
+	if z.socket != nil {
+		return nil
+	}
```
gregfurman, line 133: Wonder if this check is necessary?

internal/impl/zmq4n/input_zmq4n.go
```diff
@@ -0,0 +1,219 @@
...
+	switch t {
+	case "SUB":
+		return zmq4.Sub, nil
+	case "PULL":
+		return zmq4.Pull, nil
+	}
+	return zmq4.Pull, errors.New("invalid ZMQ socket type")
+}
+
+func (z *zmqInput) Connect(ignored context.Context) (err error) {
```
gregfurman, line 130: Don't think this should be ignored -- pass context as a parameter `Connect(ctx content.Context)` and don't set `ctx` as background on line 140

internal/impl/zmq4n/input_zmq4n.go
```diff
@@ -0,0 +1,219 @@
...
+	if z.socket != nil {
+		return nil
+	}
+
+	t, err := getZMQInputType(z.socketType)
+	if err != nil {
+		return err
+	}
+
+	ctx := context.Background()
```
gregfurman, line 140: Mentioned above. Maybe pass as a parameter instead of set as background here.

internal/impl/zmq4n/input_zmq4n.go
```diff
@@ -0,0 +1,219 @@
...
+	default:
+		return nil, nil, errors.New("unsupported message type")
+	}
+
+	return batch, func(ctx context.Context, err error) error {
+		return nil
+	}, nil
+}
+
+// CloseAsync shuts down the zmqInput input and stops processing requests.
```
gregfurman, line 212:
```suggestion
// Close shuts down the zmqInput input and stops processing requests.
```

internal/impl/zmq4n/input_zmq4n.go
```diff
@@ -0,0 +1,219 @@
+package zmq4n
```
gregfurman, line 1: Thoughts on placing all of this into the `zeromq` package? Looks like a lot of shared components/code between these two.

Perhaps even just a `shared.go` with common config.

internal/impl/zmq4n/output_zmq4n.go
```diff
@@ -0,0 +1,191 @@
...
+	if z.socket != nil {
+		return nil
+	}
+
+	t, err := getZMQOutputType(z.socketType)
+	if err != nil {
+		return err
+	}
+
+	ctx := context.Background()
```
gregfurman, line 129: Pass in context:

```suggestion
func (z *zmqOutput) Connect(ctx context.Context) (err error) {
	if z.socket != nil {
		return nil
	}

	t, err := getZMQOutputType(z.socketType)
	if err != nil {
		return err
	}
```

internal/impl/zmq4n/integration_test.go
```diff
@@ -0,0 +1,58 @@
...
+input:
+  zmq4n:
+    urls:
+      - tcp://*:$PORT
+    bind: true
+    socket_type: $VAR2
+    sub_filters: [ $VAR3 ]
+
+`
+	suite := integration.StreamTests(
```
gregfurman, line 31: nit: We should probably extend this test suite (as well as `zeromq`) to use the [zeromq docker image](https://hub.docker.com/r/zeromq/zeromq). Not necessary for this PR but perhaps just add a TODO in this function.

gregfurman (changes requested):
Thanks for this! Have left some comments. Nothing super pressing but would like your thoughts on some implementation details.

internal/impl/zmq4n/input_zmq4n.go
kmpm: I can throw an error about not connecting while already connected if thats better.

internal/impl/zmq4n/input_zmq4n.go
kmpm: Yes a lot of code is shared but types are not... Example: zmq4.SocketType is not the same as zmq4.SocketType depending on imports. Possibly zmqInputFromConfig and zmqInputFromConfig can be shared for now but the rest just has a very similar structure and the same signatures (except from some types).
I would prefer to keep them separate. I can rename all neccesary types and functions and use the same package or keep separate packages.

internal/impl/zmq4n/integration_test.go
kmpm: That image is 9 years old and in a deprecated format. I would rather not use it.
It would be nice to validate against the native library but I haven't a reasonable approach of doing it at the moment.
TODO for now I guess

kmpm: Added, and tested, a mix of native and CGO ins/outs

internal/impl/zmq4n/input_zmq4n.go
gregfurman: Yeah an error is better.

You can also return component. ErrAlreadyStarted

internal/impl/zmq4n/integration_test.go
gregfurman: Brilliant! PR is looking great. Just going to run some manual tests and will approve if all passes.

gregfurman:
@kmpm Thanks for this! Last request... can you run `make docs` and commit the generated markdown files?

Once they're in I'm happy to approve and merge 🙂

kmpm:
Now you have the docs.
P.S. Been running the zmq4n input with SUB for days now on a Rpi5 with 4k messages per minute without issues

internal/impl/zmq4n/input_zmq4n.go
kmpm: Done in latest commit

gregfurman (approved):

PR #222, opened 2025-01-28
feat: S2 Input and Output plugins

vrongmeal:


vrongmeal:
The lint failures don't seem to be related to this change.

vrongmeal:
There seems to be some issue with the specific `golangci-lint` version on why the lint timeouts.

internal/impl/s2/output.go
```diff
@@ -0,0 +1,251 @@
...
+
+const (
+	batchingField     = "batching"
+	streamField       = "stream"
+	matchSeqNumField  = "match_seq_num"
+	fencingTokenField = "fencing_token"
+)
+
+var (
+	batchingFieldSpec = service.NewBatchPolicyField(batchingField)
```
gregfurman, line 64: nit: You can also add custom linting rules for your batching limits if you do not want to do in-line checks.

i.e

https://github.com/warpstreamlabs/bento/blob/8d108eef8c4bf1626269fb3f955a4cfec437b026/internal/impl/gcp/output_bigquery_storage.go#L64

internal/impl/s2/output.go
```diff
@@ -0,0 +1,251 @@
...
+	return nil
+}
+
+func (o *Output) WriteBatch(ctx context.Context, batch service.MessageBatch) error {
+	o.logger.Debug("Writing batch to S2")
+
+	recordBatch, err := s2.NewAppendRecordBatch()
+	if err != nil {
+		panic("empty record batch shouldn't error")
+	}
```
gregfurman, line 198: Don't think we should be panicking here. Can you rather return an error? If you want the pipeline to terminate I'd recommend returning a `service.ErrEndOfInput`.

Else, if you have more information on what the error is perhaps return a `ErrNotConnected` (to retry connection) or look into a custom error that can be handled by your `ackFn`.

gregfurman (commented):
Awesome contribution! 🚀  Have left some comments. Mostly cosmetic suggestions (feel free to ignore nits) but some (like the use of `panic`) require some feedback.

Also, is there any chance you can add an integration test? Doesn't need to be complicated but would just provide some sanity checks 🙂 See [this NATS test](https://github.com/warpstreamlabs/bento/blob/8d108eef8c4bf1626269fb3f955a4cfec437b026/internal/impl/nats/integration_stream_test.go) for an example.

internal/impl/s2/bento.go
```diff
@@ -0,0 +1,93 @@
...
+
+	return &s2bentobox.Config{
+		Basin:     basin,
+		AuthToken: authToken,
+	}, nil
+}
+
+type bentoLogger struct {
+	L *service.Logger
+}
```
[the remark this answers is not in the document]
vrongmeal: Oh that's because I need it to implement `s2bentobox.Logger` interface. The `Logger.With` method has the signature like:

```
func With(args ...any) s2bentobox.Logger
```

whereas the `*service.Logger` returns the struct. So created a wrapped logger here to pass it down.

internal/impl/s2/input.go
```diff
@@ -0,0 +1,254 @@
...
+	"github.com/warpstreamlabs/bento/public/service"
+)
+
+var (
+	//go:embed embeds/input_description.md
+	inputDescription string
+
+	//go:embed embeds/input_with_prefix_eg.yaml
+	inputWithPrefixExample string
+)
```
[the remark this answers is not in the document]
vrongmeal: `docs.go` sounds good. Will do that.

internal/impl/s2/output.go
vrongmeal: Thanks! This is really helpful

internal/impl/s2/output.go
vrongmeal: This would never error. An empty batch should always be valid. (`s2.NewAppendRecordBatch` takes batches as variadic args, not visible here).

I'll create a variant that doesn't error - `s2.NewEmptyAppendRecordBatch`. Or maybe just do `s2.NewAppendRecordBatch(records...)` and check for error then.

Also, we definitely want `service.ErrEndOfInput` here. This reminds me, there are a bunch of places where this would be helpful.

vrongmeal:
Hey @gregfurman. Thank you for your review. I'll address most of your concerns with the PR.

> Also, is there any chance you can add an integration test?

I don't think I can do this right-now. We do want to create a mock-server that we can use for testing. Will add an integration test as soon as we have that.

vrongmeal:
@gregfurman I have updated the PR.

go.mod
```diff
@@ -378,7 +385,7 @@ require (
 	modernc.org/token v1.1.0 // indirect
 )
 
-go 1.22.0
+go 1.23.4
```
gregfurman, line 388: I noticed the version of Go was being upgraded due to one of your s2 dependencies pinning `1.23.4`. 

```shell
$ go mod tidy -go=1.22
go: github.com/s2-streamstore/optr@v1.0.0 requires go@1.23.4, but 1.22 is requested
```

I cloned and downgraded this to `1.22` and all was working fine. Is there any chance this can be changed or is the higher go version necessary for that repo/library?

gregfurman (commented):
Thanks for addressing comments!

Golang lint job seemed to be OOM killed. Looks like an issue with Go `v1.23`. (from https://github.com/golangci/golangci-lint/issues/4909) and fixed in `1.63.x`. Will update the CI to a later version.

Also, can you double check on that s2 dependency requiring `1.23.4`?

go.mod
vrongmeal: Can definitely change it. `1.23.4` is just the version I created the project with :)

vrongmeal: Updated!

gregfurman: Brilliant! Thanks 🚀

gregfurman (approved):
LGTM! Thanks for the fast response time on my feedback @vrongmeal!

I'll look into creating a mock server for this plugin similar to what we do for [GCP](https://github.com/warpstreamlabs/bento/blob/8d108eef8c4bf1626269fb3f955a4cfec437b026/internal/impl/gcp/output_bigquery_test.go#L240-L264) with http handlers created on-the-fly.

Lemme know if all works fine! Else, reach out on Slack or Discord if you need any help. 

BTW, love the product! I recall seeing S2 on HackerNews a couple of months ago and am excited to see where the project goes 🚀 🙂

vrongmeal:
Thanks for your kind words!

Loved working on Bento plugins. The architecture's super extendable and everything just works so smoothly! Really glad to have S2 plugins merged.

> I'll look into creating a mock server for this plugin similar to what we do for [GCP](https://github.com/warpstreamlabs/bento/blob/8d108eef8c4bf1626269fb3f955a4cfec437b026/internal/impl/gcp/output_bigquery_test.go#L240-L264) with http handlers created on-the-fly.

This is so great! For testing Bento, I think implementing just the [StreamService](https://s2.dev/docs/grpc/api#streamservice) should suffice. There might be some changes required on the SDK side (for eg, HTTPS is enabled by default and cannot be configured). Would love to help out!

PR #224, opened 2025-02-03
add tls config options to cypher output

jem-davies:


internal/impl/cypher/output_cypher.go
```diff
@@ -211,10 +244,17 @@ func (cyp *CypherOutput) Connect(ctx context.Context) error {
 	var driver neo4j.DriverWithContext
 	var err error
 
+	var connConfigurers []func(*config.Config)
```
gregfurman, line 247: Nice one here! Think using functional options like this is nice and extendable 😎

gregfurman (approved):
LGTM!

PR #233, opened 2025-02-09
Disable strict mode in the presence of incompatible processors

gregfurman:
## Changes

Adds an atomic flag to all strict processors used to toggle whether strict mode is enabled or not. This is currently used to prevent strict mode being activated when a `try`, `catch`, or `switch` processor is present.

Future versions will better cater for these custom scenarios in strict mode, but for now, we'll favour declarative error handling over the experimental strict mode.

### DIFF
--- a/internal/bundle/strict/package.go
+++ b/internal/bundle/strict/package.go
@@ -1,0 +1,18 @@
+package strict
+
+import "slices"
+
+var incompatibleProcessors []string
+
+func init() {
+	incompatibleProcessors = []string{
+		"try",
+		"catch",
+		"switch",
+	}
+}
+
+func isProcessorIncompatible(name string) bool {
+	return slices.Contains(incompatibleProcessors, name)
+
+}
--- a/internal/bundle/strict/package.go
+++ b/internal/bundle/strict/package.go
@@ -2,17 +2,13 @@ package strict
 
 import "slices"
 
-var incompatibleProcessors []string
-
-func init() {
-	incompatibleProcessors = []string{
-		"try",
-		"catch",
-		"switch",
-	}
+var incompatibleProcessors = []string{
+	"try",
+	"catch",
+	"switch",
+	"retry",
 }
 
 func isProcessorIncompatible(name string) bool {
 	return slices.Contains(incompatibleProcessors, name)
-
 }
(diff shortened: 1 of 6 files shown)


internal/bundle/strict/package.go
```diff
@@ -0,0 +1,18 @@
+package strict
+
+import "slices"
+
+var incompatibleProcessors []string
+
+func init() {
```
jem-davies, line 7: nit: how come we have an call to init() here? Could we not declare them as package variables instead?

internal/bundle/strict/package.go
```diff
@@ -0,0 +1,18 @@
...
+	incompatibleProcessors = []string{
+		"try",
+		"catch",
+		"switch",
```
jem-davies, line 11: I added 'retry' on the https://github.com/warpstreamlabs/bento/pull/238

website/docs/components/processors/about.md
```diff
@@ -75,7 +75,9 @@ error_handling:
   strategy: reject
 ```
 
-This behavior also bypasses Bento's traditional error handling mechanisms like `catch` and `try` (described in [Error Handling][error_handling]) since the entire transaction is rejected before messages can reach error handling components.
+Note, that `try`, `catch`, and `switch` processors (described in [Error Handling][error_handling]) are currently incompatible with a global error handling strategy since the entire transaction is rejected before messages can reach error handling components.
```
jem-davies, line 78: I think that it would be good to cover all the possible _described_ error handling methods on the [error_handling](https://warpstreamlabs.github.io/bento/docs/configuration/error_handling/) page. 

After adding in the DLQ patterns and reject_errored - we are then left with: 

- [dropping a failed message](https://warpstreamlabs.github.io/bento/docs/configuration/error_handling/#drop-failed-messages)
- And then the general approach regarding `switch` in both processor pipeline + output because we could have false positives that they are used for multiplexing rather than error handling - but checking that would require some parsing of the bloblang to see if error() or errored() is being called? 

wdyt?

gregfurman: Seems like a good idea. Let's do it in a follow-up PR?

internal/bundle/strict/package.go
gregfurman: Initially made them an `init` incase we wanted some more logic to determine which processors to include. I guess we don't need it for now so will just make it a global variable (which makes me sad)

jem-davies (approved):
LGTM - happy to add the other stuff on follow up PR

Issue #216, opened 2025-01-17
panic: send on closed channel


jem-davies:
```yaml
input:
  generate:
    mapping: |
        root = {"name": fake("name"), "age": random_int(0, 95)}

output:
  fallback:
    - http_client:
        url: http://xxxx # fail this output 
        verb: POST
        retries: 3
        retry_period: 10s
        max_retry_backoff: 20s
    - stdout: {}
```

It is possible to cause a panic:

```
panic: send on closed channel

goroutine 108 [running]:
github.com/warpstreamlabs/bento/internal/impl/pure.(*fallbackBroker).loop.func3({0x10b000620, 0x14002c162d0}, {0x10af93ea0?, 0x14001d81b60?})
        /home/runner/work/bento/bento/internal/impl/pure/output_fallback.go:230 +0x134
github.com/warpstreamlabs/bento/internal/message.(*Transaction).Ack(0x140003f70e0?, {0x10b000620?, 0x14002c162d0?}, {0x10af93ea0?, 0x14001d81b60?})
        /home/runner/work/bento/bento/internal/message/transaction.go:88 +0xd8
github.com/warpstreamlabs/bento/internal/component/output.(*AsyncWriter).loop.func4()
        /home/runner/work/bento/bento/internal/component/output/async_writer.go:242 +0xdc
created by github.com/warpstreamlabs/bento/internal/component/output.(*AsyncWriter).loop in goroutine 11
        /home/runner/work/bento/bento/internal/component/output/async_writer.go:247 +0x570
```

by running the above stream - and starting the shutdown process, when the time taken to retry on the http_client output is longer than the shutdown_timeout duration. 

Presumably after the graceful shutdown time expires, Bento will start mercilessly obliterating channels, but later there will be an attempt made to send on a  closed channel - causing a panic

PR #234, opened 2025-02-11
fix: Check for shutdown before using fallback output component

jem-davies:


jem-davies:
https://github.com/warpstreamlabs/bento/issues/216

PR #249, opened 2025-02-19
small refactor for WSMessageType in http_server

jem-davies:


internal/impl/io/output_http_server.go
```diff
@@ -151,8 +149,9 @@ Please note, messages are considered delivered as soon as the data is written to
 			service.NewStringField(hsoFieldWSPath).
 				Description("The path from which websocket connections can be established.").
 				Default("/get/ws"),
-			service.NewStringField(hsoFieldWSMessageType).
+			service.NewStringEnumField(hsoFieldWSMessageType, "binary", "text").
```
gregfurman, line 152: Think we should also change that `switch` statement to not default to `binary` if not `text`

```
			var msgType int
			switch h.conf.WSMessageType {
			case "text":
				msgType = websocket.TextMessage
			default:
				msgType = websocket.BinaryMessage
			}
```

gregfurman (approved):
LGTM! Thanks for addressing

PR #251, opened 2025-02-21
doc updates

jem-davies:
- mention labels/dimensions on aws cloudwatch metric 
- use NewInterpolatedStringField for bucket & key in aws_s3 processor + note regarding delete object field
- add version tag to s2 components 
- add caution to kafka_franz re. strict ordered processing / checkpoint_limit = 1

### DIFF
--- a/website/docs/components/processors/aws_s3.md
+++ b/website/docs/components/processors/aws_s3.md
@@ -160,7 +160,7 @@ Default: `false`
 
 ### `delete_objects`
 
-Whether to delete downloaded objects from the bucket once they are processed. Note: the S3 Object will be deleted as soon as this processor has consumed the object, but the message hasn't left the Bento stream.
+Whether to delete downloaded objects from the bucket once they are processed. Note: the S3 Object will be deleted from AWS as soon as this processor has consumed the object.
 
 
 Type: `bool`  


internal/impl/kafka/input_kafka_franz.go
```diff
@@ -79,7 +79,11 @@ Finally, it's also possible to specify an explicit offset to consume from by add
 			Default("").
 			Advanced()).
 		Field(service.NewIntField("checkpoint_limit").
-			Description("Determines how many messages of the same partition can be processed in parallel before applying back pressure. When a message of a given offset is delivered to the output the offset is only allowed to be committed when all messages of prior offsets have also been delivered, this ensures at-least-once delivery guarantees. However, this mechanism also increases the likelihood of duplicates in the event of crashes or server faults, reducing the checkpoint limit will mitigate this.").
+			Description(`:::caution 
+			Setting this ` + "`checkpoint_limit = 1`" + `_will not_ enforce 'strict ordered' processing of records. Use the [kafka input processor](/docs/components/inputs/kafka/) for 'strict ordered' processing.
```
gregfurman, line 83:
```suggestion
			Setting this ` + "`checkpoint_limit: 1`" + `_will not_ enforce 'strict ordered' processing of records. Use the [kafka input processor](/docs/components/inputs/kafka/) for 'strict ordered' processing.
```

internal/impl/aws/processor_s3.go
```diff
@@ -78,15 +78,15 @@ input:
...
 			Description("The bucket to perform the GetObject operation on."),
-		service.NewStringField(s3pFieldKey).
+		service.NewInterpolatedStringField(s3pFieldKey).
 			Description("The key of the object you wish to retrive."),
 		service.NewBoolField(s3pFieldForcePathStyleURLs).
 			Description("Forces the client API to use path style URLs for downloading keys, which is often required when connecting to custom endpoints.").
 			Default(false),
 		service.NewBoolField(s3iFieldDeleteObjects).
-			Description("Whether to delete downloaded objects from the bucket once they are processed.").
+			Description("Whether to delete downloaded objects from the bucket once they are processed. Note: the S3 Object will be deleted as soon as this processor has consumed the object, but the message hasn't left the Bento stream.").
```
gregfurman, line 89: > but the message hasn't left the Bento stream.

Wdym by this? Is this saying that so long as a message is still in the stream that the object will _not_ be deleted?

So only once acked then we'd expect a delete operation to be sent

internal/impl/aws/processor_s3.go
```diff
@@ -78,15 +78,15 @@ input:
                 bucket: ${! this.1.s3BucketName }
                 key: ${! this.1.s3Key }
 `).Fields(
-		service.NewStringField(s3pFieldBucket).
+		service.NewInterpolatedStringField(s3pFieldBucket).
```
gregfurman, line 81: Is the service configured for these fields being interpolated?

internal/impl/aws/processor_s3.go
jem-davies: yeah so I think that if you were using the aws_s3 input then it would only delete the object on successful delivery to the configured output. But with the aws_s3 processor it would delete on consumption but then not delivered to the configured output?

jem-davies: Or what you have said is the opposite of what I mean here

internal/impl/aws/processor_s3.go
jem-davies: Yeah it is - it should have been like this before

PR #262, opened 2025-03-03
Add secret field to discord input

jem-davies:


gregfurman (approved):
LGTM. There are probably quite a few instances where we should tag a field as a `Secret()` but this is definitely a good start.

PR #263, opened 2025-03-04
add input cypher

jem-davies:
Adds a new input for cypher / neo4j and refactors the output component

### DIFF
--- a/internal/impl/cypher/conn_fields.go
+++ b/internal/impl/cypher/conn_fields.go
@@ -3,14 +3,13 @@ package cypher
 import "github.com/warpstreamlabs/bento/public/service"
 
 const (
-	cypherDatabase    = "database"
-	cypherURI         = "uri"
-	cypherNoAuth      = "no_auth"
-	cypherBasicAuth   = "basic_auth"
-	cypherQuery       = "query"
-	cypherValues      = "values"
-	cypherMaxInFlight = "max_in_flight"
-	cypherBatching    = "batching"
+	cypherDatabase  = "database"
+	cypherURI       = "uri"
+	cypherNoAuth    = "no_auth"
+	cypherBasicAuth = "basic_auth"
+	cypherQuery     = "query"
+	cypherValues    = "values"
+	cypherBatching  = "batching"
 
 	// Basic Auth
 	cypherUser     = "user"
--- a/internal/impl/cypher/conn_fields.go
+++ b/internal/impl/cypher/conn_fields.go
@@ -1,0 +1,59 @@
+package cypher
+
+import "github.com/warpstreamlabs/bento/public/service"
+
+const (
+	cypherDatabase    = "database"
+	cypherURI         = "uri"
+	cypherNoAuth      = "no_auth"
+	cypherBasicAuth   = "basic_auth"
+	cypherQuery       = "query"
+	cypherValues      = "values"
+	cypherMaxInFlight = "max_in_flight"
+	cypherBatching    = "batching"
+
+	// Basic Auth
+	cypherUser     = "user"
+	cypherPassword = "password"
+	cypherRealm    = "realm"
+
+	cypherTLS = "tls"
+)
+
+type CypherBasicAuth struct {
+	user     string
+	password string
+	realm    string
+}
+
+func basicAuthSpec() []*service.ConfigField {
+	return []*service.ConfigField{
+		service.NewStringField(cypherUser).
+			Default("").
+			Description("The username for basic auth."),
+		service.NewStringField(cypherPassword).
+			Default("").
+			Secret().
+			Description("The password for basic auth."),
+		service.NewStringField(cypherRealm).
+			Default("").
+			Description("The realm for basic auth."),
+	}
+}
+
+func connFields() []*service.ConfigField {
+	return []*service.ConfigField{
+		service.NewStringField(cypherDatabase).
+			Description("The name of the database to connect to.").
+			Example("neo4j"),
+		service.NewStringField(cypherURI).
+			Description("The URL of the database engine.").
+			Example("bolt://localhost:7687"),
+		service.NewBoolField(cypherNoAuth).
+			Description("Set to true to connect without authentication.").
+			Default(false),
+		service.NewObjectField(cypherBasicAuth, basicAuthSpec()...).
+			Description("Basic Authentication fields"),
+		service.NewTLSToggledField(cypherTLS),
+	}
+}
--- a/internal/impl/cypher/input_cypher.go
+++ b/internal/impl/cypher/input_cypher.go
@@ -152,24 +152,29 @@ func (cyp *CypherInput) Connect(ctx context.Context) error {
 
 	go func() {
 		session := driver.NewSession(ctx, neo4j.SessionConfig{AccessMode: neo4j.AccessModeRead})
+		cypCtx, done := cyp.shutSig.HardStopCtx(context.Background())
+
 		defer func() {
+			done()
 			close(cyp.recordsChan)
 			session.Close(ctx)
 		}()
 
-		if _, err := session.ExecuteRead(ctx,
+		if _, err := session.ExecuteRead(cypCtx,
 			func(tx neo4j.ManagedTransaction) (any, error) {
-				result, err := tx.Run(ctx, cyp.query, nil)
+				result, err := tx.Run(cypCtx, cyp.query, nil)
 				if err != nil {
 					cyp.logger.With("err", err).Error("unexpected error while executing cypher query")
 					return nil, err
 				}
 
-				for result.Next(ctx) {
+				for result.Next(cypCtx) {
 					select {
-					case <-ctx.Done():
-						return nil, ctx.Err()
+					case <-cypCtx.Done():
+						cyp.shutSig.TriggerHasStopped()
+						return nil, cypCtx.Err()
 					case <-cyp.shutSig.HardStopChan():
+						cyp.shutSig.TriggerHasStopped()
 						return nil, service.ErrEndOfInput
 					case cyp.recordsChan <- result.Record():
 					}
--- a/internal/impl/cypher/input_cypher.go
+++ b/internal/impl/cypher/input_cypher.go
@@ -148,6 +148,13 @@ func (cyp *CypherInput) Connect(ctx context.Context) error {
 		return err
 	}
 
+	if err := driver.VerifyConnectivity(ctx); err != nil {
+		return err
+	}
+	if err := driver.VerifyAuthentication(ctx, nil); err != nil {
+		return err
+	}
+
 	cyp.driver = driver
 
 	go func() {
@@ -171,10 +178,8 @@ func (cyp *CypherInput) Connect(ctx context.Context) error {
 				for result.Next(cypCtx) {
 					select {
 					case <-cypCtx.Done():
-						//cyp.shutSig.TriggerHasStopped()
 						return nil, cypCtx.Err()
 					case <-cyp.shutSig.HardStopChan():
-						//cyp.shutSig.TriggerHasStopped()
 						return nil, service.ErrEndOfInput
 					case cyp.recordsChan <- result.Record():
 					}
--- a/internal/impl/cypher/input_cypher.go
+++ b/internal/impl/cypher/input_cypher.go
@@ -153,7 +153,7 @@ func (cyp *CypherInput) Connect(ctx context.Context) error {
 	go func() {
 		defer close(cyp.recordsChan)
 
-		cyp.session.ExecuteRead(ctx,
+		if _, err := cyp.session.ExecuteRead(ctx,
 			func(tx neo4j.ManagedTransaction) (any, error) {
 				result, err := tx.Run(ctx, cyp.query, nil)
 				if err != nil {
@@ -171,7 +171,9 @@ func (cyp *CypherInput) Connect(ctx context.Context) error {
 					}
 				}
 				return nil, err
-			})
+			}); err != nil {
+			cyp.logger.With("err", err).Error("unexpected error while executing cypher query")
+		}
 	}()
 
 	return nil
--- a/internal/impl/cypher/input_cypher.go
+++ b/internal/impl/cypher/input_cypher.go
@@ -1,0 +1,201 @@
+package cypher
+
+import (
+	"context"
+	"crypto/tls"
+
+	"github.com/Jeffail/shutdown"
+	"github.com/neo4j/neo4j-go-driver/v5/neo4j"
+	"github.com/neo4j/neo4j-go-driver/v5/neo4j/config"
+	"github.com/warpstreamlabs/bento/public/service"
+)
+
+var cypherInputDescription string = `Once the records from the query are exhausted this input shuts down, allowing the pipeline to gracefully terminate (or the next input in a sequence to execute).`
+
+func cypherInputSpec() *service.ConfigSpec {
+	spec := service.NewConfigSpec().
+		Version("1.6.0").
+		Categories("Services").
+		Summary("Executes a cypher query and creates a message for each record received.").
+		Description(cypherInputDescription).
+		Fields(
+			service.NewStringField(cypherQuery).
+				Description("The cypher query to execute.").
+				Example("MATCH (n) RETURN n"),
+		)
+
+	for _, f := range connFields() {
+		spec = spec.Field(f)
+	}
+
+	spec = spec.Field(service.NewAutoRetryNacksToggleField())
+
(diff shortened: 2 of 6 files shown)


internal/impl/cypher/input_cypher.go
```diff
@@ -0,0 +1,203 @@
...
+	}
+
+	record, ok := <-cyp.recordsChan
+
+	if !ok {
+		return nil, nil, service.ErrEndOfInput
+	}
+
+	msg := service.NewMessage(nil)
+	msg.SetStructuredMut(record.AsMap())
```
gregfurman, line 194: Think we should be defensive here and do a `select` with the `Read` context:

```suggestion
	msg := service.NewMessage(nil)
	select {
	case record, open := <-cyp.recordsChan:
		if !open {
			return nil, nil, service.ErrEndOfInput
		}
		msg.SetStructuredMut(record.AsMap())
	case <-ctx.Done():
		return nil, nil, service.ErrEndOfInput
	}        	
```

Also, consider locking when we receive the channel from the struct. If there are multiple reads in parallel there's a chance we close the `cyp.recordsChan` and some other process tries to run `cyp.Read` -- which will result in us trying to read in from a closed channel and panic.

See mqtt for reference:

https://github.com/warpstreamlabs/bento/blob/16ebd504192b67b49827ece3ac68aae1d339ce5a/internal/impl/mqtt/input.go#L190-L193

internal/impl/cypher/input_cypher.go
```diff
@@ -0,0 +1,203 @@
...
+
+	return msg, func(ctx context.Context, err error) error {
+		return nil
+	}, nil
+}
```
gregfurman, line 199: question: Consider whether a `nack` should do anything

internal/impl/cypher/output_cypher.go
```diff
@@ -77,8 +48,10 @@ func cypherOutputSpec() *service.ConfigSpec {
 			service.NewIntField(cypherMaxInFlight).
 				Description("The maximum number of queries to run in parallel.").
 				Default(64),
-			service.NewTLSToggledField(cypherTLS),
 		)
+	for _, f := range connFields() {
+		spec = spec.Field(f)
+	}
```
gregfurman, line 54: nit: Why not just append these onto the spec with `Fields`?

```suggestion
		).
	  Fields(connFields()...)
```

internal/impl/cypher/conn_fields.go
```diff
@@ -0,0 +1,59 @@
...
+import "github.com/warpstreamlabs/bento/public/service"
+
+const (
+	cypherDatabase    = "database"
+	cypherURI         = "uri"
+	cypherNoAuth      = "no_auth"
+	cypherBasicAuth   = "basic_auth"
+	cypherQuery       = "query"
+	cypherValues      = "values"
+	cypherMaxInFlight = "max_in_flight"
```
gregfurman, line 12: nit: Feel free to ignore but just realised you can set max in flight with:

```go
service.NewOutputMaxInFlightField().Description("The maximum number of queries to run in parallel.").
```

and extract it via:

```go
conf.FieldMaxInFlight()
```

gregfurman (changes requested):
Good stuff on the new `input` component @jem-davies  🙌  Also, thx for the `conn_fields` refactor.

Left quite a few comments, mostly nits + questions, with some learnings from our previous misadventures 🙃 

Mostly concerned about initialisation of the session in the `Connect` and the `Read` potentially having a race condition on concurrent `Read` calls.

internal/impl/cypher/conn_fields.go
```diff
@@ -0,0 +1,59 @@
...
+			Description("The username for basic auth."),
+		service.NewStringField(cypherPassword).
+			Default("").
+			Secret().
+			Description("The password for basic auth."),
+		service.NewStringField(cypherRealm).
+			Default("").
+			Description("The realm for basic auth."),
+	}
+}
```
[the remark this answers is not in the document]
jem-davies: agree that I could have used: [BasicAuthConfig](https://github.com/warpstreamlabs/bento/blob/8764794414212266cf3baa7b4807cc0f4b2d98a2/internal/httpserver/basic_auth.go#L36) or something but think would want to look at it another time

internal/impl/cypher/input_cypher.go
jem-davies: Think that this is similar to the [sql_select input](https://github.com/warpstreamlabs/bento/blob/main/internal/impl/sql/input_sql_select.go#L286-L290) 

perhaps it would be possible to add functionality to delete a record after successful delivery to the output but afaik I don't think a user would expect ack functionality from a db read?

internal/impl/cypher/conn_fields.go
```diff
@@ -0,0 +1,59 @@
...
+	return []*service.ConfigField{
+		service.NewStringField(cypherDatabase).
+			Description("The name of the database to connect to.").
+			Example("neo4j"),
+		service.NewStringField(cypherURI).
+			Description("The URL of the database engine.").
+			Example("bolt://localhost:7687"),
+		service.NewBoolField(cypherNoAuth).
+			Description("Set to true to connect without authentication.").
+			Default(false),
```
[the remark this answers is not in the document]
jem-davies: Think that for some reason when writing the output component I went for a no_auth boolean that would make the user explicitly define they wished to connect without auth, it's default value being false. 

The above config would result in a connection without auth because no_auth is set to true - I guess that you could alter the logic such that it would connect with auth if it was set to true and basic_auth was not nil - but I think that complicates things? Removing it entirely would remove the explicitness I went for. 

Imagining a situation where someone was using a cypher component and has the above set - but was failing to connect because no_auth connections are being rejected by the db - feel they would first see that they would need to alter the no_auth value?

internal/impl/cypher/input_cypher.go
```diff
@@ -0,0 +1,203 @@
...
+	"context"
+	"crypto/tls"
+
+	"github.com/Jeffail/shutdown"
+	"github.com/neo4j/neo4j-go-driver/v5/neo4j"
+	"github.com/neo4j/neo4j-go-driver/v5/neo4j/config"
+	"github.com/warpstreamlabs/bento/public/service"
+)
+
+var cypherInputDescription string = `Once the records from the query are exhausted this input shuts down, allowing the pipeline to gracefully terminate (or the next input in a sequence to execute).`
```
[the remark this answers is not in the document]
gregfurman: Lol that's definitely correct. I meant to place it there

Maybe just add a suggestion

internal/impl/cypher/input_cypher.go
gregfurman: Yeah sure that makes sense. Just wanted to make sure this was expected behavior

internal/impl/cypher/input_cypher.go
```diff
@@ -0,0 +1,203 @@
...
+	msg.SetStructuredMut(record.AsMap())
+
+	return msg, func(ctx context.Context, err error) error {
+		return nil
+	}, nil
+}
+
+func (cyp *CypherInput) Close(ctx context.Context) (err error) {
+	return nil
+}
```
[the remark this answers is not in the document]
jem-davies: Think that we don't need to do this - it looks like the DriverWithContext will close itself, when I am running the input with the -race detector enabled and a call to driver.close() here I get a race condition ...

internal/impl/cypher/input_cypher.go
```diff
@@ -0,0 +1,203 @@
...
+	var connConfigurers []func(*config.Config)
+	if cyp.tlsConfig != nil {
+		connConfigurers = append(connConfigurers, func(config *config.Config) {
+			config.TlsConfig = cyp.tlsConfig
+		})
+	}
+
+	if cyp.noAuth {
+		driver, err = neo4j.NewDriverWithContext(cyp.uri, neo4j.NoAuth())
+	} else {
```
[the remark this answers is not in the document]
jem-davies: guess so but I just don't think it's that big of a deal

internal/impl/cypher/input_cypher.go
```diff
@@ -0,0 +1,203 @@
...
+
+	if err != nil {
+		return err
+	}
+
+	cyp.driver = driver
+	cyp.session = driver.NewSession(ctx, neo4j.SessionConfig{AccessMode: neo4j.AccessModeRead})
+
+	go func() {
+		defer close(cyp.recordsChan)
```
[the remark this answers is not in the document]
jem-davies: think that will be getting handled by the neo4j package

internal/impl/cypher/input_cypher.go
jem-davies: I think that this is fine as is, 

reads on a closed buffered channel won't result in a panic - a read on a closed buffered channel will yield the remaining values until the buffer is empty and will then yield a zero value for the channel's type. 

We are using the comma ok idiom to check if the channel is open or closed, and if it is closed we are returning appropriate values. 

we defer the closing of the channel in the go routine started by the Connect function

internal/impl/cypher/conn_fields.go
jem-davies: done

internal/impl/cypher/output_cypher.go
jem-davies: done

internal/impl/cypher/input_cypher.go
```diff
@@ -0,0 +1,203 @@
...
+	} else {
+		driver, err = neo4j.NewDriverWithContext(cyp.uri, neo4j.BasicAuth(cyp.basicAuth.user, cyp.basicAuth.password, cyp.basicAuth.realm), connConfigurers...)
+	}
+
+	if err != nil {
+		return err
+	}
+
+	cyp.driver = driver
+	cyp.session = driver.NewSession(ctx, neo4j.SessionConfig{AccessMode: neo4j.AccessModeRead})
```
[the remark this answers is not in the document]
jem-davies: Added 

```go
	if cyp.driver != nil {
		return component.ErrAlreadyStarted
	}
```

to top of Connect() 👍 

and removed the session from the struct

internal/impl/cypher/input_cypher.go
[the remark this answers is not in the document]
jem-davies: done

internal/impl/cypher/input_cypher.go
```diff
@@ -0,0 +1,207 @@
...
+	go func() {
+		session := driver.NewSession(ctx, neo4j.SessionConfig{AccessMode: neo4j.AccessModeRead})
+		defer func() {
+			close(cyp.recordsChan)
+			session.Close(ctx)
+		}()
+
+		if _, err := session.ExecuteRead(ctx,
+			func(tx neo4j.ManagedTransaction) (any, error) {
+				result, err := tx.Run(ctx, cyp.query, nil)
```
jem-davies, line 162: @gregfurman - was trying to think of a way to not have it error with context cancelled on this tx.Run in the case of a SIGTERM shutdown but wasn't able to think of anything...

jem-davies (commented):
@gregfurman could you take a quick look at the comment 🙏

internal/impl/cypher/input_cypher.go
```diff
@@ -0,0 +1,203 @@
...
+		driver, err = neo4j.NewDriverWithContext(cyp.uri, neo4j.NoAuth())
+	} else {
+		driver, err = neo4j.NewDriverWithContext(cyp.uri, neo4j.BasicAuth(cyp.basicAuth.user, cyp.basicAuth.password, cyp.basicAuth.realm), connConfigurers...)
+	}
+
+	if err != nil {
+		return err
+	}
+
+	cyp.driver = driver
```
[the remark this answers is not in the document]
jem-davies: added

internal/impl/cypher/input_cypher.go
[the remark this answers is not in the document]
gregfurman: Yeah because it's currently being used by the goroutine in the `Connect` function -- so we'd need to lock when getting and closing.

https://github.com/warpstreamlabs/bento/blob/769e158663965c90d97d3f1fc58b72bade3883bf/internal/impl/cypher/input_cypher.go#L158-L162

How else would I be able to break out of the read loop in `Connect` if `Close` is a no-op other than the case where the actual connection to cypher gets severed?

internal/impl/cypher/input_cypher.go
gregfurman: Ahh you're correct on the reading from a closed channel:

> For a channel c, the built-in function close(c) records that no more values will be sent on the channel. After calling close, and after any previously sent values have been received, receive operations will return the zero value for the channel's type without blocking.

Will have to re-look at why these other implementations, like mqtt, lock on receieving that channel. I'll address the `ctx` in another comment

internal/impl/cypher/input_cypher.go
```diff
@@ -0,0 +1,217 @@
...
+
+				for result.Next(cypCtx) {
+					select {
+					case <-cypCtx.Done():
+						return nil, cypCtx.Err()
+					case <-cyp.shutSig.HardStopChan():
+						return nil, service.ErrEndOfInput
+					case cyp.recordsChan <- result.Record():
+					}
+				}
```
gregfurman, line 186: Just a thought: how come we aren't doing this in the `Read`? Tha way, we could remove the channel in its entirety.

internal/impl/cypher/input_cypher.go
gregfurman: > Think we should be defensive here and do a select with the Read context:

Having a `select` here means we can be defensively call `Read` and allow a parent caller to cancel/return early if context is cancelled. Like, the expected/ideal behaviour of these components is that a passed in ctx can be used to determine internal behaviour.

Imagine the scenario where `for result.Next(cypCtx)` is just waiting on new data. A call to `Read` will then block so long as no new records are consumed. What if we're wanting to time out a `Read` request after 5 seconds?

internal/impl/cypher/input_cypher.go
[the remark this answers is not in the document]
gregfurman: Maybe consider adding a `shutSig.TriggerSoftStop()` or `shutSig.TriggerHardStop()` and using that to break out of the `Connect`.

See the SQL input as an example:

https://github.com/warpstreamlabs/bento/blob/113b1ae5271c5957f13ad2405a004afdea311026/internal/impl/sql/input_sql_raw.go#L238-L252

gregfurman (changes requested):

internal/impl/cypher/input_cypher.go
```diff
@@ -0,0 +1,217 @@
...
+	if err != nil {
+		return err
+	}
+
+	if err := driver.VerifyConnectivity(ctx); err != nil {
+		return err
+	}
+	if err := driver.VerifyAuthentication(ctx, nil); err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
jem-davies: https://pkg.go.dev/github.com/neo4j/neo4j-go-driver/v5@v5.24.0/neo4j#DriverWithContext.VerifyAuthentication:~:text=//%20Passing%20%60nil%60%20as%20%60auth%60%20will%20use%20the%20authentication%20information%20that%20was%20used%20to%20create%20the%20driver.

internal/impl/cypher/input_cypher.go
jem-davies: added a select in the Read() function - thanks

internal/impl/cypher/input_cypher.go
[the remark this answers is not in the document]
jem-davies: Yes good spot - this call to TriggerHardStop should be here 👍

internal/impl/cypher/input_cypher.go
jem-davies: Guess I just had a scan through the https://neo4j.com/docs/go-manual/ saw the example with the channels and thought it would make sense 

Think a couple of go routines communicating via the one channel like this isn't too bad.

internal/impl/cypher/input_cypher.go
```diff
@@ -0,0 +1,222 @@
...
+
+	return msg, func(ctx context.Context, err error) error {
+		return nil
+	}, nil
+}
+
+func (cyp *CypherInput) Close(ctx context.Context) (err error) {
+	cyp.shutSig.TriggerHardStop()
+	return nil
+}
```
gregfurman, line 222: Thx! Let's consider adding a wait on actually closing in a follow-up.

gregfurman (approved):
LGTM! Thanks for addressing my comments 🙂

PR #264, opened 2025-03-04
[SQS] Add custom HTTP headers and visibility update toggle to input

gregfurman:
## Changes
- Allows for a user to toggle whether they want to regularly update the visiblity timeout of messages with `update_visibility` -- which disables at-most-once delivery when set.
- Custom HTTP headers can be specified and added to each request to AWS with `custom_request_headers`.
- Add SQS `Attributes` to metadata.


internal/impl/aws/input_sqs.go
```diff
@@ -101,6 +116,10 @@ You can access these metadata fields using
 				Description("Whether to set the wait time. Enabling this activates long-polling. Valid values: 0 to 20.").
 				Default(0).
 				Advanced(),
+			service.NewStringMapField(sqsiFieldCustomRequestHeaders).
+				Description("An map used to send custom HTTP headers alongside each SQS operation to AWS.").
```
jem-davies, line 120: add version here?

internal/impl/aws/input_sqs.go
```diff
@@ -478,6 +505,11 @@ func addSQSMetadata(p *service.Message, sqsMsg types.Message) {
 	if rCountStr, exists := sqsMsg.Attributes["ApproximateReceiveCount"]; exists {
 		p.MetaSetMut("sqs_approximate_receive_count", rCountStr)
 	}
+
+	for k, v := range sqsMsg.Attributes {
+		p.MetaSetMut(k, v)
```
jem-davies, line 510: should this be mentioned in the docs?

jem-davies:
Could we add some tests for the new fields?

internal/impl/aws/input_sqs.go
gregfurman: The docs already say that it does this. I guess we could consider this a fix:

```
- sqs_message_id
- sqs_receipt_handle
- sqs_approximate_receive_count
- All message attributes
```

jem-davies: ah ok

jem-davies (approved):
LGTM

PR #265, opened 2025-03-05
Replace `golang.org/x/exp/rand` with `math/rand/v2`

Juneezee:
`golang.org/x/exp/rand` has been deprecated and scheduled to be deleted.

Reference: https://github.com/golang/exp/commit/f9890c6ad9f380fd3cdb66a33843f522d4790e03
Reference: https://go.dev/doc/go1.22#math_rand_v2

### DIFF
--- a/internal/impl/lang/bloblang.go
+++ b/internal/impl/lang/bloblang.go
@@ -2,9 +2,11 @@ package lang
 
 import (
 	"crypto/rand"
+	"encoding/binary"
 	"encoding/hex"
 	"fmt"
 	"io"
+	frand "math/rand/v2"
 	"strings"
 	"time"
 
@@ -12,7 +14,6 @@ import (
 	"github.com/go-faker/faker/v4"
 	"github.com/gosimple/slug"
 	"github.com/oklog/ulid"
-	frand "golang.org/x/exp/rand"
 
 	"github.com/warpstreamlabs/bento/public/bloblang"
 )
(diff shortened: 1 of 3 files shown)


internal/impl/lang/bloblang.go
```diff
@@ -271,10 +272,11 @@ func registerULID() error {
 		)
 
 	secureRandom := rand.Reader
-	fastRandom := frand.New(new(frand.LockedSource))
 	// The cast to uint64 is done on the assumption that we will not get a
 	// negative value for time.
-	fastRandom.Seed(uint64(time.Now().UnixNano()))
+	var seed [32]byte
+	binary.LittleEndian.PutUint64(seed[:], uint64(time.Now().UnixNano()))
```
gregfurman, line 278: nit: Am wondering why we are needing to even seed this since (from those release notes you added):

> The global generator accessed by top-level functions is **unconditionally randomly seeded**.

internal/impl/lang/bloblang.go
```diff
@@ -271,10 +272,11 @@ func registerULID() error {
...
+	fastRandom := frand.NewChaCha8(seed)
```
gregfurman, line 279: nit: For posterity, looks like `ChaCha8` is the shiny-new defacto algorithm used for RNG in Go. Think it's a good thing we are changing this over 👍 

From the [release notes](https://go.dev/doc/go1.22#math_rand_v2):
> [...] ChaCha8 is a new, cryptographically strong random number generator roughly similar to PCG in efficiency. ChaCha8 is the algorithm used for the top-level functions in math/rand/v2. 

Which continues:
> As of Go 1.22, `math/rand`’s top-level functions (when not explicitly seeded) and the Go runtime also use ChaCha8 for randomness.

gregfurman (approved):
LGTM! Thanks for taking the time to address this deprecation. This is a great step in preparation for `Go1.23` 🙂

PR #272, opened 2025-03-12
Disable strict mode with incompatible Bloblang functions & method

gregfurman:
## Changes

- Disables strict mode when using a Bloblang:
  - `error` or `errored` function
  - `catch` method
- The `switch` keyword/processor no longer disables strict mode.

## Testing

- Fixes `TestStrictBundleOutput` DLQ test not cleaning up the temporary file.
- Added a `TestStrictBundleBloblangFunction` and `TestDisableStrictBundleBloblangMethods` test

### DIFF
--- a/internal/bundle/strict/bundle_test.go
+++ b/internal/bundle/strict/bundle_test.go
@@ -3,6 +3,7 @@ package strict_test
 import (
 	"bytes"
 	"context"
+	"fmt"
 	"io"
 	"os"
 	"testing"
@@ -203,20 +204,20 @@ func TestStrictBundleOutput(t *testing.T) {
 	require.NoError(t, err)
 
 	// include a DLQ where errored messages go to stdout
-	err = streamBuilder.AddOutputYAML(`
+	err = streamBuilder.AddOutputYAML(fmt.Sprintf(`
 switch: 
   cases:
     - check: !errored()
       continue: true
       output:
         file:
-          path: ./tmp/data.txt
+          path: %s/data.txt
           codec: lines
     - check: errored()
       continue: false
       output:
         stdout: {}
-`)
+`, t.TempDir()))
 	require.NoError(t, err)
 
 	sendFn, err := streamBuilder.AddProducerFunc()
--- a/internal/bundle/strict/bundle_test.go
+++ b/internal/bundle/strict/bundle_test.go
@@ -272,7 +272,7 @@ processors:
 
 	mgr, err := manager.New(
 		manager.ResourceConfig{},
-		strict.OptSetRetryModeFromManager()...,
+		strict.OptSetStrictModeFromManager()...,
 	)
 	require.NoError(t, err)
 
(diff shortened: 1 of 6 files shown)


internal/bundle/strict/bundle_test.go
```diff
@@ -170,8 +175,25 @@ bloblang: root = this
...
+	require.NoError(t, res)
+
+	assert.Equal(t, 3, msgs[0].Len())
+	assert.Equal(t, "What is love?", string(msgs[0].Get(0).AsBytes()))
+	assert.Equal(t, "What is love?", string(msgs[0].Get(1).AsBytes()))
+	assert.Equal(t, "What is love?", string(msgs[0].Get(2).AsBytes()))
+
 }
 
+// FIXME: The output should be in memory, not a file.
```
jem-davies, line 196: not sure what you mean here?

internal/bundle/strict/bundle_test.go
```diff
@@ -236,3 +258,71 @@ switch:
...
+  - mapping: |
+      root = this
+  - mapping: |
+      root = if errored() { "naughty" } else { "nice" }
+`)
+	require.NoError(t, err)
+
+	mgr, err := manager.New(
+		manager.ResourceConfig{},
+		strict.OptSetRetryModeFromManager()...,
```
jem-davies, line 275: nit: it doesn't really matter but just thought that the test name and calling OptSetRetryModeFromManager confused me a little bit

public/service/stream_builder.go
```diff
@@ -854,11 +854,11 @@ func (s *StreamBuilder) BuildTraced() (*Stream, *TracingSummary, error) {
 // version releases.
 func (s *StreamBuilder) BuildStrict() (*Stream, error) {
```
jem-davies, line 855: Should we add another function BuildRetry for the other Retry Mode here?

internal/bundle/strict/bundle_test.go
gregfurman: The `output` writes to a file, which creates a new dir and doesn't clean-up after the test is run.

I think we should consider testing w. an output component that writes to something in memory as opposed to storage -- since a new file being created on each test run is unideal.

internal/bundle/strict/bundle_test.go
gregfurman: Oh yeah you're right. This should be `StrictMode` instead

public/service/stream_builder.go
gregfurman: Tbh I am a bit hesitant about exposing strict mode stuff to the public API since it's experimental. That I added this is already a bit 🤮

internal/bundle/strict/bundle_test.go
jem-davies: aight fair enough - if the test works then I don't think it should write to a file but it's not great either way I suppose.

jem-davies (approved):
looks good to me

Issue #277, opened 2025-03-20
add 'expected_headers' field to csv scanner


jem-davies:
It would be good if we could add a field to the csv scanner `expected_headers` that would be a list of strings that we expect to exist in the first row of the csv, and if not, we cancel processing the csv file.

PR #279, opened 2025-03-21
add expected_headers & expected_number_of_fields to csv scanner

jem-davies:
adds a couple of fields to the csv scanner

- expected_headers - optional list of strings to check the csv file headers against
- expected_number_of_fields - optional int that will check against the file's number of fields

### DIFF
--- a/internal/impl/pure/scanner_csv.go
+++ b/internal/impl/pure/scanner_csv.go
@@ -49,7 +49,7 @@ This scanner adds the following metadata to each message:
 				Example([]string{"first_name", "last_name", "age"}).
 				Optional(),
 			service.NewIntField(scsvFieldExpectedNumberOfFields).
-				Description("The numer of expected fields in the csv file.").
+				Description("The number of expected fields in the csv file.").
 				LintRule(`root = if this < 1 { [ "`+scsvFieldExpectedNumberOfFields+` must be at least 1" ] }`).
 				Optional(),
 		)
--- a/internal/impl/pure/scanner_csv.go
+++ b/internal/impl/pure/scanner_csv.go
@@ -5,15 +5,18 @@ import (
 	"encoding/csv"
 	"errors"
 	"io"
+	"slices"
 
 	"github.com/warpstreamlabs/bento/public/service"
 )
 
 const (
-	scsvFieldCustomDelimiter = "custom_delimiter"
-	scsvFieldParseHeaderRow  = "parse_header_row"
-	scsvFieldLazyQuotes      = "lazy_quotes"
-	scsvFieldContinueOnError = "continue_on_error"
+	scsvFieldCustomDelimiter        = "custom_delimiter"
+	scsvFieldParseHeaderRow         = "parse_header_row"
+	scsvFieldLazyQuotes             = "lazy_quotes"
+	scsvFieldContinueOnError        = "continue_on_error"
+	scsvFieldExpectedHeaders        = "expected_headers"
+	scsvFieldExpectedNumberOfFields = "expected_number_of_fields"
 )
 
 func csvScannerSpec() *service.ConfigSpec {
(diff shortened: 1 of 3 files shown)


jem-davies:
https://github.com/warpstreamlabs/bento/issues/277

jem-davies:
will raise another ticket/PR to add them into [csv input](https://warpstreamlabs.github.io/bento/docs/components/inputs/csv/)

internal/impl/pure/scanner_csv.go
```diff
@@ -41,6 +44,14 @@ This scanner adds the following metadata to each message:
 			service.NewBoolField(scsvFieldContinueOnError).
 				Description("If a row fails to parse due to any error emit an empty message marked with the error and then continue consuming subsequent rows when possible. This can sometimes be useful in situations where input data contains individual rows which are malformed. However, when a row encounters a parsing error it is impossible to guarantee that following rows are valid, as this indicates that the input data is unreliable and could potentially emit misaligned rows.").
 				Default(false),
+			service.NewStringListField(scsvFieldExpectedHeaders).
+				Description("An optional list of expected headers in the header row, that will be checked against the file contents.").
```
gregfurman, line 48: Should we mention that this will emit an error if the headers are not found?

internal/impl/pure/scanner_csv_test.go
```diff
@@ -81,3 +84,111 @@ a4,b4,c4
...
+
+	rdr, err := pConf.FieldScanner("test")
+	require.NoError(t, err)
+
+	testutil.ScannerTestSuite(t, rdr, nil, []byte(`a,b,c
+a1,b1,c1
+a2,b2,c2
+a3,b3,c3
+a4,b4,c4
+`),
```
gregfurman, line 108: mega nit: you can make these multi-line strings more readable by adding a newline and cutting it off on evaluation

```suggestion
	testutil.ScannerTestSuite(t, rdr, nil, []byte([1:]`
a,b,c
a1,b1,c1
a2,b2,c2
a3,b3,c3
a4,b4,c4
`),
```

internal/impl/pure/scanner_csv.go
```diff
@@ -86,15 +109,18 @@ func (c *csvScannerCreator) Create(rdr io.ReadCloser, aFn service.AckFunc, detai
...
 	var headers []string
 	if c.parseHeaderRow {
 		tmpHeaders, err := cRdr.Read()
 		if err != nil {
 			return nil, err
 		}
-		headers = make([]string, len(tmpHeaders))
-		_ = copy(headers, tmpHeaders)
+		if len(c.expectedHeaders) > 0 && slices.Compare(c.expectedHeaders, tmpHeaders) != 0 {
+			return nil, errors.New("expected_headers don't match file contents")
```
gregfurman, line 121: Is this supposed to be message or file contents?

```suggestion
			return nil, errors.New("expected_headers don't match message contents")
```

internal/impl/pure/scanner_csv.go
```diff
@@ -5,15 +5,18 @@ import (
...
-	scsvFieldCustomDelimiter = "custom_delimiter"
-	scsvFieldParseHeaderRow  = "parse_header_row"
-	scsvFieldLazyQuotes      = "lazy_quotes"
-	scsvFieldContinueOnError = "continue_on_error"
+	scsvFieldCustomDelimiter        = "custom_delimiter"
+	scsvFieldParseHeaderRow         = "parse_header_row"
+	scsvFieldLazyQuotes             = "lazy_quotes"
+	scsvFieldContinueOnError        = "continue_on_error"
+	scsvFieldExpectedHeaders        = "expected_headers"
+	scsvFieldExpectedNumberOfFields = "expected_number_of_fields"
```
gregfurman, line 19: Why not:

```suggestion
	scsvFieldExpectedNumberOfFields = "expected_number_of_headers"
```

IMO using different terminology (`_headers` versus `_fields`) is a bit confusing. Wdyt?

gregfurman (approved):
LGTM! Left some comments re: naming of the new fields. Unsure if we should be using `header` and `field` interchangably. 

Otherwise, a solid addition with some great test-cases as well 👍

jem-davies:
> will raise another ticket/PR to add them into [csv input](https://warpstreamlabs.github.io/bento/docs/components/inputs/csv/)

[issue](https://github.com/warpstreamlabs/bento/issues/281)

internal/impl/pure/scanner_csv.go
jem-davies: Yes I think it should be file because the expected_headers will be checked before a message is created - so it really is checking against the file contents.

internal/impl/pure/scanner_csv.go
jem-davies: Yes I thought this - but there is a slight difference - headers is referring to the first row in a csv file that would typically contain column names - with `expected_headers` we are checking the value of the column names. 

But with `expected_number_of_fields` where 'field' would refer to a value in a row, we are checking that there is the correct amount of values in a row.

So changing the 'expected_number_of_fields' to 'expected_number_of_headers' wouldn't make sense because it would still work if `parse_header_row` was set to false indicating that there aren't headers in the file. 

We could change `expected_headers` to `expected_field_names` perhaps but the other already existing fields do use the terminology 'headers' so I think it's ok as is.

internal/impl/pure/scanner_csv.go
jem-davies: updated the doc string 😄

internal/impl/pure/scanner_csv_test.go
jem-davies: Don't think that will work

jem-davies:
> LGTM! Left some comments re: naming of the new fields. Unsure if we should be using `header` and `field` interchangably.
> 
> Otherwise, a solid addition with some great test-cases as well 👍

@gregfurman thanks for the review

PR #280, opened 2025-03-25
add: kafka_franz input support for auto offset reset field

gregfurman:
## Changes

- Adds an `auto_offset_reset` field for specifying automatic offset consumption -- which takes presedence over the `start_from_oldest` field.
- Upgrades `franz-go` to `v1.18` to give us access to the `kgo.NewOffset().AtCommitted()` value when `auto_offset_reset` is `none`.
- Replaces the non-existent Docker image in the the Kafka integration tests.

### DIFF
--- a/internal/impl/kafka/input_kafka_franz.go
+++ b/internal/impl/kafka/input_kafka_franz.go
@@ -251,6 +251,30 @@ func (f *franzKafkaReader) waitForAccess(ctx context.Context, batch service.Mess
 	}
 }
 
+func getOffsetReset(conf *service.ParsedConfig) (string, error) {
+	// Allow the newer auto_offset_reset to take presedence.
+	if conf.Contains("auto_offset_reset") {
+		autoOffsetReset, err := conf.FieldString("auto_offset_reset")
+		if err != nil {
+			return "", err
+		}
+		return autoOffsetReset, nil
+	}
+
+	if conf.Contains("start_from_oldest") {
+		startFromOldest, err := conf.FieldBool("start_from_oldest")
+		if err != nil {
+			return "", err
+		}
+		if startFromOldest {
+			return "earliest", nil
+		}
+		return "latest", nil
+	}
+
+	return "earliest", nil
+}
+
 func newFranzKafkaReaderFromConfig(conf *service.ParsedConfig, res *service.Resources) (*franzKafkaReader, error) {
 	f := franzKafkaReader{
 		res:     res,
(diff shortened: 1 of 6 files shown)


internal/impl/kafka/input_kafka_franz.go
```diff
@@ -94,6 +94,13 @@ Finally, it's also possible to specify an explicit offset to consume from by add
 		Field(service.NewBoolField("start_from_oldest").
 			Description("Determines whether to consume from the oldest available offset, otherwise messages are consumed from the latest offset. The setting is applied when creating a new consumer group or the saved offset no longer exists.").
 			Default(true).
+			Optional().
+			Advanced()).
+		Field(service.NewStringEnumField("auto_offset_reset", "earliest", "latest", "none").
+			Description("Determines which offset to automatically consume from, matching Kafka's `auto.offset.reset` property. When specified, this takes precedence over `start_from_oldest`.").
+			Version("1.6.0").
+			Default("earliest").
```
jem-davies, line 102: Seeing that in a few places [here](https://learn.conduktor.io/kafka/consumer-auto-offsets-reset-behavior/#auto.offset.reset-0), [here](https://kafka.apache.org/documentation/#consumerconfigs_auto.offset.reset) that the default value is typically 'latest'? 

But I guess that it would make sense for the default to be earliest because otherwise we would be making a change to the behaviour because the 'start_from_oldest' was default to 'true'?

internal/impl/kafka/input_kafka_franz.go
```diff
@@ -94,6 +94,13 @@ Finally, it's also possible to specify an explicit offset to consume from by add
 		Field(service.NewBoolField("start_from_oldest").
 			Description("Determines whether to consume from the oldest available offset, otherwise messages are consumed from the latest offset. The setting is applied when creating a new consumer group or the saved offset no longer exists.").
 			Default(true).
+			Optional().
```
jem-davies, line 97: Think that having the `start_from_oldest` and `auto_offset_reset` fields having a default and being marked as optional isn't right. 

For example if I have a config: 

```yaml
  start_from_oldest: false
  # auto_offset_reset not explicitly set in config
```

The value of autoOffsetReset in the franzKafkaReader will end up being 'earliest' when it should be 'latest'? 

I think that the logic will work if we remove the optional() on the `start_from_oldest` field and remove the default on the `auto_offset_reset` field?

internal/impl/kafka/input_kafka_franz.go
```diff
@@ -244,6 +251,30 @@ func (f *franzKafkaReader) waitForAccess(ctx context.Context, batch service.Mess
 	}
 }
 
+func getOffsetReset(conf *service.ParsedConfig) (string, error) {
+	// Allow the newer auto_offset_reset to take presedence.
+	if conf.Contains("auto_offset_reset") {
```
jem-davies, line 256: because there is a default value for `auto_offset_reset` this boolean statement will always return `true`.

internal/impl/kafka/input_kafka_franz.go
```diff
@@ -244,6 +251,30 @@ func (f *franzKafkaReader) waitForAccess(ctx context.Context, batch service.Mess
...
+		autoOffsetReset, err := conf.FieldString("auto_offset_reset")
+		if err != nil {
+			return "", err
+		}
+		return autoOffsetReset, nil
+	}
+
+	if conf.Contains("start_from_oldest") {
```
jem-davies, line 264: same as above re. defaults and conf.Contains

jem-davies (changes requested):
Seems there is a logic error with the parsing of config and setting franzKafkaReader.autoOffsetReset

internal/impl/kafka/input_kafka_franz.go
gregfurman: Yeah exactly. I'm trying to mirror what Bento alreaydy did -- which was starting from `earliest`

internal/impl/kafka/input_kafka_franz.go
gregfurman: Yeah that makes sense. We don't really have a nice way of handling deprecating one field for another like this so am glad you spotted this logic error.

PR #284, opened 2025-03-27
feat(http): make streamHandler compatible with SSE and allow cross-origin WS upgrades

asoorm:
…igin WS upgrades

### DIFF
--- a/internal/impl/io/output_http_server.go
+++ b/internal/impl/io/output_http_server.go
@@ -492,7 +492,22 @@ func (h *httpServerOutput) wsHandler(w http.ResponseWriter, r *http.Request) {
 
 	upgrader := websocket.Upgrader{
 		CheckOrigin: func(r *http.Request) bool {
-			return true
+			if !h.conf.CORS.Enabled {
+				return true
+			}
+			origin := r.Header.Get("Origin")
+			if origin == "" {
+				return true
+			}
+			if len(h.conf.CORS.AllowedOrigins) == 0 {
+				return false
+			}
+			for _, allowed := range h.conf.CORS.AllowedOrigins {
+				if allowed == "*" || allowed == origin {
+					return true
+				}
+			}
+			return false
 		},
 	}
 
--- a/internal/impl/io/output_http_server.go
+++ b/internal/impl/io/output_http_server.go
@@ -46,21 +46,21 @@ const (
 	hsoFieldWriteWait          = "write_wait"
 	hsoFieldPongWait           = "pong_wait"
 	hsoFieldPingPeriod         = "ping_period"
-	hsoFieldStreamMode         = "stream_mode"
+	hsoFieldStreamFormat       = "stream_format"
 )
 
-type StreamMode string
+type StreamFormat string
 
 const (
-	StreamModeRaw StreamMode = "raw"
-	StreamModeSSE StreamMode = "sse"
+	StreamFormatRawBytes    StreamFormat = "raw_bytes"
+	StreamFormatEventSource StreamFormat = "event_source"
 )
 
 type hsoConfig struct {
 	Address       string
 	Path          string
 	StreamPath    string
-	StreamMode    StreamMode
+	StreamFormat  StreamFormat
 	WSPath        string
 	WSMessageType string
 	AllowedVerbs  map[string]struct{}
@@ -83,10 +83,10 @@ func hsoConfigFromParsed(pConf *service.ParsedConfig) (conf hsoConfig, err error
 	if conf.StreamPath, err = pConf.FieldString(hsoFieldStreamPath); err != nil {
 		return
 	}
-	if sMode, err := pConf.FieldString(hsoFieldStreamMode); err != nil {
+	if sFormat, err := pConf.FieldString(hsoFieldStreamFormat); err != nil {
 		return conf, err
 	} else {
-		conf.StreamMode = StreamMode(sMode)
+		conf.StreamFormat = StreamFormat(sFormat)
 	}
 	if conf.WSPath, err = pConf.FieldString(hsoFieldWSPath); err != nil {
 		return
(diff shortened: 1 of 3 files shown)


gregfurman (commented):
This looks great! Any chance you have capacity to add a test for an SSE case in `internal/impl/io/output_http_server_test.go`? Otherwise, let me know and I can also push to your branch. 

Also, since you've added a new field, can you run `make docs` to regenerate the `output_http_server` docs?

asoorm:
> This looks great! Any chance you have capacity to add a test for an SSE case in `internal/impl/io/output_http_server_test.go`? Otherwise, let me know and I can also push to your branch.
> 
> Also, since you've added a new field, can you run `make docs` to regenerate the `output_http_server` docs?

Thank you @gregfurman 🙏  - I have updated the PR accordingly - PTAL.

internal/impl/io/output_http_server.go
```diff
@@ -386,39 +412,107 @@ func (h *httpServerOutput) streamHandler(w http.ResponseWriter, r *http.Request)
...
+	}
+
 	for !h.shutSig.IsSoftStopSignalled() {
 		var ts message.Transaction
 		var open bool
+		var heartbeatChan <-chan time.Time
+
+		if heartbeatTicker != nil {
+			heartbeatChan = heartbeatTicker.C
+		}
```
gregfurman, line 455: This is re-assigned every iteration of the for loop. What about a single `heartbeatChan` that is optionally assigned to the value of a `heartbeatTicker.C` inside the above `h.conf.StreamFormat == StreamFormatEventSource && h.conf.Heartbeat > 0` check? A nil channel is never selected so it'll. effectively be a no-op.

I'll add a comment above as well.

internal/impl/io/output_http_server.go
```diff
@@ -386,39 +412,107 @@ func (h *httpServerOutput) streamHandler(w http.ResponseWriter, r *http.Request)
...
+	}
+
 	ctx, done := h.shutSig.SoftStopCtx(r.Context())
 	defer done()
 
+	var heartbeatTicker *time.Ticker
+	if h.conf.StreamFormat == StreamFormatEventSource && h.conf.Heartbeat > 0 {
+		heartbeatTicker = time.NewTicker(h.conf.Heartbeat)
+		defer heartbeatTicker.Stop()
+	}
```
gregfurman, line 446: Consider declaring the heartbeat channel once:

```suggestion
	var heartbeatChan <-chan time.Time
	if h.conf.StreamFormat == StreamFormatEventSource && h.conf.Heartbeat > 0 {
		heartbeatTicker = time.NewTicker(h.conf.Heartbeat)
		defer heartbeatTicker.Stop()
		heartbeatChan = heartbeatTicker.C
	}
```

gregfurman (approved):
LGTM! I ran a manual test (for sanity purposes) for this feature on my local and things were running smoothly, so kudos on the addition! I've just added a comment re: the heartbeat implementation. Happy to concede on it if you think I'm being pedantic

Bento config:

```yaml
input:
  generate:
    interval: 5s
    batch_size: 5
    mapping: |
      root.start = now()
      root.uuid = uuid_v4()
    auto_replay_nacks: false
  
output:
  http_server:
    stream_path: /teststream
    stream_format: event_source
    cors:
      enabled: true
      allowed_origins:
        - "*"
```

HTML file:

```html
<!doctype html>
<html>
    <body>
        <ul id="list"></ul>
    </body>

    <script type="text/javascript">
        const eventSrc = new EventSource("http://0.0.0.0:4195/teststream");

        const list = document.getElementById("list");

        eventSrc.onmessage = (event) => {
            const li = document.createElement("li");
            li.textContent = `message: ${event.data}`;

            list.appendChild(li);
        };
    </script>
</html>

```

internal/impl/io/output_http_server.go
asoorm: Great suggestion - fixed.

internal/impl/io/output_http_server_test.go
```diff
@@ -159,3 +160,426 @@ http_server:
 	h.TriggerCloseNow()
 	require.NoError(t, h.WaitForClose(ctx))
 }
+
+func TestHTTPServerOutputSSEStream(t *testing.T) {
```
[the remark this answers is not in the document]
asoorm: can we include your manual test as an md file called something like `output_http_server_test_sse_stream.md`?

asoorm:
> LGTM! I ran a manual test (for sanity purposes) for this feature on my local and things were running smoothly, so kudos on the addition! I've just added a comment re: the heartbeat implementation. Happy to concede on it if you think I'm being pedantic
> 
> Bento config:
> 
> ```yaml
> input:
>   generate:
>     interval: 5s
>     batch_size: 5
>     mapping: |
>       root.start = now()
>       root.uuid = uuid_v4()
>     auto_replay_nacks: false
>   
> output:
>   http_server:
>     stream_path: /teststream
>     stream_format: event_source
>     cors:
>       enabled: true
>       allowed_origins:
>         - "*"
> ```
> 
> HTML file:
> 
> ```
> <!doctype html>
> <html>
>     <body>
>         <ul id="list"></ul>
>     </body>
> 
>     <script type="text/javascript">
>         const eventSrc = new EventSource("http://0.0.0.0:4195/teststream");
> 
>         const list = document.getElementById("list");
> 
>         eventSrc.onmessage = (event) => {
>             const li = document.createElement("li");
>             li.textContent = `message: ${event.data}`;
> 
>             list.appendChild(li);
>         };
>     </script>
> </html>
> ```

Thank you - I've updated to optimise the heartbeat logic. 
I think that your manual test above has value to folks who want a working example, or who want to further evolve the SSE / WS logic:

Ideas off the top of my head that might be useful in the future:

- retry
- id
- custom / named events

gregfurman:
> I think that your manual test above has value to folks who want a working example, or who want to further evolve the SSE / WS logic.


OK awesome. Perhaps we can add this config as an example in the docs? It'll come up as a carousel like this [example](https://warpstreamlabs.github.io/bento/docs/components/processors/sql_select/#examples).

Your changes LGTM but I'll give this a final look a bit later today or tomorrow. If you need this in ASAP then I'm happy to merge and we can add example on a follow up.

gregfurman:
@asoorm Yeah sure I'll take this over 👍 FYI you can just run `make  docs` on your local to automatically generate docs. 

Think the changes are fine as-is so I'll merge tomorrow and add more documentation in a follow up.

PR #289, opened 2025-03-30
[SQS] Reformat SQS message attributes as metadata

gregfurman:
Reformats the message attribute data from pascal case to snake case before assigning to metadata. i.e
- `SequenceNumber => sqs_sequence_number`
- `AWSTraceHeader => sqs_aws_trace_header`


jem-davies (approved):
LGTM

Issue #295, opened 2025-04-03
sending multiple `open_message` using websocket input


jr200:
This is the same issue as seen here: [https://github.com/redpanda-data/connect/issues/1381](https://github.com/redpanda-data/connect/issues/1381)

I'm trying to connect to a websocket server that requires two messages to initiate the feed. The solution suggested in the thread is to replace `open_message` (`string`) with `open_messages` (`[]string`).

I hesitate to go down this path since websocket is quite a popular plugin and would likely break a lot of users. I have instead added an additional field `open_message_sep` (`string`), which splits up `open_message` by a separator if specified.

jr200:
On second thoughts, I went with deprecating `open_message` and adding `open_messages`. I've committed an equivalent PR to the MIT benthos repo - so hopefully the two APIs can be kept in sync for a bit longer...

PR #297, opened 2025-04-04
deprecated open_message; added open_messages, which allows multiple m…

jr200:
…essages to be sent on connect.

This is meant to resolve issue [#295](https://github.com/warpstreamlabs/bento/issues/295)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -38,7 +38,6 @@ Please always make sure a pull request has been:
 - Unit tested with `make test`
 - Linted with `make lint`
 - Formatted with `make fmt`
-- Regenerated docs with `make docs`
 
 If your change impacts inputs, outputs or other connectors then try to test them with `make test-integration`. If the integration tests aren't working on your machine then don't panic, just mention it in your PR.
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -38,6 +38,7 @@ Please always make sure a pull request has been:
 - Unit tested with `make test`
 - Linted with `make lint`
 - Formatted with `make fmt`
+- Regenerated docs with `make docs`
 
 If your change impacts inputs, outputs or other connectors then try to test them with `make test-integration`. If the integration tests aren't working on your machine then don't panic, just mention it in your PR.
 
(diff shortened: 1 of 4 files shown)


CONTRIBUTING.md
```diff
@@ -38,6 +38,7 @@ Please always make sure a pull request has been:
 - Unit tested with `make test`
 - Linted with `make lint`
 - Formatted with `make fmt`
+- Regenerated docs with `make docs`
```
jem-davies, line 41: needing to run `make docs` conditionally is mentioned below

internal/impl/io/input_websocket.go
```diff
@@ -45,6 +45,9 @@ func websocketInputSpec() *service.ConfigSpec {
 				Advanced().Optional(),
 			service.NewStringField("open_message").
 				Description("An optional message to send to the server upon connection.").
+				Advanced().Optional().Deprecated(),
+			service.NewStringListField("open_messages").
+				Description("An optional list of messages to send to the server upon connection.").
```
jem-davies, line 50: I think that because this field has depreciated `open_message` - which means it won't appear in the docs anymore it is worthwhile mentioning here.

jem-davies (commented):
Looks good - A couple of things: 

Could we add a LintRule to the Fields so a config like: 

```yaml
input:
  websocket:
    url: ws://localhost:8080/ws
    open_message: "test_three"
    open_messages: ["test", "test"]
```

would be reject because the message in `open_message` will be ignored currently? 

LintRule something like:

```
.LintRule(`root = match {this.exists("open_message") && this.open_messages != [] => "both open_message and open_messages cannot be set"}`)
```

What does this PR solve? - the old issue seems to concern some sort of handshake, but this appears to send X number of messages on connection - so am not sure it would help there (could be wrong) - Do you have a specific use-case for this?

gregfurman:
Hey @jr200 👋  Just following up on this. Are you unable to address the review?

jr200:
My apologies @gregfurman, I got side-tracked. I will do my best to get it it tomorrow.

gregfurman:
> My apologies @gregfurman, I got side-tracked. I will do my best to get it it tomorrow.

No worries at all @jr200! Given the work you've done we'd love to see this contribution get merged 😄

jr200:
> What does this PR solve? - the old issue seems to concern some sort of handshake, but this appears to send X number of messages on connection - so am not sure it would help there (could be wrong) - Do you have a specific use-case for this?

Certain websocket apis need several message exchanges before an interesting stream of data is established. e.g., polygon/bitfinex/alpaca exchanges all require an `auth` event, followed by a `subscribe`. 

The current bento (and benthos) websocket input only allows the first message (an auth), so I can't seem to subscribe to anything once i'm auth'd.

PR #304, opened 2025-04-15
fix stream stop logs

jem-davies:


gregfurman (approved):
LGTM. For future reference, we should create a multi-error handler util to more easily check if an `error` is 1 of many types imo.

jub0bs:
@gregfurman Shameless plug, but take a look at https://pkg.go.dev/github.com/jub0bs/errutil#Find, an alternative to `errors.As` that is more ergonomic, more type-safe, and more performant.

PR #305, opened 2025-04-15
Fixing protocol buffer import_paths documentation

akhenakh:
The doc says `import_paths` should be a list of directories.

The code is actually using  https://pkg.go.dev/path/filepath#Walk  

Which gives us an option to pass a file path.
It is really helpful when a protocol buffer file is mounted from a configmap in Kubernetes, if we only specify the directory with bento, many copies of it will be mounted in dot directories (for changes purpose) and the walk will hit them and halt because it found duplicates.

By specifying the full path, we can solve that issue. (I tried using recent bento)

Not sure it was intended that way, but it should be kept, it is also how protoc `-I` works.


jem-davies (approved):
LGTM - thanks 👍

PR #311, opened 2025-04-22
nats object store components

jem-davies:
- refactors / renames func kvDocs -> Docs for reuse with object store components 
- adds an input + output for nats object store 
- adds a simple processor that supports get & put operations on a nats object store
- adds an integration test for the input & output
- unit tests for the processor operations get & put


YvanDaSilva:
Hello,

Do you need some help to get this tested / reviewed ?

gregfurman:
> Hello,
> 
> 
> 
> Do you need some help to get this tested / reviewed ? 

@YvanDaSilva Thanks for the ping + offer! I'll review this within the next few hours. 

If you have capacity, you could do a manual test on your local? 👍

internal/impl/nats/processor_os.go
```diff
@@ -0,0 +1,204 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	switch osp.operation {
+	case ospOperationGet:
+		objectBytes, err := os.GetBytes(ctx, objectName)
+		if err != nil {
+			return nil, err
+		}
```
gregfurman, line 125: Can you change these all to `msg.SetError`? We shouldn't be returning an `error` in a `Process` func

internal/impl/nats/integration_os_test.go
```diff
@@ -0,0 +1,160 @@
...
+  label: object_store_input
+  nats_object_store:
+    urls:
+      - tcp://localhost:$PORT
+    bucket: bucket-$ID
+`
+
+	suite := integration.StreamTests(
+		integration.StreamTestOpenClose(),
+		// integration.StreamTestMetadata(),
```
gregfurman, line 65: Was this test failing?

internal/impl/nats/input_os.go
```diff
@@ -0,0 +1,232 @@
...
+func (osi *osInput) Read(ctx context.Context) (*service.Message, service.AckFunc, error) {
+	osi.connMut.Lock()
+	watcher := osi.watcher
+	osi.connMut.Unlock()
+
+	if watcher == nil {
+		return nil, nil, service.ErrNotConnected
+	}
+
+	for {
```
gregfurman, line 152: Is it necessary to have a `for-loop` here considering we always return on a non-nil `objectInfo`?

internal/impl/nats/input_os.go
```diff
@@ -0,0 +1,232 @@
...
+		case <-ctx.Done():
+			return nil, nil, ctx.Err()
+		}
+
+		if !open {
+			osi.disconnect()
+			return nil, nil, service.ErrNotConnected
+		}
+
+		if objectInfo == nil {
```
gregfurman, line 167: Similar to above Q: how often do we expect a `objectInfo == nil` (if ever)?

YvanDaSilva:
@jem-davies 
I will do a test first thing in the morning.
It's midnight here. I already cloned and setup up env, will let you know if it works :)

internal/impl/nats/input_os.go
```diff
@@ -0,0 +1,232 @@
...
+	if err != nil {
+		return err
+	}
+
+	osi.os, err = js.ObjectStore(ctx, osi.bucket)
+	if err != nil {
+		return err
+	}
+
+	osi.watcher, err = osi.os.Watch(ctx, nil)
```
gregfurman, line 136: suggestion: could be worth adding some support for these opts? https://pkg.go.dev/github.com/nats-io/nats.go#WatchOpt

YvanDaSilva: Commenting here because Jem mentioned he is new to nats kv.

Based on the time you have to push the PR forward, those are good to have but not must have as an "alpha" component. It also adds more testing, so up to you and how much time you have.

In addition, I checked a few places online to see if these options are used as part of the "examples, tutorials, etc." and I haven't found one where this is used.

internal/impl/nats/input_os.go
```diff
@@ -0,0 +1,232 @@
...
+			continue
+		}
+
+		if msg, err := osi.newMessageFromObjectInfo(ctx, objectInfo); err != nil {
+			return nil, nil, err
+		} else {
+			return msg, func(ctx context.Context, res error) error {
+				return nil
+			}, nil
+		}
```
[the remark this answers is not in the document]
jem-davies: altered as suggested

internal/impl/nats/processor_os.go
jem-davies: Not sure if that is right that we shouldn't be returning an error in a `Process` func - it's done [here](https://github.com/warpstreamlabs/bento/blob/8bb458b784e26049ed9beda6217782cfd6e43a01/internal/impl/awk/processor.go#L629) in the `awk` processor - for instance?

If we were doing a ProcessBatch then we would want to use msg.SetError to associate the error with a particular msg in the batch I think.

internal/impl/nats/integration_os_test.go
jem-davies: Yeah currently not implemented any handling of metadata

internal/impl/nats/input_os.go
jem-davies: I had looked at these watchOpt - but I wasn't sure what at least one of them did and then decided that I wasn't going to add them as part of this PR at least because I would need to write the test cases. Which is inherently lazy I suppose...

internal/impl/nats/input_os.go
jem-davies: I guess that yeah - it seems redundant, the implementation is very similar to the nats_kv which does the same regarding the potential nil KeyValueEntry.

internal/impl/nats/input_os.go
jem-davies: Seems that if you refactor it to remove the for loop and the check for object == nil -> you get: `panic: runtime error: invalid memory address or nil pointer dereference`, so we are seeing the nil `objectInfo`.

jem-davies: Guess that the loop is necessary to silently ignore them without throwing an error 🤔

internal/impl/nats/integration_os_test.go
gregfurman: Maybe add a TODO? Or can we delete this comment

internal/impl/nats/processor_os.go
gregfurman: Returning an error like this means the processing steps short-circuit, causing error handling (like `try` or `catch`) not to work. It's like saying "stop all processing steps since this processor failed".

> Some processors have conditions whereby they might fail. Rather than throw these messages into the abyss Bento still attempts to send these messages onwards, and has mechanisms for filtering, recovering or dead-letter queuing messages that have failed which can be read about [here](https://warpstreamlabs.github.io/bento/docs/configuration/error_handling).

I think there are valid cases for this though

internal/impl/nats/integration_os_test.go
jem-davies: deleted the comment

internal/impl/nats/input_os.go
jem-davies: seems that you get nil objectInfo - when running the integration tests ... so would think that this is necessary

internal/impl/nats/input_os.go
jem-davies: Could either add them to this PR or create a ticket for them - personally I would go with a ticket - what do you think @gregfurman ?

gregfurman (approved):
LGTM. Let's get this bad-boy in and iterate on those watch params.

Nice one, Jem!

internal/impl/nats/input_os.go
gregfurman: Yeah let's do in a follow up. Nothing blocking

PR #313, opened 2025-04-26
fix(rateLimit): fix plugin registration for custom byte-based limiters

MaudGautier:
When registering a new rate limit via the use of the plugin (`RegisterRateLimit`), the current assertion
https://github.com/warpstreamlabs/bento/blob/b2dd8d100dc4bfb234ffdbe3d78bc9cf8a5b3ef7/public/service/environment.go#L503
will always be false since it checks whether it corresponds to the **internal** MessageAwareRateLimit
https://github.com/warpstreamlabs/bento/blob/b2dd8d100dc4bfb234ffdbe3d78bc9cf8a5b3ef7/internal/component/ratelimit/message_rate_limit.go#L11-L26

Anybody using the plugin can only implement the **public** MessageAwareRateLimit
https://github.com/warpstreamlabs/bento/blob/b2dd8d100dc4bfb234ffdbe3d78bc9cf8a5b3ef7/public/service/rate_limit.go#L23-L37

Here, I add a new branch case to handle this registration via the plugin by also checking the public interface.

Note: without doing this, when registering a custom message-aware rate limit via the plugin, the `Add` method **never** gets called by the `Process` method [here](https://github.com/warpstreamlabs/bento/blob/b2dd8d100dc4bfb234ffdbe3d78bc9cf8a5b3ef7/internal/impl/pure/processor_rate_limit.go#L77-L82) since the rate limit is **not** converted to an internal MessageAwareRateLimit upon registration, but systematically to an internal.V1 instead (i.e. non-message aware).

### DIFF
--- a/public/service/environment.go
+++ b/public/service/environment.go
@@ -500,6 +500,14 @@ func (e *Environment) RegisterRateLimit(name string, spec *ConfigSpec, ctor Rate
 		}
 		// TODO: This MessageAwareRateLimit should eventually replace V1
 		// Try to upgrade to message aware rate-limiter if possible.
+
+		// When registering from an internal rate limit (local rate limit, redis rate limit)
+		if rl, ok := r.(ratelimit.MessageAwareRateLimit); ok {
+			agrl := newReverseAirGapMessageAwareRateLimit(rl)
+			return newAirGapMessageAwareRateLimit(agrl, nm.Metrics()), nil
+		}
+
+		// When registering an external rate limit via the plugin
 		if rl, ok := r.(MessageAwareRateLimit); ok {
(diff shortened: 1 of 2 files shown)


gregfurman (approved):
LGTM and great catch, @MaudGautier! This interoperability between public + internal services is confusing on a good day -- so kudos on the quick-and-easy fix!

Some context, while this `MessageAwareRateLimiter` was (at some point) supposed to replace the original rate limiter it was definitely an oversight not catering for external plugins as well :slightly_smiling_face: 

Think using generics could've helped out here for extending the interface to deal with both `service.Message` and `message.Part` in the `Add(...)` signature 🤔

PR #334, opened 2025-05-14
Log warnings for deprecated fields

jem-davies:
Currently if you run a config that contains a deprecated field or component - you will not get a warning. 

This PR adds warning logs when you create a stream with deprecated field/component when you create a stream via: 

 - bento 'normal mode' `bento -c config.yaml` 
 - bento 'streams mode' `bento streams ./*.yaml` 
 - creation of streams via REST API 
 
It also will now print any deprecated warnings when calling `bento lint` without setting `--deprecated`. The new behaviour is: 

`bento lint config.yaml` -> print any deprecated warnings but still return exit code 0
`bento lint --deprecated config.yaml` -> print deprecated warnings but return exit code 1 (same as before the PR)

 It will not change the behaviour when creating streams via the library using StreamBuilder. 
 
 ```
 INFO Running main config from specified file       @service=bento bento_version="" path=./tmp/log_deprecated_fields/deprecated.yaml
WARN Config lint warning                           @service=bento lint="tmp/log_deprecated_fields/deprecated.yaml(3,1) field url is deprecated"
INFO Listening for HTTP requests at: http://0.0.0.0:4195  @service=bento
INFO Launching a Bento instance, use CTRL+C to close  @service=bento
INFO Output type stdout is now active              @service=bento label="" path=root.output
INFO Input type amqp_1 is now active               @service=bento label="" path=root.input
 ```
 
 - [x] Add Lint Warning Logs to subcommand `bento lint`

### DIFF
--- a/config/template_examples/processor_log_and_drop.yaml
+++ b/config/template_examples/processor_log_and_drop.yaml
@@ -11,9 +11,7 @@ mapping: |
     {
       "log": {
         "level": "ERROR",
-        "fields": {
-          "content": "${! content() }"
-        },
+        "fields_mapping": "root.content = content()",
         "message": "${! error() }"
       }
     },
@@ -34,7 +32,6 @@ tests:
       catch:
         - log:
             level: ERROR
-            fields:
-              content: "${! content() }"
+            fields_mapping: root.content = content()
             message: "${! error() }"
         - bloblang: root = deleted()
--- a/internal/cli/common/reader.go
+++ b/internal/cli/common/reader.go
@@ -26,6 +26,7 @@ func ReadConfig(c *cli.Context, cliOpts *CLIOpts, streamsMode bool) (mainPath st
 		config.OptSetFullSpec(cliOpts.MainConfigSpecCtor),
 		config.OptAddOverrides(c.StringSlice("set")...),
 		config.OptTestSuffix("_bento_test"),
+		config.OptSetLintConfigWarnDeprecated(),
 	}
 	if streamsMode {
 		opts = append(opts, config.OptSetStreamPaths(c.Args().Slice()...))
--- a/internal/cli/common/service.go
+++ b/internal/cli/common/service.go
@@ -124,7 +124,7 @@ func initStreamsMode(
 	streamMgr := strmmgr.New(mgr, strmmgr.OptAPIEnabled(enableAPI))
 
 	streamConfs := map[string]stream.Config{}
-	lints, err := confReader.ReadStreams(streamConfs)
+	lints, lintWarns, err := confReader.ReadStreams(streamConfs)
 	if err != nil {
 		fmt.Fprintf(os.Stderr, "Stream configuration file read error: %v\n", err)
 		os.Exit(1)
@@ -140,6 +140,10 @@ func initStreamsMode(
 	if strict && len(lints) > 0 {
 		logger.Error(opts.ExecTemplate("Shutting down due to stream linter errors, to prevent shutdown run {{.ProductName}} with --chilled"))
 		os.Exit(1)
+	}
+
+	for _, lintWarn := range lintWarns {
+		logger.With("lint", lintWarn).Warn("Config lint warning")
 	}
 
 	for id, conf := range streamConfs {
--- a/internal/cli/common/service.go
+++ b/internal/cli/common/service.go
@@ -23,7 +23,7 @@ import (
 func RunService(c *cli.Context, cliOpts *CLIOpts, streamsMode bool) int {
 	mainPath, inferredMainPath, confReader := ReadConfig(c, cliOpts, streamsMode)
 
-	conf, pConf, lints, err := confReader.Read()
+	conf, pConf, lints, lintWarns, err := confReader.Read()
 	if err != nil {
 		fmt.Fprintf(os.Stderr, "Configuration file read error: %v\n", err)
 		return 1
@@ -58,6 +58,10 @@ func RunService(c *cli.Context, cliOpts *CLIOpts, streamsMode bool) int {
 	if strict && len(lints) > 0 {
 		logger.Error(cliOpts.ExecTemplate("Shutting down due to linter errors, to prevent shutdown run {{.ProductName}} with --chilled"))
 		return 1
+	}
+
+	for _, lintWarn := range lintWarns {
+		logger.With("lint", lintWarn).Warn("Config lint warning")
 	}
 
 	stoppableManager, err := CreateManager(c, cliOpts, logger, streamsMode, conf)
--- a/internal/cli/run.go
+++ b/internal/cli/run.go
@@ -252,7 +252,7 @@ variables have been resolved:
   {{.BinaryName}} -c ./config.yaml echo | less`)[1:],
 				Action: func(c *cli.Context) error {
 					_, _, confReader := common.ReadConfig(c, opts, false)
-					_, pConf, _, err := confReader.Read()
+					_, pConf, _, _, err := confReader.Read()
 					if err != nil {
 						fmt.Fprintf(os.Stderr, "Configuration file read error: %v\n", err)
 						os.Exit(1)
--- a/internal/config/config_test.go
+++ b/internal/config/config_test.go
@@ -20,8 +20,10 @@ import (
 	"github.com/warpstreamlabs/bento/internal/manager/mock"
 	"github.com/warpstreamlabs/bento/internal/stream"
 
+	_ "github.com/warpstreamlabs/bento/public/components/amqp1"
 	_ "github.com/warpstreamlabs/bento/public/components/io"
 	_ "github.com/warpstreamlabs/bento/public/components/pure"
+	_ "github.com/warpstreamlabs/bento/public/components/sql"
 )
 
 func testConfToAny(t testing.TB, conf any) any {
@@ -47,9 +49,10 @@ func TestSetOverridesOnNothing(t *testing.T) {
 		"output.type=drop",
 	))
 
-	conf, _, lints, _, err := rdr.Read()
+	conf, _, lints, lintWarns, err := rdr.Read()
 	require.NoError(t, err)
 	assert.Empty(t, lints)
+	assert.Empty(t, lintWarns)
 
 	v := gabs.Wrap(testConfToAny(t, conf))
 
@@ -110,9 +113,10 @@ input:
 		"output.type=drop",
 	))
 
-	conf, _, lints, _, err := rdr.Read()
+	conf, _, lints, lintWarns, err := rdr.Read()
 	require.NoError(t, err)
 	assert.Empty(t, lints)
+	assert.Empty(t, lintWarns)
 
 	v := gabs.Wrap(testConfToAny(t, conf))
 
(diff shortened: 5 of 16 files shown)


internal/config/stream_reader.go
```diff
@@ -230,6 +236,9 @@ func (r *Reader) TriggerStreamUpdate(mgr bundle.NewManagement, strict bool, path
 		mgr.Logger().Error("Rejecting updated stream %v config due to linter errors, to allow linting errors run Bento with --chilled.", info.id)
 		return noReread(errors.New("file contained linting errors and is running in strict mode"))
 	}
+	for _, lintWarn := range lintWarns {
+		lintlog.Warn(lintWarn)
+	}
```
gregfurman, line 241: Do we need to split up the lint logs here? There's already that `lintLog.Info` above:

https://github.com/warpstreamlabs/bento/blob/cdbf38a147c0a958c379766f62d2cbc3e678d34b/internal/config/stream_reader.go#L232-L234

internal/cli/common/service.go
```diff
@@ -60,6 +60,10 @@ func RunService(c *cli.Context, cliOpts *CLIOpts, streamsMode bool) int {
 		return 1
 	}
 
+	for _, lintWarn := range lintWarns {
+		logger.With("lint", lintWarn).Warn("Config lint warning")
+	}
+
```
gregfurman, line 66: Will splitting the lint errors/warnings up like this not cause the `strict := !c.Bool("chilled")` flag being set to not actually quit on a deprecated flag?

internal/config/config_test.go
```diff
@@ -238,6 +244,47 @@ cache_resources:
 	assert.Equal(t, "13s", v.S("cache_resources", "1", "memory", "default_ttl").Data())
 }
 
+func TestLintWarns(t *testing.T) {
```
gregfurman, line 247: Maybe add a test for strict mode here? Do we expect an exit or not?

internal/stream/manager/api.go
```diff
@@ -68,12 +68,18 @@ type lintErrors struct {
...
 	lConf := docs.NewLintConfig(m.manager.Environment())
 	lConf.BloblangEnv = bloblang.XWrapEnvironment(m.manager.BloblEnvironment()).Deactivated()
+	lConf.WarnDeprecated = true
 	return docs.NewLintContext(lConf)
 }
 
-func (m *Type) lintStreamConfigNode(node *yaml.Node) (lints []string) {
+func (m *Type) lintStreamConfigNode(node *yaml.Node) (lints []string, lintWarns []string) {
 	for _, dLint := range stream.Spec().LintYAML(m.lintCtx(), node) {
-		lints = append(lints, dLint.Error())
```
gregfurman, line 76: Would a warning log otherwise be sent in the `lintError` payload when creating/updating a stream?

gregfurman (commented):
Looks good. Can you clarify what happens when we are in strict mode? Will this not cause a deprecated log to not trigger an exit?

jem-davies:
> Looks good. Can you clarify what happens when we are in strict mode? Will this not cause a deprecated log to not trigger an exit?

You should see the same new behaviour of logging config lint errors whether you have set `error_handling.reject` or not. I am thinking that this logging will happen at the time of instantiation of streams - but the `error_handling` stuff pertains to errors that happen as a result of a stream pipeline?

This new behaviour of logging warnings on stream instantiation is independent of the `error_handling` stuff. 

EDIT - or are we talking about the `--chilled` cli flag when we mention strict mode?

internal/config/stream_reader.go
jem-davies: But here we are iterating over `lintWarns` not `lints`?

internal/cli/common/service.go
jem-davies: I think that this is confusing because currently (currently -> bento main branch), we don't really consider using a deprecated component/field to be a lint error.

So in the `docs` package where we are doing the linting in `LintYaml()` and `func (f FieldSpec) LintYAML()` - We do consider them to be errors if we have set LintContexts.conf.RejectDeprecated = true. However it seems that the only way to set this to be true is when using the `bento lint` subcommand. Therefore in normal operation, i.e. starting a stream with `bento -c config.yaml`: if there are deprecated components/fields in the config, no element of `lints` returned from `confReader.Read()` will pertain to deprecated components/fields. 

So this PR alters those LintYaml() funcs such we will now do: 

```go
if ctx.conf.RejectDeprecated && cSpec.Status == StatusDeprecated {
		lints = append(lints, NewLintError(node.Line, LintDeprecated, fmt.Errorf("component %v is deprecated", cSpec.Name)))
	} else if ctx.conf.WarnDeprecated && cSpec.Status == StatusDeprecated {
		lints = append(lints, NewLintWarning(node.Line, LintDeprecated, fmt.Sprintf("component %v is deprecated", cSpec.Name)))
	}
```

Where WarnDeprecated is default to `true` but we add it to `lints` as a NewLintWarning because it's later we filter them. 

We don't ever quit if there is a deprecated field in the config.

internal/config/config_test.go
jem-davies: When we are referring to `strict_mode` are we thinking the --chilled CLI option or are we talking about `error_handling.reject` ?

internal/stream/manager/api.go
jem-davies: No - tested it out: 

```sh
bento streams
```

```sh
curl http://localhost:4195/streams/bar -X POST --data-binary @- <<EOF
input:
  amqp_1:
    url: amqp://guest:guest@localhost:5672/
    source_address: foo
pipeline:
  threads: 1
  processors:
    - mapping: 'root = this.uppercase()'
output:
  stdout: {}
EOF
```

```
INFO Running without a main config file            @service=bento bento_version=1.8.2
INFO Launching Bento in streams mode, use CTRL+C to close  @service=bento
INFO Listening for HTTP requests at: http://0.0.0.0:4195  @service=bento
INFO Output type stdout is now active              @service=bento label="" path=root.output stream=bar
INFO Input type amqp_1 is now active               @service=bento label="" path=root.input stream=bar
^CINFO Received SIGINT, the service is closing       @service=bento
```

Yeah pretty sure that LintWarnings are just being universally discarded currently in Bento

internal/config/config_test.go
jem-davies: Not sure either are that relevant though to be honest

gregfurman: I mean `--chilled`

gregfurman:
> EDIT - or are we talking about the `--chilled` cli flag when we mention strict mode?

I mean when using the chilled flag.

Running the following config with `--chilled` does not suppress the deprecation warning:

```yaml
input:
  stdin:
    max_buffer: 10
```

Output:

```
$ bento -c config.yaml --chilled
INFO Running main config from specified file       @service=bento bento_version="" path=config.yaml
WARN Config lint warning                           @service=bento lint="config.yaml(3,1) field max_buffer is deprecated"
```

Also, running `bento lint config.yaml` does not seem to pick up the deprecated field either. I.e when running the config:

```yaml
input:
  stdin:
    max_buffer: 10
    foo: xyz
```

The `lint` command only returns `foo` as invalid:

```
$ bento lint config.yaml
gcp.yaml(4,1) field foo not recognised
exit status 1
```

internal/config/config_test.go
jem-davies: Yeah think that if `--chilled` is set or not - we still don't terminate the program if we get a LintWarning.

jem-davies:
> > EDIT - or are we talking about the `--chilled` cli flag when we mention strict mode?
> 
> I mean when using the chilled flag.
> 
> Running the following config with `--chilled` does not suppress the deprecation warning:
> 
> ```yaml
> input:
>   stdin:
>     max_buffer: 10
> ```
> 
> Output:
> 
> ```
> $ bento -c config.yaml --chilled
> INFO Running main config from specified file       @service=bento bento_version="" path=config.yaml
> WARN Config lint warning                           @service=bento lint="config.yaml(3,1) field max_buffer is deprecated"
> ```
> 
> Also, running `bento lint config.yaml` does not seem to pick up the deprecated field either. I.e when running the config:
> 
> ```yaml
> input:
>   stdin:
>     max_buffer: 10
>     foo: xyz
> ```
> 
> The `lint` command only returns `foo` as invalid:
> 
> ```
> $ bento lint config.yaml
> gcp.yaml(4,1) field foo not recognised
> exit status 1
> ```

Yeah I think that the `--chilled` behaviour is ok, from the cli: "--chilled continue to execute a config containing linter errors (default: false)" 

So in normal bento use, if you try to execute a config with lint errors it will terminate execution, you then have the option to override this with `--chilled`, but either way, if your config has deprecated fields (or lint warnings) it will still execute? 

Yeah we probably should add a TODO: Show Warnings when doing `bento lint`

jem-davies:
@gregfurman - added a commit to print out deprecated warnings when using the subcommand `lint` without the deprecated flag set - so that it has the same behaviour as running `bento -c config.yaml`

jem-davies:
I think that I initially didn't want to change the `bento lint` subcommand because it already has `--deprecated` (although I think that it is odd that the default is `false` here...)

config/template_examples/processor_log_and_drop.yaml
```diff
@@ -34,7 +32,6 @@ tests:
       catch:
         - log:
             level: ERROR
-            fields:
-              content: "${! content() }"
+            fields_mapping: root.content = content()
```
gregfurman, line 35: nit: Why the change here?

internal/cli/common/reader.go
```diff
@@ -26,6 +26,7 @@ func ReadConfig(c *cli.Context, cliOpts *CLIOpts, streamsMode bool) (mainPath st
 		config.OptSetFullSpec(cliOpts.MainConfigSpecCtor),
 		config.OptAddOverrides(c.StringSlice("set")...),
 		config.OptTestSuffix("_bento_test"),
+		config.OptSetLintConfigWarnDeprecated(),
```
gregfurman, line 29: 🙌  Nice and clean!

internal/cli/run.go
```diff
@@ -252,7 +252,7 @@ variables have been resolved:
   {{.BinaryName}} -c ./config.yaml echo | less`)[1:],
 				Action: func(c *cli.Context) error {
 					_, _, confReader := common.ReadConfig(c, opts, false)
-					_, pConf, _, err := confReader.Read()
+					_, pConf, _, _, err := confReader.Read()
```
gregfurman, line 255: nit: Lol we should probably give this  `Read()` signature a re-visit since returning more than 3 values (let alone 5) is quite excessive.

Can handle this in a follow-up/refactor though.

internal/config/stream_reader.go
gregfurman: Ahh I see. I was thinking that we were going to be adding deprecated lints to the lint logs. Fine to seperate them I suppose 👍

internal/stream/manager/api.go
gregfurman: Hmm... is this correct behaviour? Perhaps another ticket for the backlog (i.e handle lint warnings in streams mode)

jem-davies: In Streams mode in this PR branch you will get a warning if you start a stream with deprecated fields now. Above was current behaviour.

config/template_examples/processor_log_and_drop.yaml
jem-davies: It's just a deprecated field

internal/cli/run.go
jem-davies: Yeah agree - think the whole error / linting configs could do with a fix up it seems ...

PR #340, opened 2025-05-22
Feat: implement spanner CDC for Bento

anicoll:
~TODO: add integration tests.~ Added


go.mod
```diff
@@ -398,7 +405,9 @@ require (
 	modernc.org/token v1.1.0 // indirect
 )
 
-go 1.23.0
+go 1.24
+
+toolchain go1.24.0
```
anicoll, line 410: If this is a problem, happy to undo and push a change to keep 1.23 if required

Makefile
```diff
@@ -12,7 +12,7 @@ PATHINSTSERVERLESS = $(DEST_DIR)/serverless
 PATHINSTDOCKER     = $(DEST_DIR)/docker
 DOCKER_IMAGE       ?= ghcr.io/warpstreamlabs/bento
 
-VERSION   := $(shell git describe --tags || echo "v0.0.0")
+VERSION   := $(shell git describe --tags 2>/dev/null || echo "v0.0.0")
```
anicoll, line 15: handles failures gracefully (some reason happening locally) 🤷

jem-davies:
Hi @anicoll - thanks for the contribution! 🚀  - I should (or @gregfurman) get around to taking a look in a few days. Looks like there is just one lint check failing atm - is this something you could take a look at 🙏

anicoll:
> Hi @anicoll - thanks for the contribution! 🚀 - I should (or @gregfurman) get around to taking a look in a few days. Looks like there is just one lint check failing atm - is this something you could take a look at 🙏

I have fixed the lint issue and added some extra metadata tests to the READ func. Also made those variables.

internal/impl/gcp/input_spanner_cdc.go
```diff
@@ -0,0 +1,309 @@
...
+	if conf.HeartbeatInterval, err = pConf.FieldDuration(cdcFieldHeartbeatInterval); err != nil {
+		return
+	}
+
+	return
+}
+
+func spannerCdcSpec() *service.ConfigSpec {
+	return service.NewConfigSpec().
+		Stable().
```
jem-davies:
```suggestion
		Beta().
		Version("1.8.0").
```

maybe we add it as a Beta component for now - or even remove it entirely so that it is then 'experimental'? 
and add version tag 1.8.0

internal/impl/gcp/input_spanner_cdc.go
```diff
@@ -0,0 +1,309 @@
...
+		Categories("Services", "GCP").
+		Summary(`Consumes spanner Change Stream Events from a GCP Spanner instance.`).
```
jem-davies:
```suggestion
		Summary(`Consumes Spanner Change Stream Events from a GCP Spanner instance.`).
```

nit: should we capitalise Spanner here?

internal/impl/gcp/input_spanner_cdc.go
```diff
@@ -0,0 +1,309 @@
...
+		Description(`
+For information on how to set up credentials check out [this guide](https://cloud.google.com/docs/authentication/production).
+
+This Input uses [screamer](https://github.com/anicoll/screamer) for the reading and tracking of partitions within spanner.\
+Currently does not support Postgresql Dialect for the Spanner CDC.\
+It does support multiple runners using a distributed lock to ensure that only one runner is reading from a partition at a time.
```
jem-davies:
```suggestion
This input uses [screamer](https://github.com/anicoll/screamer) for the reading and tracking of partitions within Spanner.
Currently, it does not support PostgreSQL Dialect for the Spanner CDC.
It supports multiple runners using a distributed lock to ensure that only one runner reads from a given partition at a time.
```

internal/impl/gcp/input_spanner_cdc.go
```diff
@@ -0,0 +1,309 @@
...
+
+### Event Data Structure
+The data structure of the events emitted by this input can be found here:
+* [google](https://cloud.google.com/spanner/docs/change-streams/details#data-change-records).
+* [go structure](https://pkg.go.dev/github.com/anicoll/screamer#DataChangeRecord).
```
jem-davies:
```suggestion
* [google](https://cloud.google.com/spanner/docs/change-streams/details#data-change-records)
* [go structure](https://pkg.go.dev/github.com/anicoll/screamer#DataChangeRecord)
```

nit: get rid of the full stops at the end

internal/impl/gcp/input_spanner_cdc.go
```diff
@@ -0,0 +1,309 @@
...
+	cdcFieldSpannerMetadataTable = "metadata_table"
+	cdcFieldSpannerMetadataDSN   = "spanner_metadata_dsn"
+	cdcFieldStreamName           = "stream_name"
+	cdcFieldStartTime            = "start_time"
+	cdcFieldEndTime              = "end_time"
+	cdcFieldHeartbeatInterval    = "heartbeat_interval"
+
+	metadataTimestamp   = "gcp_spanner_commit_timestamp"
+	metadataModType     = "gcp_spanner_cdc_mod_type"
+	metadataTableName   = "gcp_spanner_table_name"
```
jem-davies, line 35:
```suggestion
	metadataTimestamp   = "gcp_spanner_cdc_commit_timestamp"
	metadataModType     = "gcp_spanner_cdc_mod_type"
	metadataTableName   = "gcp_spanner_cdc_table_name"
```

Feel I like the idea of the component name being the a sort of prefix for the metadata key values - what do you reckon? Just thinking about name collisions in future/general 🤔

internal/impl/gcp/input_spanner_cdc.go
```diff
@@ -0,0 +1,309 @@
...
+### Metadata
+
+This input adds the following metadata fields to each message:
+
+`+"``` text"+`
+- gcp_spanner_commit_timestamp - The time the records were committed in spanner.
+- gcp_spanner_cdc_mod_type - The type of modification that occurred (INSERT, UPDATE, DELETE).
+- gcp_spanner_table_name - The name of the table that was modified.
+- gcp_spanner_cdc_server_transaction_id - The server transaction ID of the change.
+- gcp_spanner_cdc_record_sequence - The sequence number of the record in the change stream.
```
jem-davies, line 121: Will need to update these if the names change - or even use the concatenate the const's?

internal/impl/gcp/input_spanner_cdc.go
```diff
@@ -0,0 +1,309 @@
...
+				Description("The name of the stream to track changes on."),
+			service.NewDurationField(cdcFieldHeartbeatInterval).
+				Description("An optional field to configure the heartbeat interval for partitions.").
+				Default((time.Second * 3).String()).Optional(),
+			service.NewStringField(cdcFieldStartTime).
+				Description("An optional field to define the start point to read from the changestreams, for details on valid start times please see [this document](https://cloud.google.com/spanner/docs/change-streams#data-retention)").
+				Example(time.RFC3339).
+				Optional(),
+			service.NewStringField(cdcFieldEndTime).
+				Description("An optional field to define the end time to read from the changestreams").
```
jem-davies: Would be good to add a mention of RFC3339 in the description here I think - and in the start time field

Makefile
jem-davies: _not sure_ about this change. 

What is the motivation behind suppressing errors here? Why would we want to do: 

```
~ ❯ git describe --tags 2>/dev/null  || echo "v0.0.0" 
v0.0.0
```

over 

```
~ ❯ git describe --tags || echo "v0.0.0" 
fatal: not a git repository (or any of the parent directories): .git
v0.0.0
```

It doesn't seem that this change pertains to the implementation of the spanner cdc input - so am thinking we should raise another issue ticket and fix in a separate PR. Feel that this change might impact some of the github actions or something ... and would rather consider it separately to spanner CDC changes.

jem-davies (changes requested):
Pretty much just doc changes / nits on this review, I have taken a look at the implementation and mostly looks good but want to check again regarding the use of contexts - how does this handle shutdown gracefully etc.  

Thanks

Makefile
anicoll: fair enough. I will revert this

internal/impl/gcp/input_spanner_cdc.go
anicoll: Reason I never did _cdc on some of the fields was because I didnt feel they were cdc specific.
Such as a table name doesnt pertain to the CDC aprt. nor does the commit timestamp which is just a database thing rather than a CDC thing.

Will await for your preference before making any changes

jem-davies: keep it as it is i think yeah

jem-davies (changes requested):
I think that there is a problem with gracefully shutting down this component? 

I set up a `gcr.io/cloud-spanner-emulator/emulator` similar to the integration test then manually sent SIGTERM to the bento process and got the logs in the log.txt file I have attached. 

[logs.txt](https://github.com/user-attachments/files/20618342/logs.txt)

Also I think that the connect logic doesn't retry because I did get this once as well: 
```
go run main.go -c ./tmp/0605_gcp_spanner_cdc_test/config.yaml
INFO Running main config from specified file       @service=bento bento_version="" path=./tmp/0605_gcp_spanner_cdc_test/config.yaml
INFO Listening for HTTP requests at: http://0.0.0.0:4195  @service=bento
ERRO Service closing due to: failed to init input <no label> path root.input: rpc error: code = FailedPrecondition desc = Schema change operation rejected because a concurrent schema change operation or read-write transaction is already in progress.  @service=bento
exit status 1
```

I think normally we put most of the connection logic in the Connect function of the component then the bento engine will retry to connect indefinitely rather than terminate? 

There is also a small fix needed for the tests at the moment.

anicoll:
@jem-davies 
I have fixed said tests (apologies) and please see attached screenshot of the handling of a sigterm now.
image

If you want me to alter the log levels to anything specific please let me know!


I tested it like so
`
func main() {
	ctx := context.Background()

	go func() {
		time.Sleep(15 * time.Second)
		p, err := os.FindProcess(os.Getpid())
		if err != nil {
			panic(err)
		}
		if err := p.Signal(syscall.SIGTERM); err != nil {
			panic(err)
		}
	}()

	service.RunCLI(ctx)
}
`

Any more feedback is welcome..

anicoll:
This PR resolves #341

anicoll:
@jem-davies ping

jem-davies:
> @jem-davies ping


👋 @anicoll  - haven't forgotten about - just working through a few other things on the todo list then will come back to this 

If it's a blocker for you - I would suggest looking at implementing as a plugin.

jem-davies:
I think that the issue regarding graceful shutdown is still an issue, was going to get round to taking a proper look and seeing if I could give more advice but currently taking a look at some other things.

internal/impl/gcp/input_spanner_cdc.go
```diff
@@ -0,0 +1,329 @@
...
+// Connect initializes the connection to the Spanner change stream
+func (c *gcpSpannerCDCInput) Connect(ctx context.Context) error {
+	c.cdcMut.Lock()
+	defer c.cdcMut.Unlock()
+	if c.subscriber == nil {
+		return service.ErrNotConnected
+	}
+
+	subCtx, cancel := context.WithCancel(context.Background())
+	c.closeFunc = cancel
```
gregfurman: I think we should replace managing and storing `ctx` like this with a `shutdown.Signaller` type. This is used in many of our components where reading/writing is done async and requires some shutt down coordination i.e

https://github.com/warpstreamlabs/bento/blob/477a8266d0008f1fbe7a1b9518f2b6cc35c21951/internal/impl/aws/input_sqs.go#L182

Then we can create a `ctx` that will be cancelled when we call shutdown on the `shutdown.Signaller` i.e

https://github.com/warpstreamlabs/bento/blob/477a8266d0008f1fbe7a1b9518f2b6cc35c21951/internal/impl/aws/input_sqs.go#L286-L287

internal/impl/gcp/input_spanner_cdc.go
```diff
@@ -0,0 +1,329 @@
...
+// consumer implements the message consumption interface
+type consumer struct {
+	msgQueue chan []byte // Channel for queuing received messages
+}
+
+// Consume implements the message consumption callback
+func (c consumer) Consume(data []byte) error {
+	c.msgQueue <- data
+	return nil
+}
```
gregfurman: This `Consume` call will block if the queue is full given the channel is buffered with a capacity of `1000`:

https://github.com/warpstreamlabs/bento/blob/4a3121112e21a58920609cb09630533a61f08e96/internal/impl/gcp/input_spanner_cdc.go#L237-L239

Is this expected behaviour? Am concerned the channel fills up, blocks some calls, and a context is cancelled due to timing out (while the read is blocked).

internal/impl/gcp/input_spanner_cdc.go
```diff
@@ -0,0 +1,329 @@
...
+	case data, open = <-c.consumer.msgQueue:
+	case <-ctx.Done():
+		// If context is cancelled, ensure we call closeFunc to clean up resources
+		c.cdcMut.Lock()
+		if c.closeFunc != nil {
+			c.closeFunc()
+		}
+		c.cdcMut.Unlock()
+		c.log.Error("Context cancelled while waiting for data :" + ctx.Err().Error())
+		return nil, nil, service.ErrNotConnected
```
gregfurman: If a `Read` is cancelled early (due to a timeout etc.) I'm not sure we want to be re-triggering a `Connect` loop. What if the caller just wants to cancel a request early?

internal/impl/gcp/input_spanner_cdc.go
```diff
@@ -0,0 +1,329 @@
...
+	}
+	if !open {
+		return nil, nil, service.ErrNotConnected
+	}
```
gregfurman: suggestion: Move this check inside the `select-case` above at:

https://github.com/warpstreamlabs/bento/blob/4a3121112e21a58920609cb09630533a61f08e96/internal/impl/gcp/input_spanner_cdc.go#L281

internal/impl/gcp/input_spanner_cdc.go
```diff
@@ -0,0 +1,329 @@
...
+	if conf.HeartbeatInterval, err = pConf.FieldDuration(cdcFieldHeartbeatInterval); err != nil {
+		return
+	}
+
+	return
+}
+
+func spannerCdcSpec() *service.ConfigSpec {
+	return service.NewConfigSpec().
+		Beta().
```
jem-davies, line 101:
```suggestion
```

I think we should remove this so that the component will be marked as 'experimental'.

internal/impl/gcp/input_spanner_cdc.go
```diff
@@ -0,0 +1,329 @@
...
+	}, nil
+}
+
+// Connect initializes the connection to the Spanner change stream
+func (c *gcpSpannerCDCInput) Connect(ctx context.Context) error {
+	c.cdcMut.Lock()
+	defer c.cdcMut.Unlock()
+	if c.subscriber == nil {
+		return service.ErrNotConnected
+	}
```
jem-davies: I think the issue with this is:  if c.subscriber == nil - then we won't have successfully executed the constructor `func newGcpSpannerCDCInput` but if we return a `service.ErrNotConnected` then we will retry the `func Connect`?

internal/impl/gcp/input_spanner_cdc.go
```diff
@@ -0,0 +1,329 @@
...
+			}
+		} else {
+			c.log.Errorf("Subscription error: %v\n", rerr)
+		}
+
+		c.cdcMut.Lock()
+		close(c.consumer.msgQueue)
+		c.closeFunc = nil
+		c.cdcMut.Unlock()
+	}()
```
jem-davies: We need to return the errors from the connect function - so if the .Subscribe fails we should be returning that error. 
The issue I am finding is that the .Subscribe func doesn't seem to fail even if there is no chance it will connect i.e. 

Incorrectly setting the env var: `SPANNER_EMULATOR_HOST=localhost:9015` it will just hang: 

```
INFO Running main config from specified file       @service=bento bento_version="" path=./tmp/0807_gcp_spanner_cdc/config.yaml
INFO Listening for HTTP requests at: http://0.0.0.0:4195  @service=bento
``` 

or if I cause a connection error by setting the wrong dsn - I am getting panics: 

```
ERRO API error during subscription: rpc error: code = NotFound desc = Database not found: projects/project-id/instances/instance-id/databases/stream-databases
error details: name = ResourceInfo type = type.googleapis.com/google.spanner.admin.database.v1.Database resourcename = projects/project-id/instances/instance-id/databases/stream-databases owner =  desc = Database does not exist.  @service=bento label="" path=root.input
panic: Subscriber has already started subscribe.
```

What would be handy is if we had a way of testing the connection prior to calling subscribe - like the `cypher` input: 

```go

	if err := driver.VerifyConnectivity(ctx); err != nil {
		return err
	}
	if err := driver.VerifyAuthentication(ctx, nil); err != nil {
		return err
	}
```

we then return an error from Connect then Bento will go into a retry loop trying to reconnect.

internal/impl/gcp/integration_spanner_cdc_test.go
```diff
@@ -0,0 +1,343 @@
...
+		_, err = s.client.Apply(readCtx, []*spanner.Mutation{
+			spanner.InsertOrUpdate(tableName, []string{"Int64", "String"}, []interface{}{40, "Daniel"}),
+		})
+		s.NoError(err)
+
+		// Attempt to read with a context that will be cancelled
+		_, _, err = input.Read(readCtx)
+
+		// Verify we got a context cancellation error
+		s.ErrorIs(err, service.ErrNotConnected, "Expected context.Canceled error, got: %v", err)
```
jem-davies: I think that this should be a context.Cancelled error not a ErrNotConnected

jem-davies (changes requested):
As Greg pointed out re. the shutdown - I think that there is changes to be made regarding initial connection as it doesn't currently follow bento semantics as is

internal/impl/gcp/input_spanner_cdc.go
```diff
@@ -0,0 +1,419 @@
...
+		return
+	}
+
+	return
+}
+
+func spannerCdcSpec() *service.ConfigSpec {
+	return service.NewConfigSpec().
+		Beta().
+		Version("1.8.0").
```
gregfurman:
```suggestion
		Version("1.11.0").
```

internal/impl/gcp/input_spanner_cdc.go
```diff
@@ -0,0 +1,419 @@
...
+	cdcMut       sync.RWMutex    // RWMutex for thread-safe operations
+	log          *service.Logger // Logger instance
+
+	recordsCh chan changeRecord // Channel that holds all records
+
+	partitionTokens map[string]struct{} // Track processed partition tokens
+	partitionLock   sync.RWMutex        // Mutex for thread-safe operations
+
+	shutdownSig *shutdown.Signaller // Signals to begin shutting down components
+}
```
jem-davies:
```suggestion
type gcpSpannerCDCInput struct {
	conf         cdcConfig
	streamClient *spanner.Client
	cdcMut       sync.RWMutex
	log          *service.Logger

	recordsCh chan changeRecord

	partitionTokens map[string]struct{}
	partitionLock   sync.RWMutex

	shutdownSig *shutdown.Signaller
}
```

nit

internal/impl/gcp/input_spanner_cdc.go
```diff
@@ -0,0 +1,419 @@
...
+	if c.recordsCh == nil {
+		c.recordsCh = make(chan changeRecord)
+	}
+
+	go c.loop()
+
+	return nil
+}
+
+// Read retrieves the next message from the change stream
```
jem-davies:
```suggestion
func newGcpSpannerCDCInput(conf cdcConfig, res *service.Resources) (*gcpSpannerCDCInput, error) {
	return &gcpSpannerCDCInput{
		conf:            conf,
		log:             res.Logger(),
		shutdownSig:     shutdown.NewSignaller(),
		recordsCh:       make(chan changeRecord),
		partitionTokens: make(map[string]struct{}),
	}, nil
}

func (c *gcpSpannerCDCInput) Connect(ctx context.Context) error {
	c.cdcMut.Lock()
	defer c.cdcMut.Unlock()

	if c.streamClient != nil {
		return nil
	}

	select {
	case <-c.shutdownSig.HasStoppedChan():
		return service.ErrEndOfInput
	default:
	}

	var err error
	c.streamClient, err = spanner.NewClient(ctx, c.conf.SpannerDSN)
	if err != nil {
		return err
	}

	if c.recordsCh == nil {
		c.recordsCh = make(chan changeRecord)
	}

	go c.loop()

	return nil
}

```

nit - remove comments that just repeat the function name

internal/impl/gcp/input_spanner_cdc.go
```diff
@@ -0,0 +1,419 @@
...
+				Description("The dsn for spanner from where to read the changestream.").
+				Example("projects/{projectId}/instances/{instanceId}/databases/{databaseName}"),
+			service.NewStringField(cdcFieldStreamName).
+				Description("The name of the stream to track changes on."),
+			service.NewDurationField(cdcFieldHeartbeatInterval).
+				Description("An optional field to configure the heartbeat interval for partitions.").
+				Default((time.Second * 3).String()).
+				Optional(),
+			service.NewStringField(cdcFieldStartTime).
+				Description("An optional field to define the start point to read from the changestreams, timestamp format should conform to RFC3339, for details on valid start times please see [this document](https://cloud.google.com/spanner/docs/change-streams#data-retention)").
```
jem-davies: Add note regarding you will get the time.Now() if not set

internal/impl/gcp/input_spanner_cdc.go
```diff
@@ -0,0 +1,419 @@
...
+	}
+
+	select {
+	case <-c.shutdownSig.HasStoppedChan():
+		return service.ErrEndOfInput
+	default:
+	}
+
+	var err error
+	c.streamClient, err = spanner.NewClient(ctx, c.conf.SpannerDSN)
```
jem-davies: We will need to test this client actually connects and if not, return a service.ErrNotConnected otherwise Bento will just hang after reporting:

```
INFO Input type gcp_spanner_cdc is now active      @service=bento label="" path=root.input
```

internal/impl/gcp/input_spanner_cdc.go
```diff
@@ -0,0 +1,419 @@
...
+	modType string
+}
+
+// newGcpSpannerCDCInput creates a new Spanner CDC input instance
+func newGcpSpannerCDCInput(conf cdcConfig, res *service.Resources) (*gcpSpannerCDCInput, error) {
+	return &gcpSpannerCDCInput{
+		conf:            conf,
+		log:             res.Logger(),
+		shutdownSig:     shutdown.NewSignaller(),
+		recordsCh:       make(chan changeRecord),
```
jem-davies: hmm thinking that a buffered channel would be better?

jem-davies (dismissed):
Think that the shutdown of the stream isn't handled properly, getting the following: 

ERRO Error in readPartition: spanner: code = "Canceled", desc = "context canceled", requestID = "1.bce9f9cad806f2d4.1.4.7.1"  @service=bento label="" path=root.input stream=config

and: 

```
2025/08/25 16:33:26 Number of sessions in use is negative, resetting it to currSessionsCheckedOutLocked. Stack trace: goroutine 123 [running]:
```

Edit - I did implement shutdown on: https://github.com/jem-davies/bento/tree/feat/spanner_cdc_jem (but that was for the old client)

internal/impl/gcp/input_spanner_cdc.go
gregfurman: The `NewClient` should be creating a grpc connection to the DSN. If it can't connect, because it's unauthenticated or whatever, then this `NewClient` will fail. 

Are you suggesting we do a `healthcheck` or equivalent after connecting to gurantee that the client is _actually_ able to connect?

internal/impl/gcp/input_spanner_cdc.go
gregfurman: OK sure. How large would we want it? If the channel is full, I assume we then just skip new reads?

internal/impl/gcp/input_spanner_cdc.go
jem-davies: I am not sure what will cause `spanner.NewClient(ctx, c.conf.SpannerDSN)` to error - if I run a config that has no chance of connecting because the DSN doesn't exist, it will exit Connect() without any problem. 

 ```
 INFO Input type gcp_spanner_cdc is now active      @service=bento label="" path=root.input
ERRO Error in readPartition: spanner: code = "PermissionDenied", desc = "Cloud Spanner API has not been used in project 998832733962 before or it is disabled. Enable it by visiting https://console.developers.google.com/apis/api/spanner.googleapis.com/overview?project=998832733962 then retry. If you enabled this API recently, wait a few minutes for the action to propagate to our systems and retry.", requestID = "1.32b109eff64b885d.1.4.3.1"  @service=bento label="" path=root.input
 ```
 
 So I think we need to have a healthcheck in the connect function

internal/impl/gcp/input_spanner_cdc.go
jem-davies: hmm not sure 100-1000 - we could expose via the config?

jem-davies: If the buffered channel is full then the default behaviour is to just block - just having a buffered channel will decouple the input from the downstream process

gregfurman: Changed to buffered channel 😄 Thanks!

internal/impl/gcp/input_spanner_cdc.go
```diff
@@ -0,0 +1,436 @@
...
+	defer cancel()
+
+	errgrp, _ := errgroup.WithContext(ctx)
+	errgrp.Go(func() error {
+		return c.readPartition(ctx, errgrp, nil, c.conf.StartTime)
+	})
+
+	if err := errgrp.Wait(); err != nil {
+		if !errors.Is(err, context.Canceled) && spanner.ErrCode(err) != codes.Canceled {
+			c.log.Errorf("Error readin partition: %v", err)
```
jem-davies:
```suggestion
			c.log.Errorf("Error reading partition: %v", err)
```

internal/impl/gcp/input_spanner_cdc.go
jem-davies: Added a few commits to address this one

PR #343, opened 2025-05-27
Set unique id & timestamp to each flow

sananguliyev:
In the current build of the traced stream, tracing events are currently grouped only by section. This setup makes it difficult to link these events together and understand the complete message journey in the stream flow. The new implementation addresses this issue by:

* Assigning a unique flow ID to each message journey
* Adding a timestamp to each tracing event to provide a clearer order of the steps the message went through
* Introducing `EventsByFlowID`, which organizes events by flow and arranges them in chronological order based on timestamp.

### DIFF
--- a/internal/bloblang/query/functions.go
+++ b/internal/bloblang/query/functions.go
@@ -263,6 +263,39 @@ var _ = registerSimpleFunction(
 	},
 )
 
+var _ = registerSimpleFunction(
+	NewFunctionSpec(
+		FunctionCategoryMessage, "flow_id",
+		"Provides the message flow id used for tracing the journey of a message through the pipeline. If no flow id exists, one will be created and stored in the message metadata.",
+		NewExampleSpec("",
+			`meta flow_id = flow_id()`,
+		),
+	).Experimental(),
+	func(fCtx FunctionContext) (any, error) {
+		part := fCtx.MsgBatch.Get(fCtx.Index)
+
+		// Check if we already have a flow ID in metadata
+		if flowID, exists := part.MetaGetMut("_bento_flow_id"); exists {
+			if flowIDStr, ok := flowID.(string); ok && flowIDStr != "" {
+				return flowIDStr, nil
+			}
+		}
+
+		// Try to use OpenTelemetry trace ID if available
+		if traceID := tracing.GetTraceID(part); traceID != "" && traceID != "00000000000000000000000000000000" {
+			// Store it in metadata for future use
+			part.MetaSetMut("_bento_flow_id", traceID)
+			return traceID, nil
+		}
+
+		// Generate a new flow ID - we'll use a simple approach here
+		// In a real implementation, you might want to use UUID or similar
+		flowID := fmt.Sprintf("%d_%d", time.Now().UnixNano(), rand.Int63())
+		part.MetaSetMut("_bento_flow_id", flowID)
+		return flowID, nil
+	},
+)
+
 //------------------------------------------------------------------------------
 
 var _ = registerFunction(
--- a/internal/bundle/tracing/events.go
+++ b/internal/bundle/tracing/events.go
@@ -1,10 +1,13 @@
 package tracing
 
 import (
+	"strconv"
 	"sync"
 	"sync/atomic"
+	"time"
 
 	"github.com/warpstreamlabs/bento/internal/message"
+	"github.com/warpstreamlabs/bento/internal/tracing"
 )
 
 // EventType describes the type of event a component might experience during
@@ -21,9 +24,11 @@ var (
 
 // NodeEvent represents a single event that occurred within the stream.
 type NodeEvent struct {
-	Type    EventType
-	Content string
-	Meta    map[string]any
+	Type      EventType
+	Content   string
+	Meta      map[string]any
+	FlowID    string    // Unique identifier for the message flow journey
+	Timestamp time.Time // When the event occurred
 }
 
 // EventProduceOf creates a produce event from a message part.
@@ -35,9 +40,11 @@ func EventProduceOf(part *message.Part) NodeEvent {
 	})
 
 	return NodeEvent{
-		Type:    EventProduce,
-		Content: string(part.AsBytes()),
-		Meta:    meta,
+		Type:      EventProduce,
+		Content:   string(part.AsBytes()),
+		Meta:      meta,
+		FlowID:    getOrCreateFlowID(part),
+		Timestamp: time.Now(),
 	}
 }
 
@@ -50,25 +57,83 @@ func EventConsumeOf(part *message.Part) NodeEvent {
 	})
 
 	return NodeEvent{
-		Type:    EventConsume,
-		Content: string(part.AsBytes()),
-		Meta:    meta,
+		Type:      EventConsume,
+		Content:   string(part.AsBytes()),
+		Meta:      meta,
+		FlowID:    getOrCreateFlowID(part),
+		Timestamp: time.Now(),
 	}
 }
 
 // EventDeleteOf creates a deleted event from a message part.
 func EventDeleteOf() NodeEvent {
 	return NodeEvent{
-		Type: EventDelete,
+		Type:      EventDelete,
+		FlowID:    "", // No part available, so no flow ID
+		Timestamp: time.Now(),
+	}
+}
+
+// EventDeleteOfPart creates a deleted event from a message part with flow ID.
+func EventDeleteOfPart(part *message.Part) NodeEvent {
+	return NodeEvent{
+		Type:      EventDelete,
+		FlowID:    getOrCreateFlowID(part),
+		Timestamp: time.Now(),
 	}
 }
 
 // EventErrorOf creates an error event from a message part.
 func EventErrorOf(err error) NodeEvent {
 	return NodeEvent{
-		Type:    EventError,
-		Content: err.Error(),
+		Type:      EventError,
+		Content:   err.Error(),
+		FlowID:    "", // No part available, so no flow ID
+		Timestamp: time.Now(),
 	}
+}
+
+// EventErrorOfPart creates an error event from a message part with flow ID.
+func EventErrorOfPart(part *message.Part, err error) NodeEvent {
+	return NodeEvent{
+		Type:      EventError,
+		Content:   err.Error(),
+		FlowID:    getOrCreateFlowID(part),
+		Timestamp: time.Now(),
+	}
+}
+
+// getOrCreateFlowID retrieves or creates a flow ID for a message part.
+// It first tries to get an existing flow ID from metadata, then from OpenTelemetry trace ID,
+// and finally generates a new one if neither exists.
+func getOrCreateFlowID(part *message.Part) string {
+	// First, check if we already have a flow ID in metadata
+	if flowID, exists := part.MetaGetMut("_bento_flow_id"); exists {
+		if flowIDStr, ok := flowID.(string); ok && flowIDStr != "" {
+			return flowIDStr
+		}
+	}
+
+	// Try to use OpenTelemetry trace ID if available
+	if traceID := tracing.GetTraceID(part); traceID != "" && traceID != "00000000000000000000000000000000" {
+		// Store it in metadata for future use
+		part.MetaSetMut("_bento_flow_id", traceID)
+		return traceID
+	}
+
+	// Generate a new flow ID using atomic counter for simplicity and performance
+	flowID := generateFlowID()
+	part.MetaSetMut("_bento_flow_id", flowID)
+	return flowID
+}
+
+// Flow ID counter for generating unique IDs
+var flowIDCounter uint64
+
+// generateFlowID creates a new unique flow ID using an atomic counter
+func generateFlowID() string {
+	id := atomic.AddUint64(&flowIDCounter, 1)
+	return strconv.FormatUint(id, 10)
 }
 
 type control struct {
--- a/internal/bundle/tracing/processor.go
+++ b/internal/bundle/tracing/processor.go
@@ -32,14 +32,21 @@ func (t *tracedProcessor) ProcessBatch(ctx context.Context, m message.Batch) ([]
 		return t.wrapped.ProcessBatch(ctx, m)
 	}
 
+	// Track input parts and their flow IDs for better correlation
+	inputParts := make([]*message.Part, m.Len())
 	prevErrs := make([]error, m.Len())
 	_ = m.Iter(func(i int, part *message.Part) error {
+		inputParts[i] = part
 		t.e.Add(EventConsumeOf(part))
 		prevErrs[i] = part.ErrorGet()
 		return nil
 	})
 
 	outMsgs, res := t.wrapped.ProcessBatch(ctx, m)
+
+	// Track which input parts produced outputs
+	hasOutput := make([]bool, len(inputParts))
+
 	for _, outMsg := range outMsgs {
 		_ = outMsg.Iter(func(i int, part *message.Part) error {
 			t.e.Add(EventProduceOf(part))
@@ -52,13 +59,25 @@ func (t *tracedProcessor) ProcessBatch(ctx context.Context, m message.Batch) ([]
 				return nil
 			}
 			_ = atomic.AddUint64(t.errCtr, 1)
-			t.e.Add(EventErrorOf(fail))
+			t.e.Add(EventErrorOfPart(part, fail))
 			return nil
 		})
+
+		// Mark that we have output for tracking deletions
+		if outMsg.Len() > 0 {
+			for i := 0; i < len(hasOutput) && i < outMsg.Len(); i++ {
+				hasOutput[i] = true
+			}
+		}
 	}
+
+	// If no output messages, record delete events for each input part
 	if len(outMsgs) == 0 {
-		// TODO: Find a better way of locating deletes (using batch index tracking).
-		t.e.Add(EventDeleteOf())
+		for _, part := range inputParts {
+			if part != nil {
+				t.e.Add(EventDeleteOfPart(part))
+			}
+		}
 	}
 
 	return outMsgs, res
--- a/internal/bundle/tracing/bundle_test.go
+++ b/internal/bundle/tracing/bundle_test.go
@@ -540,7 +540,14 @@ meta bar = "new bar value"
 
 	type tMap = map[string]any
 
-	assert.Equal(t, []tracing.NodeEvent{
+	// Check that all events have flow IDs and timestamps
+	for i, event := range events {
+		assert.NotEmpty(t, event.FlowID, "Event %d should have a flow ID", i)
+		assert.False(t, event.Timestamp.IsZero(), "Event %d should have a timestamp", i)
+	}
+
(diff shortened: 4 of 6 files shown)


sananguliyev:
@jem-davies @gregfurman
Could you please take a look when you have time?

gregfurman:
@sananguliyev Hey! Thanks for the contribution and apologies for the delay. Will take a look this eve 😄

internal/bundle/tracing/events.go
```diff
@@ -50,25 +57,83 @@ func EventConsumeOf(part *message.Part) NodeEvent {
...
+func getOrCreateFlowID(part *message.Part) string {
+	// First, check if we already have a flow ID in metadata
+	if flowID, exists := part.MetaGetMut("_bento_flow_id"); exists {
+		if flowIDStr, ok := flowID.(string); ok && flowIDStr != "" {
+			return flowIDStr
+		}
+	}
+
+	// Try to use OpenTelemetry trace ID if available
+	if traceID := tracing.GetTraceID(part); traceID != "" && traceID != "00000000000000000000000000000000" {
```
gregfurman: Can we not rely soley on the TraceID as the `FlowID`? Or are you trying to account for the case where a span wasn't initialised (I'm no expert in OTEL so pardon my ignorance)?

internal/bundle/tracing/events.go
```diff
@@ -50,25 +57,83 @@ func EventConsumeOf(part *message.Part) NodeEvent {
...
+// getOrCreateFlowID retrieves or creates a flow ID for a message part.
+// It first tries to get an existing flow ID from metadata, then from OpenTelemetry trace ID,
+// and finally generates a new one if neither exists.
+func getOrCreateFlowID(part *message.Part) string {
+	// First, check if we already have a flow ID in metadata
+	if flowID, exists := part.MetaGetMut("_bento_flow_id"); exists {
+		if flowIDStr, ok := flowID.(string); ok && flowIDStr != "" {
+			return flowIDStr
+		}
+	}
```
gregfurman: I'm not sure storing this in the message metadata is the best approach. For example, there are instances where metadata is dumped to a JSON body, sent as headers, and sometimes even fully reset/cleared. Also, the fact that metadata is mutable means a user could hypothetically overwrite this data and lose it.

If we are going to be storing this info, what're your thoughts on perhaps rather using the message context as a store (similar to how OTEL does it)?

i.e

```golang
type flowTraceKey struct{} // store with an internal struct as a key to avoid collisions/overwrites

func WithFlowD(ctx context.Context, flowID string) context.Context {
    return context.WithValue(ctx, traceKey{}, traceID)
}

func SetFlowID(p *message.Part) string {
	ctx := message.GetContext(p)
        // V7 UUIDs are sortable since they're time ordered
        // hence no real need for timestamps as well 
        // (see https://pkg.go.dev/github.com/google/uuid#NewV7)
        flowID, _ := uuid.NewV7() 
	return context.WithValue(ctx, traceKey{}, flowID.String())
}


func GetFlowID(ctx context.Context) string {
    if id, ok := ctx.Value(traceKey{}).(string); ok {
        return id
    }
    return ""
}
```

gregfurman (commented):
Some questions relating to the metadata approach. Think better internal tracing is a brilliant idea so I'm keen to hear your thoughts on my comments 😄

internal/bundle/tracing/events.go
sananguliyev: The purpose here is to reuse something if exists. If the user enabled tracing then message has its own ID and here I wanted to use it instead of generating new one, and it's also better in order to link trace and internal events if someone wants to use flow ID for some purposes.

internal/bundle/tracing/events.go
sananguliyev: TBH, initially I did not plan to add it as bloblang method and wanted to make it only accessible from tracing summary api, but then just decided to add it in order to make it accessible from bloblang, too. I did not investigate too much before adding bloblang method. Your suggestion sounds good, but I first need to check it.

sananguliyev: Not exposing this as a bloblang method is also an option, and I can remove this method implementation.

gregfurman: I'd prefer the context based approach. Else, the more we can rely on open telemetry tracing context the better!

gregfurman: Hey @sananguliyev 👋 Do you keen have capacity to finish up this PR? I'm happy to take it over otherwise 😄 Lmk!

internal/bundle/tracing/events.go
sananguliyev: @gregfurman I did not get whether you agree to keep this like this.

internal/bundle/tracing/events.go
sananguliyev: Hey @gregfurman, I might have some time next week. I will try to apply what you suggested until next weekend. If you want to release in the next a couple of days then feel free to take it over :)

gregfurman: No pressure 🙂 Let me move it back into draft then

sananguliyev: I found some time today, and fixed the issue you pointed here. You're absolutely right about metadata being mutable and users can mess with that, which should not be the case since flow ID is immutable by nature. So please take another look. And feel free to make changes here in case you see something small, in order to move faster.

sananguliyev: @gregfurman do you have capacity to review this? I am planning to make another contribution to enable CDC source, but first want to finish this and then contribute the component.

gregfurman: Hey! Sorry I must've missed the notification since I placed the PR into draft 🤦 

I'll give the PR another look this evening 🙂

internal/bloblang/query/functions.go
```diff
@@ -263,6 +264,41 @@ var _ = registerSimpleFunction(
...
+			return flowID, nil
+		}
+
+		// Try to use OpenTelemetry trace ID if available
+		if traceID := tracing.GetTraceID(part); traceID != "" && traceID != "00000000000000000000000000000000" {
+			// Store it in context for future use
+			ctx = tracing.WithFlowID(ctx, traceID)
+			*part = *part.WithContext(ctx)
+			return traceID, nil
+		}
```
gregfurman: Unsure this is desirable 🤔

internal/bundle/tracing/processor.go
```diff
@@ -52,13 +59,25 @@ func (t *tracedProcessor) ProcessBatch(ctx context.Context, m message.Batch) ([]
...
+			for i := 0; i < len(hasOutput) && i < outMsg.Len(); i++ {
+				hasOutput[i] = true
+			}
+		}
 	}
+
+	// If no output messages, record delete events for each input part
 	if len(outMsgs) == 0 {
-		// TODO: Find a better way of locating deletes (using batch index tracking).
-		t.e.Add(EventDeleteOf())
```
gregfurman, line 61: Nice! Should we deprecate `EventDeleteOf()` in favour of this `EventDeleteOfPart` function?

internal/impl/pure/processor_workflow_test.go
```diff
@@ -1038,6 +1038,38 @@ workflow:
 	require.NoError(t, strm.Stop(tCtx))
 
 	assert.Equal(t, `{"content":"waddup","id":"HELLO WORLD","meta":{"workflow":{"succeeded":["fooproc"]}}}`, outValue)
+
+	// Normalize events for testing by removing FlowID, Timestamp, and _bento_flow_id metadata
```
gregfurman: Are we still using this `_bento_flow_id` approach?

internal/bloblang/query/functions.go
```diff
@@ -263,6 +264,41 @@ var _ = registerSimpleFunction(
...
 
+var _ = registerSimpleFunction(
+	NewFunctionSpec(
+		FunctionCategoryMessage, "flow_id",
+		"Provides the message flow id used for tracing the journey of a message through the pipeline. If no flow ID exists, one will be created and stored in the message context.",
+		NewExampleSpec("",
+			`meta flow_id = flow_id()`,
+		),
+	).Experimental(),
+	func(fCtx FunctionContext) (any, error) {
```
gregfurman, line 274: Does it make sense to be setting the `flowID` from within this bloblang function?

Should it not rather just retrieve the `flowID`/`traceID`

internal/bloblang/query/functions.go
```diff
@@ -263,6 +264,41 @@ var _ = registerSimpleFunction(
...
+		part := fCtx.MsgBatch.Get(fCtx.Index)
+		ctx := part.GetContext()
+
+		// Check if we already have a flow ID in context
+		if flowID := tracing.GetFlowID(ctx); flowID != "" {
+			return flowID, nil
+		}
+
+		// Try to use OpenTelemetry trace ID if available
```
gregfurman: nit: Can we remove some of these comments? Also this looks largely identical to the `getOrCreateFlowID`

internal/bundle/tracing/events.go
gregfurman: Yeah I suppose this makes sense. Do you imagine a situation where the `flowID` and `traceID` should ever differ? 

Also, perhaps we set `00000000000000000000000000000000` to a const so it's not a magic string?

public/service/tracing_test.go
```diff
@@ -104,6 +104,37 @@ logger:
...
+				normalizedEvs[i] = service.TracingEvent{
+					Type:    ev.Type,
+					Content: ev.Content,
+					Meta: func() map[string]any {
+						if ev.Meta == nil {
+							return nil
+						}
+						meta := make(map[string]any)
+						for mk, mv := range ev.Meta {
+							if mk != "_bento_flow_id" {
```
gregfurman: Already mentioned. Do we still want this metadata?

gregfurman (commented):
Thanks for the follow ups on this!

I'm thinking whether we want to ever set the `flowID` values outside of the `traceInput` (i.e when a message part is first created), although I suppose there's the case where we want to create child-spans. 

https://github.com/warpstreamlabs/bento/blob/852a365ab34059d738632702645a68f792ddc489/internal/bundle/tracing/input.go#L52-L57

Let me know!

internal/impl/pure/processor_workflow_test.go
sananguliyev: Yes, we need some cleanup.

internal/bloblang/query/functions.go
sananguliyev: Since we have this, I thought it makes sense to have access to this value in case some users want to add it to logs or somewhere else for logging or even some other purposes, because it's same as trace ID which we can currently access.

internal/bloblang/query/functions.go
sananguliyev: We definitely need some general cleanup.

internal/bundle/tracing/events.go
sananguliyev: I do not think they should ever differ if tracing enabled. The whole point here is to have some ID, but not optional like trace ID. And also if there is trace ID I do not see a problem with having different IDs.

internal/impl/pure/processor_workflow_test.go
gregfurman: I thought we were using the struct{} approach instead of the string for specifying the key?

internal/bloblang/query/functions.go
gregfurman: I think this function should change to only receive the flowID, and not set it if not found. Wdyt?

sananguliyev: The main idea here is it should always be available. Regardless where it's first called it should either return existing and create new and set to ctx. E.g. we have steps a and b. it's always created at a, and when b reads it it get the already created one. If there is any case and for whatever reason we add one more step which runs before the a then it creates flow id and a just get what is there instead of creating.

gregfurman: This should always be created anyway if tracing is enabled since the `input` component is wrapped in a `tracedInput` component.

Imagine this case:

```yaml
input:
   resource: A

pipeline:
   processors:
     - resource: B
     - mapping: "meta flow_id = flow_id()" # <-- set and get here

output:
  resource: C
```

If not set at the input layer, setting a `flow_id` at the `mapping` step does not help us at all with knowing the flow of a message through a pipeline -- only that it was sent on to a resource `C`

Since all messages (in a normal pipeline) are initially created at the input layer, `EventProduce` (which is at the input layer) means there should be an attached flowID -- making the create call redundant.

https://github.com/warpstreamlabs/bento/blob/b90d55af9e2cc0efe50ad0feabe9761bc1eba530/internal/bundle/tracing/events.go#L42-L48

Does this make sense? Or perhaps I conceptually not understanding the inclusion of `flowID` and when we expect it to be called.

internal/impl/pure/processor_workflow_test.go
sananguliyev: ✅

internal/bundle/tracing/events.go
sananguliyev: Since we have other comments and we changed the way of storing flow_id, I resolve this comment.

internal/bloblang/query/functions.go
sananguliyev: Your concern is valid. I apply this logic:

Regular Build (without tracing):**
```
Input → flowIDInput (ensures flow ID) → Processor → Output
```

**Traced Build:**
```
Input → flowIDInput (ensures flow ID) → tracedInput (records events) → Processor → Output

**Comment by sananguliyev** [comment]
> Thanks for the follow ups on this!
> 
> I'm thinking whether we want to ever set the `flowID` values outside of the `traceInput` (i.e when a message part is first created), although I suppose there's the case where we want to create child-spans.
> 
> https://github.com/warpstreamlabs/bento/blob/852a365ab34059d738632702645a68f792ddc489/internal/bundle/tracing/input.go#L52-L57
> 
> Let me know!

Why We Need Both

1. **`flowIDInput`**: Ensures flow IDs are **always available** (even without tracing)
2. **`tracedInput`**: Records **what happens** with those flow IDs (only when user wants tracing)

______
- **flowIDInput** = "Give every message a passport (flow ID)"
- **tracedInput** = "Write down in a log book every time a message passes through"


And this also helps to group events not only by section but also by flow.

Please, take another look. Thanks in advance. @gregfurman

File: internal/bundle/tracing/processor.go (hunk 4 above)
Comment (sananguliyev) [reply internal/bundle/tracing/processor.go:61]: Deprecated, and also added comment where it's used to mention that it's deprecated.

File: internal/bundle/tracing/flow_wrapper.go [hunk 9]
```diff
@@ -0,0 +1,85 @@
...
+
+import (
+	"context"
+
+	"github.com/Jeffail/shutdown"
+
+	"github.com/warpstreamlabs/bento/internal/component"
+	"github.com/warpstreamlabs/bento/internal/component/input"
+	"github.com/warpstreamlabs/bento/internal/message"
+	internaltracing "github.com/warpstreamlabs/bento/internal/tracing"
```
gregfurman, line 11: nit: why `internaltracing` alias here?

public/service/stream_builder.go
```diff
@@ -841,7 +842,8 @@ func (s *StreamBuilder) Build() (*Stream, error) {
 // Experimental: The behaviour of this method could change outside of major
 // version releases.
 func (s *StreamBuilder) BuildTraced() (*Stream, *TracingSummary, error) {
-	tenv, summary := tracing.TracedBundle(s.env.internal)
+	flowEnv := tracing.FlowIDBundle(s.env.internal)
+	tenv, summary := tracing.TracedBundle(flowEnv)
```
gregfurman, line 856: Can we rather use a `BuildTracedV2` with this approach? Then we leave `BuildTraced` as-is. In the next few releases we can do a switch-over once we're confident the new tracing is stable. 

Wdyt?

public/service/stream_builder.go
```diff
@@ -828,7 +828,8 @@ func (s *StreamBuilder) runConsumerFunc(mgr *manager.Type) error {
 // Build a Bento stream pipeline according to the components specified by this
 // stream builder.
 func (s *StreamBuilder) Build() (*Stream, error) {
-	return s.buildWithEnv(s.env.internal, false)
+	flowEnv := tracing.FlowIDBundle(s.env.internal)
+	return s.buildWithEnv(flowEnv, false)
```
gregfurman: Perhaps we shouldn't add this by default 👀

internal/bundle/tracing/flow_wrapper.go
```diff
@@ -0,0 +1,85 @@
...
+
+// flowIDInput wraps an input to ensure all messages have flow IDs initialized.
+// This is independent of tracing and ensures flow IDs are always available.
+type flowIDInput struct {
+	wrapped input.Streamed
+	tChan   chan message.Transaction
+	shutSig *shutdown.Signaller
+}
+
+func wrapWithFlowID(i input.Streamed) input.Streamed {
```
gregfurman, line 22: Once this FlowID approach is GA, perhaps we unify this `wrapWithFlowID` approach with the `traceInput` approach. Perhaps we add a TODO for that. Thoughts?

sananguliyev: Yes, sure. Is writing code comments enough, or do you want to somehow document it somewhere?

internal/bundle/tracing/flow_wrapper.go
sananguliyev: Without the alias, there would be a naming conflict. The current package is already `tracing`, so importing another package with the same name would be ambiguous. The alias `internaltracing` makes it clear which `tracing` package is being referenced when used in the code.

public/service/stream_builder.go
sananguliyev: TBH, I did not think about it from the beginning. Thanks for the suggestion. It's better to have v2, and then we can make it default once we're sure that it's stable.

public/service/stream_builder.go
sananguliyev: I have rolled back to not make it default since we introduced new `BuildTracedV2` ✅

public/service/stream_builder.go
sananguliyev: ✅  Introduced new `BuildTracedV2` which is coming with flow id

sananguliyev:
@gregfurman I have added new `BuildTracedV2` as you suggested, which was good idea 👍 

Please, take another look to finalize this implementation.

internal/bloblang/query/functions.go
```diff
@@ -263,6 +263,21 @@ var _ = registerSimpleFunction(
 	},
 )
 
+var _ = registerSimpleFunction(
+	NewFunctionSpec(
+		FunctionCategoryMessage, "flow_id",
+		"Returns the message flow ID used for tracing the journey of a message through the pipeline. Flow IDs are automatically assigned at the input layer.",
```
gregfurman:
```suggestion
		"Returns the message flow ID used for tracing the journey of a message through the pipeline. Flow IDs are automatically assigned at the input layer. Note, that this is only available for those streams built with `BuildTracedV2`.",
```

PR #345, opened 2025-05-27
create elasticsearch_v2 output

jem-davies:
Creates new `elasticsearch_v2` output component using official elastic module: /elastic/go-elasticsearch/v9

Comparison with `elasticsearch` output: 

- `sniff` -> replaced with `discover_nodes_on_start` & `discover_nodes_interval` 
- `healthcheck` -> removed
- `pipeline` -> no longer supports bloblang interpolation
- `type` -> removed - `type` deprecated from elasticsearch itself since version v6.0.0
- `aws` -> removed - recommend to use the `opensearch` components 

TODO: 

- [x] refactor tests for DRY
- [x] error handling
- [x] add CommonRetryBackoffFields 
- [x] write the documentation 
- [x] mention duplicate updates to same document in 1 batch will cause a failure
- [x] fix update action 
- [x] more tests
- [x] write a test for each action 
- [x] ~~update `elasticsearch` to be a deprecated component ?~~ Needs to be done on a follow up PR after 1.9.0 is done otherwise the docs in the meantime will be wrong


internal/impl/elasticsearch/output_v2.go
```diff
@@ -0,0 +1,436 @@
...
+			out, err = EsoOutputConstructor(conf, mgr)
+			return
+		})
+	if err != nil {
+		panic(err)
+	}
+}
+
+func OutputSpecV2() *service.ConfigSpec {
+	return service.NewConfigSpec().
```
jem-davies, line 59: missing Version func

internal/impl/elasticsearch/output_v2.go
```diff
@@ -0,0 +1,436 @@
...
+		Timeout: eso.conf.timeout,
+	})
+	if err != nil {
+		return err
+	}
+
+	for i, msg := range batch {
+		index, action, id, routing, err := batchInterpolator.exec(i)
+		if err != nil {
+			eso.log.Errorf("Failed to execute bloblang interpolation: %v\n", err)
```
gregfurman: question: do we have any error logging when outputs fail with a `BatchError`? Am wondering if all of these log statements are necessary.

internal/impl/elasticsearch/output_v2.go
```diff
@@ -0,0 +1,436 @@
...
+			if err != nil {
+				eso.log.Errorf("Failed to add message to elasticsearch indexer: %v\n", err)
+				batchErrFailed(i, err)
+			}
+		case "index", "create":
+			err = indexer.Add(ctx, esutil.BulkIndexerItem{
+				Index:      index,
+				Action:     action,
+				DocumentID: id,
+				Routing:    routing,
```
gregfurman: We could make this a lot more concise i.e

```golang
		var reader io.ReadSeeker
		switch action {
		case "update", "upsert":
			updateBody := map[string]any{
				"doc":           json.RawMessage(msgBytes),
				"doc_as_upsert": action == "upsert",
			}
			bodyBytes, err := json.Marshal(updateBody)
			if err != nil {
				eso.log.Errorf("Failed to marshal update/upsert doc: %v\n", err)
				return err
			}

			reader = bytes.NewReader(bodyBytes)
		case "index", "create":
			reader = bytes.NewReader(msgBytes)
		case "delete":
		default:
			batchErrFailed(i, fmt.Errorf("elasticsearch action '%s' is not allowed", action))
		}

```

internal/impl/elasticsearch/output_v2.go
```diff
@@ -0,0 +1,436 @@
...
+	return nil
+}
+
+func (eso *EsOutput) Close(context.Context) error {
+	return nil
+}
+
+//------------------------------------------------------------------------------
+
+func (eso *EsOutput) newInterpolationExecutor(batch *service.MessageBatch) (ie interpolationExecutor) {
```
gregfurman: Does this need to be a pointer?
```suggestion
func (eso *EsOutput) newInterpolationExecutor(batch service.MessageBatch) (ie interpolationExecutor) {
```

internal/impl/elasticsearch/integration_v2_test.go
```diff
@@ -0,0 +1,269 @@
...
+		pollURL := tc.pollURL(port)
+
+		var client *http.Client
+		var configArgs []string
+		if tc.name == "TLS" {
+			// give elasticsearch a moment to create a tls cert
+			time.Sleep(time.Second * 20)
+
+			containerName := resource.Container.Name[1:]
+			cmd := exec.Command("docker", "exec", containerName, "cat", "config/certs/http_ca.crt")
```
gregfurman, line 203: Could be replaced with:

```suggestion
			resource.Exec([]string{"cat", "config/certs/http_ca.crt"}, dockertest.ExecOptions{})
```

Just give a look at whether `ExecOptions` needs to be configured for stdout.

internal/impl/elasticsearch/integration_v2_test.go
```diff
@@ -0,0 +1,269 @@
...
+}
+
+//------------------------------------------------------------------------------
+
+func startElasticsearch(t *testing.T, env []string) (pool *dockertest.Pool, resource *dockertest.Resource, port string) {
+	pool, err := dockertest.NewPool("")
+	require.NoError(t, err)
+
+	pool.MaxWait = time.Minute * 3
+	resource, err = pool.Run("elasticsearch", "8.16.5", env)
```
gregfurman, line 234: suggestion: Should we be mounting that `cert` directory to a local tmpdir? That way we may not need the `exec` via the docker SDK

internal/impl/elasticsearch/output_v2.go
```diff
@@ -0,0 +1,436 @@
...
+	esoV2FieldAuthEnabled           = "enabled"
+	esoV2FieldAuthUsername          = "username"
+	esoV2FieldAuthPassword          = "password"
+	esoV2FieldCompressRequestBody   = "compress_request_body"
+	esoV2FieldBatching              = "batching"
+	esoV2FieldRetryOnStatus         = "retry_on_status"
+)
+
+func init() {
+	err := service.RegisterBatchOutput("elasticsearch_v2", OutputSpecV2(),
```
gregfurman, line 42: Does this have AWS support? If so, can we add a test against LocalStack in integration tests?

internal/impl/elasticsearch/integration_v2_test.go
jem-davies: This would be a simpler way yeah - but I think this a test, and it does the job right?

internal/impl/elasticsearch/output_v2.go
jem-davies: yeah good spot thanks - seems if you trigger a batch error you will get double logs - so have removed them before we call `batchErrFailed()`

internal/impl/elasticsearch/output_v2.go
jem-davies: I did write it - then realised that it could be reduced in terms of number of lines... but I think more readable as-is, just calling indexer.Add in each case, with the relevant options for each action.

internal/impl/elasticsearch/output_v2.go
jem-davies: I did do this purposefully so that the potentially large service.MessageBatch isn't copied as per Go's default when passed into a function.

internal/impl/elasticsearch/integration_v2_test.go
jem-davies: Hmm that resource.Exit will only return an exitCode int, err error - but we need the cmd.Output to later write to a file?

internal/impl/elasticsearch/output_v2.go
jem-davies: No the advice would be to use the `opensearch` components instead of this.

internal/impl/elasticsearch/output_v2.go
jem-davies: Added with Version("1.11.0") 🤞

internal/impl/elasticsearch/output_v2.go
gregfurman: @jem-davies A slice already contains a pointer to the underlying array -- so am unsure what the purpose of using a reference to a struct that already contains a reference is?

gregfurman: Basically a slice has 3 fields: a pointer to the underlying array, a length field, and a capacity field. We wont be passing the entire array each time, just the 3 values mentioned before. 

I think that any performance benefit we get from passing a pointer to a slice is offset by needing to extract the pointer in the first place.

In anycase, it's not _really_ idiomatic and there are so few fields I don't think performance needs to be a consideration here 🤷

jem-davies: Sorry yeah you are correct

jem-davies: 🤷

jem-davies:
@gregfurman  thanks for the feedback - I have stopped making any further changes ...

internal/impl/elasticsearch/output_v2.go
```diff
@@ -0,0 +1,434 @@
...
+		}
+		batchErr.Failed(i, err)
+	}
+
+	batchInterpolator := eso.newInterpolationExecutor(batch)
+
+	indexer, err := esutil.NewBulkIndexer(esutil.BulkIndexerConfig{
+		Client: eso.client,
+		OnError: func(ctx context.Context, err error) {
+			eso.log.Errorf("elasticsearch_v2: Bulk Indexer Error: %s\n", err)
```
gregfurman: Wont the component name already be present in the logger?

```suggestion
			eso.log.Errorf("Bulk Indexer Error: %w", err)
```

internal/impl/elasticsearch/output_v2.go
```diff
@@ -0,0 +1,434 @@
...
+		return
+	}
+	return
+}
+
+//------------------------------------------------------------------------------
+
+func onSuccessHandler(eso *EsOutput) func(ctx context.Context, item esutil.BulkIndexerItem, res esutil.BulkIndexerResponseItem) {
+	return func(ctx context.Context, item esutil.BulkIndexerItem, res esutil.BulkIndexerResponseItem) {
+		eso.log.Trace(fmt.Sprintf("Indexed %s/%s", res.Index, res.DocumentID))
```
gregfurman:
```suggestion
		eso.log.Tracef("Indexed %s/%s", res.Index, res.DocumentID)
```

nit: Does this happen per batch or per item? Am _slightly_ concerned we could spam the log output but I suppose this is at the trace level 👀

internal/impl/elasticsearch/integration_v2_test.go
gregfurman: We can extract the output like this:

```suggestion
var stdout bytes.Buffer

errCode, err := resource.Exec(
    []string{"cat", "config/certs/http_ca.crt"}, 
    dockertest.ExecOptions{
        StdOut: &stdout,
    },
)

// handle errCode & err

```

FYI this isn't blocking! I'm just traumatised because of how some of our tests look haha. Feel free to resolve if you want

internal/impl/elasticsearch/output_v2.go
jem-davies: Per item yeah - applied the suggestion and changed the word "Indexed" for "Processed" because it could be any of the actions.

PR #346, opened 2025-05-27
fix tls connection with elasticsearch output when not using aws

jem-davies:
- fixes connectivity with elasticsearch using tls
- adds an integration test to check elasticsearch tls _without aws_

### DIFF
--- a/internal/impl/elasticsearch/integration_test.go
+++ b/internal/impl/elasticsearch/integration_test.go
@@ -241,11 +241,11 @@ func TestIntegrationConnectTLS(t *testing.T) {
 	require.NoError(t, err)
 
 	client := createHTTPClientWithCA(t, output)
-	polllUrl := fmt.Sprintf("https://elastic:password@localhost:%s", resource.GetPort("9200/tcp"))
-	configUrl := fmt.Sprintf("https://localhost:%s", resource.GetPort("9200/tcp"))
+	polllURL := fmt.Sprintf("https://elastic:password@localhost:%s", resource.GetPort("9200/tcp"))
+	configURL := fmt.Sprintf("https://localhost:%s", resource.GetPort("9200/tcp"))
 
 	err = pool.Retry(func() error {
-		resp, err := client.Get(fmt.Sprintf("%s/_cluster/health", polllUrl))
+		resp, err := client.Get(fmt.Sprintf("%s/_cluster/health", polllURL))
 		if err != nil {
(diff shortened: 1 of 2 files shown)


internal/impl/elasticsearch/output.go
```diff
@@ -98,20 +98,23 @@ func esoConfigFromParsed(pConf *service.ParsedConfig) (conf esoConfig, err error
...
-		}
+		transport.TLSClientConfig = tlsConf
 	}
 
+	httpClient := &http.Client{
+		Transport: transport,
+		Timeout:   timeout,
+	}
+
+	conf.clientOpts = append(conf.clientOpts, elastic.SetHttpClient(httpClient))
```
gregfurman, line 116: Is setting the HTTP client not the only change that's required here?

internal/impl/elasticsearch/integration_test.go
```diff
@@ -262,3 +342,15 @@ output:
 		integration.StreamTestOptPort(resource.GetPort("9200/tcp")),
 	)
 }
+
+func createHTTPClientWithCA(t *testing.T, caBytes []byte) *http.Client {
```
gregfurman, line 346: question: Are these changes only required for improved testing?

gregfurman (approved):
LGTM. Nothing blocking 👍

internal/impl/elasticsearch/integration_test.go
jem-davies: Yeah, just a change that tests the default connection setup with Elasticsearch — TLS using a self-signed CA.

jem-davies: where we poll a ready check  before executing the test - we need that to work with tls

internal/impl/elasticsearch/output.go
jem-davies:
```suggestion
	if !pConf.Contains(esoFieldAWS) {
		conf.clientOpts = append(conf.clientOpts, elastic.SetHttpClient(httpClient))
	}
```

should we do this instead? @gregfurman

jem-davies: that actually doesn't seem to work for some reason...

jem-davies: pretty sure there is an issue with the .Contains method - returning true even if you don't set an AWS field 😥

PR #348, opened 2025-05-29
[NOT READY FOR REVIEW] Recreate kafka client on unknown topic

brianshih1:


internal/impl/kafka/input_kafka_franz.go
```diff
@@ -757,6 +768,20 @@ func (c *checkpointTracker) removeTopicPartitions(ctx context.Context, m map[str
...
+	if errors.Is(err, context.DeadlineExceeded) ||
+		errors.Is(err, context.Canceled) {
+		return true
+	}
+
+	isUnknownTopicErr := err == kerr.UnknownTopicOrPartition || err == kerr.UnknownTopicID
+	if isUnknownTopicErr && f.reconnectOnUnknownTopic {
+		return false
+	}
+	return kerr.IsRetriable(err)
```
gregfurman, line 789: The addition of this `IsRetriable` means we're going to be recreating the client on more than just `kerr.UnknownTopicOrPartition` or `kerr.UnknownTopicID` errors (when `reconnectOnUnknownTopic` is enabled).

Can we update the documentation to reflect this?

internal/impl/kafka/input_kafka_franz.go
```diff
@@ -95,6 +96,10 @@ Finally, it's also possible to specify an explicit offset to consume from by add
 			Description("Determines whether to consume from the oldest available offset, otherwise messages are consumed from the latest offset. The setting is applied when creating a new consumer group or the saved offset no longer exists.").
 			Default(true).
 			Advanced()).
+		Field(service.NewBoolField("reconnect_on_unknown_topic").
+			Description("Determines whether to close the client and force a reconnect after seeing an UNKNOWN_TOPIC_OR_PARTITION or UNKNOWN_TOPIC_ID error.").
```
gregfurman, line 100: nit: I wonder if we should rather name this `reconnect_on_unretriable_error` (or some equivalent) and mention all errors considered unretriable will trigger a re-connect, including `UNKNOWN_TOPIC_OR_PARTITION` or `UNKNOWN_TOPIC_ID`.

internal/impl/kafka/input_kafka_franz.go
```diff
@@ -95,6 +96,10 @@ Finally, it's also possible to specify an explicit offset to consume from by add
...
+			Default(false).
```
gregfurman, line 101: Let's add a version to this field:

```suggestion
			Version("1.8.0").
			Default(false).
```

gregfurman (approved):
LGTM! I ran the integration tests manually and all looks good.

Just a note on the new field, will it not introduce new behaviour and cause more than just the `UNKNOWN_TOPIC` errors to recreate the client? If so, perhaps we should add some docs for this. Else, maybe we should even just enable this behaviour by default 🤔

internal/impl/kafka/input_kafka_franz.go
brianshih1: we won't be recreating the client for retriable errors outside of `kerr.UnknownTopicOrPartition` or `kerr.UnknownTopicID`.

```
if f.isRetriableError(err.Err) {
          continue
}
nonTemporalErr = true
```
Retriable errors will not mark the error as a `nonTemporalErr` which will not trigger a reconnect.

internal/impl/kafka/input_kafka_franz.go
brianshih1: I think for now I'd prefer a more granular flag just for unknown topics/partitions. 

Currently `input_kafka_franz` already reconnects on "unretriable" franzgo errors. The `UNKNOWN_TOPIC_OR_PARTITION` and `UNKNOWN_TOPIC_ID` errors are a bit unique in that they are classified as retriable by franzgo but consumers would actually be stuck. This is a known issue: https://github.com/twmb/franz-go/issues/676 and there is an issue opened for it: https://github.com/twmb/franz-go/issues/908.

If they fix that issue we can remove this flag as we won't need to force reconnect on that error.

internal/impl/kafka/input_kafka_franz.go
gregfurman: Maybe I'm misunderstanding but it looks like there is a long list of error types that `IsRetriable` would determine to be unretriable returns (from [pkg/kerr/kerr.go](https://github.com/twmb/franz-go/blob/master/pkg/kerr/kerr.go#L75-L206)).

Are the only errors that can be returned on `fetch` that are retriable `UnknownTopicOrPartition` and `UnknownTopicOrPartition`?

Feel free to tell me I'm overthinking this but since my kafka knowledge is limited I want to be sure I fully understand the changes we're introducing here (especially since this component is hella important).

brianshih1: Ahh yea the logic is a bit weird, let me know if my thinking is wrong:
cc: @aratz-lasa 

So by default, franzgo strips retryable errors from fetch responses (as stated [here](https://github.com/twmb/franz-go/blob/master/pkg/kgo/config.go#L1598)), unless `kgo.KeepRetryableFetchErrors()` is added to the client opts.

Before this PR, `kgo.KeepRetryableFetchErrors()` is not provided, so none of the errors [here](https://github.com/warpstreamlabs/bento/blob/main/internal/impl/kafka/input_kafka_franz.go#L875) are retryable Franzgo errors.

`UnknownTopicOrPartition` is a retryable error according to franzgo, as defined [here](https://github.com/twmb/franz-go/blob/master/pkg/kerr/kerr.go#L79).

This means that before this PR, if the client gets an unknown topic error, it will just keep failing instead of forcing a reconnect as it will never set nonTemporalErr to true [here](https://github.com/warpstreamlabs/bento/blob/main/internal/impl/kafka/input_kafka_franz.go#L884).

What I want to do in this PR is:
1. set the `kgo.KeepRetryableFetchErrors()` error so that franzgo doesn't strip away retriable errors
```
if f.reconnectOnUnknownTopic {
        clientOpts = append(clientOpts, kgo.KeepRetryableFetchErrors())
}
```
2. make sure that if the error is `UnknownTopicOrPartition` or  `UnknownTopicID`, we force a reconnect by setting `nonTemporalErr` to true. But for all other retriable errors, don't force a reconnect.
- since the logic got a bit complex, I moved the logic into a `isRetriableError` method. If it returns true, then the client won't reconnect. If it returns false, it will be reconnected. Right now, the only errors where we force a reconnect are:
       - `UnknownTopicOrPartition` or `UnknownTopicID`
       - errors that are unretryable according to franzgo

This means that apart from `UnknownTopicOrPartition` and `UnknownTopicID`, all other errors will behave as it did before.

brianshih1: I also just added some comments inside `isRetriableError` that explains why we need special casing for those 2, lmk if it doesn't make sense!

internal/impl/kafka/input_kafka_franz.go
```diff
@@ -872,19 +908,15 @@ func (f *franzKafkaReader) Connect(ctx context.Context) error {
 				// forcing a reconnect.
 				nonTemporalErr := false
 
-				for _, kerr := range errs {
-					// TODO: The documentation from franz-go is top-tier, it
-					// should be straight forward to expand this to include more
-					// errors that are safe to disregard.
-					if errors.Is(kerr.Err, context.DeadlineExceeded) ||
-						errors.Is(kerr.Err, context.Canceled) {
+				for _, err := range errs {
```
aratz-lasa, line 911: why not keep it as `kerr`?

brianshih1: name collision with the `kerr` package

internal/impl/kafka/input_kafka_franz.go
gregfurman: > This means that apart from UnknownTopicOrPartition and UnknownTopicID, all other errors will behave as it did before.

Ahh I see. I was thinking that this change would cause more errors types (both retryable and non-retryable) to propogate as responses (because I'm clearly a bit slow). 

Thanks for the explanation 😄

gregfurman:
@brianshih1 Neat changes! Can we change the title to include the component name (since this is only for `kafka_franz` not the sarama Kafka input) i.e 

> Recreate kafka_franz input client on unknown topic

When that's done, I'm happy to merge 👍

PR #352, opened 2025-06-01
fix(govet): Refactor test helpers to fix non-constant format string error

gregfurman:
## Motivation

In preperation for updating to Go 1.24, format print statements now require non-constant format strings to be explicitly handled -- otherwise govet causes the build to fail (see [docs](https://tip.golang.org/doc/go1.24#vet)).

## Changes

- Reworked helper functions to appease the the go vet checker.


jem-davies (approved):
Looks good!

PR #365, opened 2025-06-10
feature: add duckdb support

James-Gilbert-:
### Changes
- Adds basic support for DuckDB in file based mode using [go-duckdb](https://github.com/marcboeker/go-duckdb/) with the `x_bento_extra` tag
- Upgrades to Go 1.24

### Testing
- I verified some basic functionality with reading/exporting to parquet by mounting a writeable volume. According to the `go-duckdb` docs, JSON + Parquet support should be included in the binaries.
- Purely in-memory behavior wasn't tested. As far as I can tell, there is some remaining effort required for safe concurrent access via the go bindings. 

### Notes

Due to some ABI compatibility issues with the pre-built static binaries, a newer base image and changes to the Dockerfile.cgo were needed. It is possible to build each of the dependencies from source (either dynamically or statically).

From experiments, it's possible to get a statically linked binary down to ~200MB building from source, with build time and licensing requirements being the primary trade-off.


internal/impl/sql/conn_fields.go
```diff
@@ -286,6 +290,13 @@ func reworkDSN(driver, dsn string) (string, error) {
 }
 
 func sqlOpenWithReworks(ctx context.Context, logger *service.Logger, driver, dsn string, connSettings *connSettings) (*sql.DB, error) {
+	if !slices.Contains(sql.Drivers(), driver) {
```
gregfurman: I really like this approach! I think, however, that we should leverage build tags to dynamically set how we should be validating this. For example, see the `driverField` variable above at https://github.com/warpstreamlabs/bento/blob/9fc0fcddde6d18175da9013f48b8618580b1c091/internal/impl/sql/conn_fields.go#L17-L18

What if instead we added a `LintRule` that (by default) returned this error message you've added below except, we only set this when the `x_bento_extra` tag has NOT been used. 

So we add a `conn_fields_pure.go` where we overwrite this `driverField` to raise a lint error when not in CGO/bento_extra:

```golang
//go:build !x_bento_extra

package sql

func init() {
   driverField = service.NewStringEnumField("driver", "mysql", "postgres", "clickhouse", "mssql", "sqlite", "oracle", "snowflake", "trino", "gocosmos", "spanner").
	Description("A database [driver](#drivers) to use.").LintRule(`root = if this == "duckdb" { [ "Cannot use DuckDB driver outside of CGO build. See ..." ] }`)
}
```

internal/impl/sql/integration_extras_test.go
```diff
@@ -0,0 +1,41 @@
...
+		_, err := db.Exec(fmt.Sprintf(`create table %s (
+			"foo" varchar(50) not null,
+			"bar" integer not null,
+			"baz" varchar(50) not null,
+			primary key ("foo")
+		)`, name))
+		return name, err
+	}
+
+	require.NoError(t, db.Ping())
```
gregfurman: Could this `db.Ping()` be flaky if the DB is still being created? Perhaps add a `require.Eventually` that can check whether this is true multiple times before erroring out.

gregfurman (commented):
Wow thanks for the PR! Have left some comments.

I was giving the original thread and linked issues a read. It appears that duckdb v1.3.1 will be fixing this. See https://github.com/marcboeker/go-duckdb/issues/454#issuecomment-2948878397

Should we perhaps just wait until this has been fixed and see if that'll solve the CGO dependency? Would be great to have this in the pure Go bento build 😄

internal/impl/sql/integration_extras_test.go
James-Gilbert-: Added a `require.Eventually` before table creation 👍

James-Gilbert-:
> Wow thanks for the PR! Have left some comments.
> 
> I was giving the original thread and linked issues a read. It appears that duckdb v1.3.1 will be fixing this. See [marcboeker/go-duckdb#454 (comment)](https://github.com/marcboeker/go-duckdb/issues/454#issuecomment-2948878397)
> 
> Should we perhaps just wait until this has been fixed and see if that'll solve the CGO dependency? Would be great to have this in the pure Go bento build 😄

Due to C++ deps, CGO builds are needed for duckdb. But there are a few strategies to reduce overall image and/or binary size.

gregfurman:
Can you run a `make docs` and commit the updated`.md` files?

internal/impl/sql/conn_fields_init.go
```diff
@@ -0,0 +1,10 @@
+//go:build !x_bento_extra
+
+package sql
+
+import "github.com/warpstreamlabs/bento/public/service"
+
+func init() {
+	driverField = service.NewStringEnumField("driver", "mysql", "postgres", "clickhouse", "mssql", "sqlite", "oracle", "snowflake", "trino", "gocosmos", "spanner").
+		Description("A database [driver](#drivers) to use.").LintRule(`root = if this == "duckdb" { [ "Cannot use DuckDB driver outside of CGO build. Use a binary built with tag x_bento_extra, or the CGO Docker image" ] }`)
+}
```
gregfurman, line 9: Tested this locally and all looks good!

resources/docker/Dockerfile.cgo
```diff
@@ -1,36 +1,72 @@
-FROM golang:1.23 AS build
+FROM ubuntu:24.04 AS build
```
gregfurman, line 1: Can we not stay on an official `golang` image when building?

resources/docker/Dockerfile.cgo
```diff
@@ -1,36 +1,72 @@
...
+COPY go.mod go.sum ./
+RUN go mod download
+
+# Copy the full source and build
+COPY . ./
 RUN make TAGS=x_bento_extra
 
 # Pack
-FROM debian:latest
+FROM ubuntu:24.04 AS final
```
gregfurman, line 41: Similar question as above: why the move over to `ubuntu`?

resources/docker/Dockerfile.cgo
```diff
@@ -1,36 +1,72 @@
...
+ENV PATH="/usr/local/go/bin:${PATH}"
 
+WORKDIR /go/src/github.com/warpstreamlabs/bento/
+
+# Separate go.mod copy for cache
+COPY go.mod go.sum ./
+RUN go mod download
+
+# Copy the full source and build
+COPY . ./
```
gregfurman, line 37: Wont this also copy the entire `website/`, `.git/`, `resources/` etc.?

resources/docker/Dockerfile.cgo
```diff
@@ -1,36 +1,72 @@
...
 
 LABEL maintainer="WarpStream Labs <security@warpstreamlabs.com>"
 LABEL org.opencontainers.image.source="https://github.com/warpstreamlabs/bento"
 
-WORKDIR /root/
+WORKDIR /app
+
+# Install minimal runtime dependencies
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    libzmq5 ca-certificates \
```
gregfurman, line 50: Why install `ca-certificates` both in `builder` and again in `final`?

resources/docker/Dockerfile.cgo
James-Gilbert-: Yeah, has the same behavior as before but uses the relative path.

resources/docker/Dockerfile.cgo
James-Gilbert-: Needed a newer C++ runtime environment for the duckdb dependencies since it's not a fully static build

resources/docker/Dockerfile.cgo
James-Gilbert-: This was a holdover from trying to use slimmer images, can just copy it over again

resources/docker/Dockerfile.cgo
James-Gilbert-: Might be able to now with the duckdb update, I had a PR there that downgraded the C++ ABI for better compatibility.

James-Gilbert- (commented):
To clarify, is the goal of the build process to produce a fully static binary? Or is it to create an image with the extra CGO runtime dependencies?

gregfurman:
> Or is it to create an image with the extra CGO runtime dependencies?

@James-Gilbert- I suppose that since we don't release the CGO/x_bento_extra builds as a static binary, the goal would be to create an image with runtime dependencies baked in.

> To clarify, is the goal of the build process to produce a fully static binary?

Curious, would something like this be possible? Perhaps using `embed` with those C dependencies being statically linked. I'm no expert here so excuse the ignorance 🤷

gregfurman:
Also, lmk if you need me to re-build the new SQL documentation since it's failing in your CI 😄

James-Gilbert-:
> > Or is it to create an image with the extra CGO runtime dependencies?
> 
> @James-Gilbert- I suppose that since we don't release the CGO/x_bento_extra builds as a static binary, the goal would be to create an image with runtime dependencies baked in.
> 
> > To clarify, is the goal of the build process to produce a fully static binary?
> 
> Curious, would something like this be possible? Perhaps using `embed` with those C dependencies being statically linked. I'm no expert here so excuse the ignorance 🤷

I think it'd be possible to statically-cross compile, but with significant effort (probably some upstream effort w/DuckDB). Having a final layer with runtime dependencies is much less of a change.

resources/docker/Dockerfile.cgo
gregfurman: OK sick. Have these upstream changes been merged? Or should we address in a follow up

gregfurman:
> > > Or is it to create an image with the extra CGO runtime dependencies?
> 
> > 
> 
> > @James-Gilbert- I suppose that since we don't release the CGO/x_bento_extra builds as a static binary, the goal would be to create an image with runtime dependencies baked in.
> 
> > 
> 
> > > To clarify, is the goal of the build process to produce a fully static binary?
> 
> > 
> 
> > Curious, would something like this be possible? Perhaps using `embed` with those C dependencies being statically linked. I'm no expert here so excuse the ignorance 🤷
> 
> 
> 
> I think it'd be possible to statically-cross compile, but with significant effort (probably some upstream effort w/DuckDB). Having a final layer with runtime dependencies is much less of a change. 

Gotcha! Once you run a make docs and commit the generated files, it should unblock the CI 🟢 

Otherwise, lmk re: your capacity to address the remainder of image comments. I'm happy to also follow up if you'd prefer 🙂

James-Gilbert-:
It seems that because of the build tag `init()`, the "duckdb" driver string is dropped after re-running `make docs`

internal/impl/sql/conn_fields_init.go
gregfurman, line 9: This should fix the docs not building with the duckdb plugin while preserving a lint error when not using CGO 👇 

```suggestion
func init() {
	driverField = driverField.LintRule(`root = if this == "duckdb" { [ "Cannot use DuckDB driver outside of CGO build. Use a binary built with tag x_bento_extra, or the CGO Docker image" ] }`)
}
```

aronchick:
would really love to have this! lmk if i can help at all

internal/impl/sql/conn_fields_init.go
```diff
@@ -0,0 +1,9 @@
+//go:build !x_bento_extra
+
+package sql
+
+import "github.com/warpstreamlabs/bento/public/service"
```
iamramtin, line 5: Is this import being used?

public/components/sql/duckdb.go
```diff
@@ -0,0 +1,7 @@
+//go:build x_bento_extra
+
+package sql
+
+import (
+	_ "github.com/marcboeker/go-duckdb/v2"
```
iamramtin, line 6: Should we not be using `github.com/duckdb/duckdb-go/v2` instead? See https://duckdb.org/docs/stable/clients/go. I believe this is the new official repository for the driver maintained by the DuckDB  team.

internal/impl/sql/integration_extras_test.go
```diff
@@ -0,0 +1,48 @@
...
+
+	t.Cleanup(func() {
+		if db != nil {
+			if err := db.Close(); err != nil {
+				t.Errorf("error closing duckdb database: %v", err)
+			}
+		}
+	})
+
+	require.Eventually(t, func() bool {
```
iamramtin, line 31: DuckDB is embedded and should not need to wait for the ping? given that there's no network call it should just succeed on the first call, right?

if I'm not mistaken, I think you could just do `require.NoError(t, db.Ping())` instead

internal/impl/sql/integration_extras_test.go
```diff
@@ -0,0 +1,48 @@
...
+		return db.Ping() == nil
+	}, time.Second, 100*time.Millisecond, "duckdb did not respond to ping in time")
```
iamramtin, line 33: note, the test doesn't compile because `time` is not in the import block

```
❯ go test -tags x_bento_extra ./internal/impl/sql/...
ok      github.com/warpstreamlabs/bento/internal/impl/sql       1.895s
?       github.com/warpstreamlabs/bento/internal/impl/sql/aws   [no test files]
```

after adding it, your tests pass

internal/impl/sql/conn_fields_init.go
```diff
@@ -0,0 +1,9 @@
+//go:build !x_bento_extra
```
iamramtin, line 1: probably a nit, but the build tag here uses `!x_bento_extra` and not something like `!cgo`. Is there a scenario where a user sets `x_bento_extra` and `CGO_ENABLED=0`?

internal/impl/sql/conn_fields.go
iamramtin, line 1: Have you had a chance to test in-memory mode? DuckDB supports `:memory:` as a DSN. Some users may expect to use this for ephemeral states but it's not documented here

gregfurman:
Closing this in favour of https://github.com/warpstreamlabs/bento/pull/674.

Issue #366, opened 2025-06-11
[Feature] support Data stream write for OpenSearch output


arnitolog:
Apart from standard indices, OpenSearch has data streams that are intended for append-only, immutable data. Since data is immutable, only the `create` action is supported. 
It will be good to add this type to the action list.

I will create PR for that

PR #368, opened 2025-06-11
added create action to opensearch output

arnitolog:
### Changes:
- Adds `create` action to opensearch output to support write to data streams
- upgrades opensearch-go library to 4.5.0

### Tests:
- adds integration test for Data streams write

### Docs:
- updates documentation for opensearch output regarding supported actions

resolves https://github.com/warpstreamlabs/bento/issues/366

### DIFF
--- a/internal/impl/opensearch/integration_test.go
+++ b/internal/impl/opensearch/integration_test.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"strconv"
 	"strings"
 	"sync"
 	"testing"
@@ -182,6 +183,10 @@ func TestIntegration(t *testing.T) {
 
 	t.Run("TestOpenSearchDataStream", func(te *testing.T) {
 		testOpenSearchDataStream(urls, client, te)
+	})
+
+	t.Run("TestOpenSearchBatchCreate", func(te *testing.T) {
+		testOpenSearchBatchCreate(urls, client, te)
 	})
 }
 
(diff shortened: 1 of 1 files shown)


arnitolog:
@jem-davies , @gregfurman can you guys help me to understand what is wrong with tests? the first one is just panicked and lint is failing on the code that I didn't change

jem-davies:
> @jem-davies , @gregfurman can you guys help me to understand what is wrong with tests? the first one is just panicked and lint is failing on the code that I didn't change

Think periodically there are new rules added to the linter - and code you won't have change get flagged - looks like it was an easy fix, that's great. 

So the test is now passing - but it seems it's because you have removed the change made to the doc .md file. So the way the docs work - you don't update the .md files directly - you modify the the text in the .Description() of the ConfigSpec definition. Then you run `make docs` and the .md will be updated. 

From the ./CONTRIBUTING.md: 

If your change has an impact on documentation then make sure it is generated with make docs. You can test out the documentation site locally by running yarn && yarn start in the ./website directory.

internal/impl/opensearch/output.go
jem-davies, line 1: Update the doc strings in the ConfigSpec definition - then run `make docs` to generate the .md files.

jem-davies (dismissed):
will take another look after but just need to sort the docs out I think?

internal/impl/opensearch/output.go
arnitolog: updated, thanks @jem-davies

jem-davies:
@arnitolog - thanks for the contribution! - I just added a simple test I had written for my elasticsearch_v2 component

PR #369, opened 2025-06-12
feat: kafka aws msk iam sasl signer

gitphill:
Use github.com/aws/aws-msk-iam-sasl-signer-go/signer to support
'AWS_MSK_IAM' as a sasl mechanims for sarama kafka component.

The franz_kafka component already supports this authentication
mechanism. The franz-go library implements the AWS MSK SASL signer
itself based on the original Java library published by AWS.
See https://github.com/twmb/franz-go/blob/master/pkg/sasl/aws/aws.go.
This probably pre-dates aws-msk-iam-sasl-signer-go being available.

This change uses aws-msk-iam-sasl-signer-go implemented by AWS to
bring the sarama kafka component up to speed with franz_kafka.

Users will now be able to define the sarama kafka component like so:

```yaml
kafka:
  sasl:
    mechanism: AWS_MSK_IAM
      aws:
        region: '${AWS_REGION}'
        credentials:
          role: '${AWS_ROLE}'
```

See https://github.com/aws/aws-msk-iam-sasl-signer-go

Uses Bento Kafka AWS package to create an AWS session passing
its credentials provider to the SASL signer.


internal/impl/kafka/sasl.go
```diff
@@ -230,6 +236,26 @@ func SaramaSASLField() *service.ConfigField {
 		service.NewStringField(saramaFieldSASLTokenKey).
 			Description("Required when using a `token_cache`, the key to query the cache with for tokens.").
 			Default(""),
+		service.NewObjectField(saramaFieldSASLAws, []*service.ConfigField{
```
jem-davies: Is it not possible to use the config.SessionFields() from `"github.com/warpstreamlabs/bento/internal/impl/aws/config"` for this?

jem-davies (changes requested):
Had a quick scan - noticed that we have a ObjectField for AWS that we aren't using in the sasl mechanism for `kafka` but we are for `kafka_franz`

internal/impl/kafka/sasl.go
gitphill: I initially used this (and still can), however it includes properties that are not applicable in this case, things like:
* `endpoint` the aws lib defaults to "kafka.region.amazonaws.com" https://github.com/aws/aws-msk-iam-sasl-signer-go/blob/main/signer/msk_auth_token_provider.go#L37
* `credentials` -> `id`, `secret`, `token`, `from_ec2_role` none of which we can use to configure the behaviour of the sasl signer

So I only included the properties that are actually effective to avoid any confusion in the generated docs.

Session name is a property that doesn't exist in `internal/impl/aws/config` that could be presented as a config value. I left it as defaulting to https://github.com/aws/aws-msk-iam-sasl-signer-go/blob/main/signer/msk_auth_token_provider.go#L32 "MSKSASLDefaultSession". Maybe it would be useful to surface it as a configuration property though?

Happy to change this as you see fit

jem-davies: This makes more sense then 👍  - I'll get a proper look at this today

internal/impl/kafka/sasl.go
```diff
@@ -230,6 +236,26 @@ func SaramaSASLField() *service.ConfigField {
...
+					Description("A profile from `~/.aws/credentials` to use.").
+					Default("").Advanced(),
+				service.NewStringField("role").
+					Description("A role ARN to assume.").
+					Default("").Advanced(),
+				service.NewStringField("role_external_id").
+					Description("An external ID to provide when assuming a role.").
+					Default("").Advanced()).
+				Advanced().
+				Description("Optional manual configuration of AWS credentials to use. More information can be found [in this document](/docs/guides/cloud/aws)."),
```
jem-davies: Perhaps a bit nit picky - but maybe here we remove the link to the /docs/guides/cloud/aws page - because is a different object on that page: 

```yaml
credentials:
  profile: ""
  id: ""
  secret: ""
  token: ""
  role: ""
  role_external_id: ""
```

And we list the ways & order in which we try to find credentials like this:

```suggestion
				Description(`Optional manual configuration of AWS credentials to use. An Auth token will be made from AWS credentials using the following config fields:
 - ` + "`profile`" +
					`
 - ` + "`role` & `role_external_id`" +
					`
 - ` + "`role`" +
					`
 - ` + "Or leaving these fields blank will load AWS Credentials according to AWS's [LoadDefaultConfig](https://docs.aws.amazon.com/sdk-for-go/v2/developer-guide/configure-gosdk.html#specifying-credentials)"),
```

jem-davies (approved):
Looks great! Thanks for sorting so quickly.

internal/impl/kafka/sasl.go
jem-davies: If you decide to alter  - don't forget to `make docs`

jem-davies: And actually the information on the https://warpstreamlabs.github.io/bento/docs/guides/cloud/aws#none-of-these-fields-are-compulsory - regarding where we will look for credentials - I think ENV VARS actually take precedence over ~/.aws/credentials ...

jem-davies (approved):
LGTM - thanks 🎉

Issue #370, opened 2025-06-15
Rebuild and Rebrand Bloblang Editor as Embedded WASM Playground


iamramtin:
# Proposal: Rebuild and Rebrand Bloblang Editor as Embedded WASM Playground

## Purpose

The current Bloblang editor requires running `bento blobl server` locally, which creates unnecessary friction, limits discoverability, and presents a restricted editing experience with a minimal UI that doesn't reflect Bento’s updated branding. This proposal replaces it with a modern, browser-based Bloblang playground that:

- Runs entirely in-browser via WebAssembly
- Requires no local setup or backend
- Aligns with Bento’s new design system
- Integrates directly into the documentation site

## Goals

- Remove the dependency of having to run a local server to view the editor
- Rebrand the UI to use Bento’s design language
- Improve real-time mapping execution and error handling
- Embed the editor directly into the hosted Docusaurus documentation: `/docs/guides/bloblang/playground`
- Make the editor accessible, intuitive, and mobile-responsive

## Approach

This will be delivered in **multiple PRs** for easier review and maintenance:

This feature will be delivered across **multiple focused PRs** to keep each stage reviewable and atomic. Proposed breakdown:

### ✅ PR 1: Shiny new Bloblang editor & Enhanced Features ([#371](https://github.com/warpstreamlabs/bento/pull/371))

Merged.

- Replaced `<textarea>` inputs with **ACE editor**
- Applied **Bento design system** (layout, colors, typography)
- Implemented responsive **3-panel layout**
- Added **live execution with debounce**, **visual status indicators**, and **error highlighting**
- Enabled **import/export**, **copy-to-clipboard**, and **auto-formatting** for JSON
- Graceful fallback to `<textarea>` if ACE fails to load
- Other features (see PR linked above for more info)

### ✅ PR 2: Dark Mode, Full Syntax Linting & Dynamic Editor Context ([#380](https://github.com/warpstreamlabs/bento/pull/380))

Merged.

- **Smart Syntax Highlighting** built on CoffeeScript mode with Bloblang-specific overrides
- **Hover Tooltips** that show function signatures, stability tags, and example usage
- **Intelligent Autocompletion** with context-aware suggestions and prioritization
- **Bloblang Formatting Engine** for consistent syntax:
  - e.g. `a=1+2` → `a = 1 + 2`
- Fully integrated with Bento's design system for consistent visual theming

[...]

PR #380, opened 2025-06-23
Add syntax highlighting, autocompletion, and tooltips to Bloblang editor

iamramtin:
Transforms the Bloblang playground into a full-featured, IDE-like editor with smarter autocompletion, contextual docs, and robust formatting.

- **Smart Syntax Highlighting**: Leverages CoffeeScript mode as foundation with minimal Bloblang-specific overrides
- **Rich Documentation**: Hover tooltips display function specs, parameters, examples, and stability badges  
- **Formatting Engine**: Lightweight formatter ensures consistent spacing and structure on Bloblang syntax (e.g., a=1+2 → a = 1 + 2)
- **Intelligent Autocompletion**: Context-aware suggestions with signatures, types, and status-based prioritization
- **Cohesive Design**: Consistent Bento theme styling throughout editor and documentation interfaces

## Screenshots

image
image
image
image
image

References issue https://github.com/warpstreamlabs/bento/issues/370


internal/bloblang/syntax/rules.go
```diff
@@ -0,0 +1,76 @@
...
+
+// Token → semantic class used to apply styling.
+// Regex → JavaScript-compatible regex pattern used to match code tokens.
+type HighlightRule struct {
+	Token string `json:"token"`
+	Regex string `json:"regex"`
+}
+
+// BloblangSyntax contains everything needed for Bloblang syntax highlighting and autocompletion
+type BloblangSyntax struct {
```
gregfurman: Perhaps this `syntax` package should be moved closer to the server -- since AFAIK it is not used by any other component.

internal/bloblang/syntax/rules.go
```diff
@@ -0,0 +1,76 @@
...
+	"encoding/json"
+	"strings"
+
+	"github.com/warpstreamlabs/bento/internal/bloblang"
+	"github.com/warpstreamlabs/bento/internal/bloblang/query"
+)
+
+// Token → semantic class used to apply styling.
+// Regex → JavaScript-compatible regex pattern used to match code tokens.
+type HighlightRule struct {
```
gregfurman: nit: Does `HighlightRule` need to be public?

internal/bloblang/syntax/rules.go
```diff
@@ -0,0 +1,76 @@
...
+		functions[name] = spec
+		functionNames = append(functionNames, name)
+	})
+
+	env.WalkMethods(func(name string, spec query.MethodSpec) {
+		methods[name] = spec
+		methodNames = append(methodNames, name)
+	})
+
+	rules := createHighlightRules(functionNames, methodNames)
```
gregfurman: nit: if we're only ever calling `createHighlightRules` within this function, perhaps just set rules there i.e
and remove the original function.

```suggestion
	// Generate syntax highlighting rules
	rules := []HighlightRule{
		// Matches `root` (highest priority)
		{Token: "bloblang_root", Regex: `\broot\b`},
		// Matches `this` (highest priority)
		{Token: "bloblang_this", Regex: `\bthis\b`},
		// Matches `foo()`
		{Token: "support.function", Regex: `(?<![\.\w])(` + strings.Join(functionNames, "|") + `)(?=\s*\()`},
		// Matches `.bar()`
		{Token: "support.method", Regex: `\.(` + strings.Join(methodNames, "|") + `)(?=\s*\()`},
	}
```

internal/cli/blobl/resources/playground/js/playground.js
```diff
@@ -17,40 +17,43 @@ class BloblangPlayground {
       toggleFormatInputBtn: document.getElementById("toggleFormatInputBtn"),
       toggleFormatOutputBtn: document.getElementById("toggleFormatOutputBtn"),
     };
+
+    this.editor = new EditorManager();
```
gregfurman, line 21: question: Should this be imported or is it fine for it to live in another file? I can't javascript btw so pardon my ignorance

internal/cli/blobl/server.go
```diff
@@ -229,6 +230,19 @@ func runServer(c *cli.Context) error {
 		}
 	})
 
+	mux.HandleFunc("/bloblang-syntax", func(w http.ResponseWriter, r *http.Request) {
+		response, err := syntax.GenerateBloblangSyntax()
+		if err != nil {
+			http.Error(w, "Failed to generate Bloblang syntax", http.StatusInternalServerError)
+			return
+		}
```
gregfurman: question: If this fails to render, should we be panicking? Or is it fine to just return an error.

internal/bloblang/syntax/rules.go
```diff
@@ -0,0 +1,76 @@
...
+
+	syntax := BloblangSyntax{
+		Functions:     functions,
+		Methods:       methods,
+		Rules:         rules,
+		FunctionNames: functionNames,
+		MethodNames:   methodNames,
+	}
+
+	return json.Marshal(syntax)
```
gregfurman: IMO rather return a struct here and allow the caller function to handle the marshalling to JSON.

internal/bloblang/syntax/rules.go
```diff
@@ -0,0 +1,76 @@
...
+	// Minimal syntax highlighting rules that extend CoffeeScript
+	Rules []HighlightRule `json:"rules"`
+
+	// Quick lookup arrays for regex generation (internal use)
+	FunctionNames []string `json:"function_names"`
+	MethodNames   []string `json:"method_names"`
+}
+
+func GenerateBloblangSyntax() ([]byte, error) {
+	env := bloblang.GlobalEnvironment()
```
gregfurman: suggestion: Let's rather use an interface here and pass in the bloblang environment at the caller:

```suggestion
type Walker interface {
	WalkFunctions(fn func(name string, spec query.FunctionSpec))
	WalkMethods(fn func(name string, spec query.MethodSpec))
}

func GenerateBloblangSyntax(env Walker) BloblangSyntax {
```

internal/cli/blobl/resources/playground/js/editor.js
```diff
@@ -1,219 +1,508 @@
...
+    } catch (error) {
+      console.warn("Autocompletion disabled:", error);
     }
-    return document.getElementById("fallbackInput").value;
   }
 
-  getMapping() {
-    if (this.aceMappingEditor) {
-      return this.aceMappingEditor.getValue();
+  createBloblangCompleter() {
```
gregfurman, line 227: Can we disable suggestions after a comment? Am getting a suggestion despite being in a commented out block.

image

internal/cli/blobl/resources/playground/js/editor.js
```diff
@@ -1,219 +1,508 @@
...
+      : "";
+
+    let html = `
+      <div class="ace-doc">
+        <div class="ace-doc-signature">
+          <strong>${signature}</strong>
+          ${statusBadge}
+        </div>
+        <div class="ace-doc-description">${
+          functionSpec.description || "No description available"
```
gregfurman: Any chance we can try render the markdown within these descriptions? Namely when referencing URLs and stuff. Also, if we could get [docusaurus admonitions](https://docusaurus.io/docs/markdown-features/admonitions) to work that'd be great. Otherwise, perhaps we should strip them before passing in the data i.e

image

internal/cli/blobl/server.go
```diff
@@ -229,6 +230,19 @@ func runServer(c *cli.Context) error {
 		}
 	})
 
+	mux.HandleFunc("/bloblang-syntax", func(w http.ResponseWriter, r *http.Request) {
+		response, err := syntax.GenerateBloblangSyntax()
```
gregfurman: I think we should be generating this once and passing it in via the `template.Must`. We can set some go templating it in the `playground.html` and pass in data on startup. Something like:

In `playground.html`
```html
    <!-- Inject in Bloblang syntax rules -->
    <script>
      window.BLOBLANG_SYNTAX = {{.BloblangSyntax}};
    </script>
```

then in `editor.js`
```js
  async loadBloblangSyntax() {
    try {
      this.bloblangSyntax = window.BLOBLANG_SYNTAX;
    }
```

Then in the `mux.HandleFunc("/", ...)`

```go
		if err := indexTemplate.Execute(w, struct {
			InitialInput   string
			InitialMapping string
			BloblangSyntax template.JS
		}{
			fSync.input(),
			fSync.mapping(),
			template.JS(resBloblSyntax),
		}); err != nil {
			http.Error(w, "Template error", http.StatusBadGateway)
			return
		}
```

internal/bloblang/syntax/rules.go
iamramtin: It's only being used by the `server.go` file but  I didn't want to bloat it with more functionality. Any suggestions where to put it?

internal/cli/blobl/resources/playground/js/playground.js
iamramtin: These files are included using <script> tags in the HTML file for the editor, so they don't need to be explicitly imported and exported like in a Node.js project. As long as they are referenced in the correct order, the browser will automatically load and execute them, making dependencies work as expected.

internal/bloblang/syntax/rules.go
gregfurman: I rate just add it as a syntax.go alongside the server.go. Maybe we can add a playground package there if you think we should be extending our tooling?

internal/cli/blobl/server.go
iamramtin: I'm still relatively new to using Go, but I imagine it makes more sense to just return an error in this case? I set this up to just aid and enhance the editor, should this fail it just won't provide the full syntax highlighting and documentation on the Bloblang methods and functions, so I don't think this would need to necessarily panic

internal/cli/blobl/resources/playground/js/editor.js
iamramtin: Good call, missed this!

internal/cli/blobl/resources/playground/js/editor.js
iamramtin: I didn't realise admonitions were part of the descriptions, I think it's best to strip them as I've already added a badge representing the same status/type of the function or method

internal/cli/blobl/server.go
iamramtin: Thanks for the recc, that's a good idea, I've addressed this in my latest commit

internal/cli/blobl/resources/playground/index.html
```diff
@@ -249,6 +249,11 @@
       defer
     ></script>
 
+    <!-- Inject the Bloblang syntax into the global JS context -->
+    <script>
+      window.BLOBLANG_SYNTAX = {{.BloblangSyntax}};
```
gregfurman, line 254: Nice! 😎

internal/cli/blobl/resources/playground/js/editor.js
```diff
@@ -122,9 +107,16 @@ class EditorManager {
...
 
     this.aceMappingEditor = ace.edit(mappingEl);
-    this.aceMappingEditor.setValue(this.defaultMapping, 1);
+    const contentWithNewline = this.defaultMapping.trimEnd() + "\n"; // Trailing newline
+    this.aceMappingEditor.setValue(contentWithNewline, 1);
     this.aceMappingEditor.session.setMode(MODE_BLOBLANG);
     this.aceMappingEditor.setTheme(THEME_BENTO);
+    this.aceMappingEditor.commands.addCommand({
+      name: "formatBloblang",
+      bindKey: { mac: "Cmd-Shift-F", win: "Ctrl-Shift-F" },
```
gregfurman, line 116: Lol this is great

internal/cli/blobl/resources/playground/js/editor.js
```diff
@@ -175,10 +167,44 @@ class EditorManager {
     });
   }
 
+  patchCommentShortcut(editor) {
```
gregfurman: suggestion: Can you perhaps add a comment on what this is doing?

internal/cli/blobl/resources/playground/js/editor.js
```diff
@@ -198,248 +225,176 @@ class EditorManager {
...
-  getMethodCategory(methodSpec) {
-    if (methodSpec.categories && methodSpec.categories.length > 0) {
-      return methodSpec.categories[0].category;
-    }
+  // Categorize by name patterns
+  getMetaCategory({ name, status }) {
+    if (status === "deprecated") return "Deprecated";
+    if (status === "experimental") return "Experimental";
+
+    const patterns = {
```
gregfurman, line 309: We should _probably_ be dynamically generating these values

internal/cli/blobl/server.go
```diff
@@ -23,19 +23,30 @@ import (
...
+}
+
 //go:embed resources/playground
 var playgroundFS embed.FS
 
 var bloblangPlaygroundPage string
 
 func init() {
-	page, err := playgroundFS.ReadFile("resources/playground/playground.html")
+	page, err := playgroundFS.ReadFile("resources/playground/index.html")
```
gregfurman, line 47: praise: Ahh good catch. Now that the old `index.html` is gone renaming this is better 👍

PR #383, opened 2025-06-25
fix-aws-cloudwatch-metric-shutdown

jem-davies:
Fixes an issue where `aws_cloudwatch` metrics were not sent on Close().


internal/impl/aws/metrics_cloudwatch_test.go
```diff
@@ -379,7 +388,52 @@ func TestCloudWatchTagsMoreThan20(t *testing.T) {
...
+
+	ctrFoo := cw.NewCounterCtor("counter.foo", tagNames...)
+	ctrFoo(tagValues...).Incr(3)
+
+	go cw.loop()
+
+	time.Sleep(time.Millisecond * 100)
+
+	ctx := context.Background()
+	cw.cancel()
```
jem-davies: maybe don't need to call this because it should be called by Cancel()

PR #384, opened 2025-06-25
fix-aws-cloudwatch-metric-shutdown

jem-davies:


internal/impl/aws/metrics_cloudwatch.go
```diff
@@ -523,7 +523,9 @@ func (c *cwMetrics) HandlerFunc() http.HandlerFunc {
 }
 
 func (c *cwMetrics) Close(ctx context.Context) error {
-	c.cancel()
-	c.flush()
+	defer c.cancel()
+	if err := c.flush(); err != nil {
+		return err
+	}
```
gregfurman: This doesn't account for if we want to cancel early. Like, this implementation will _always_ be graceful, where ignoring the parent/closer context could cause some shutdown issues.

gregfurman: I think we should use an `errgroup` instead. That way, if the parent `ctx` is cancelled we can return early (a hard-stop), else if the flush returns early then we can also gracefully shutdown (a soft-stop).


```suggestion
	defer c.cancel()

	eg, ctx := errgroup.WithContext(ctx)
	eg.Go(c.flush)

	return eg.Wait()
```

gregfurman (approved):
LGTM. Can you fix the commit message? Happy for you to merge once that's done 😄

internal/impl/aws/metrics_cloudwatch.go
jem-davies: So the function will be: 

```go

func (c *cwMetrics) Close(ctx context.Context) error {
	defer c.cancel()
	eg, _ := errgroup.WithContext(ctx)
	eg.Go(c.flush)
	return eg.Wait()
}
```

?

gregfurman: Yeah that makes sense. We can just discard the new derived context since it's unused anyway.

PR #393, opened 2025-07-05
Integrate Bloblang Playground into Bento Website

iamramtin:
Add a browser-based Bloblang playground powered by WebAssembly. The playground is embedded directly into the Bento docs and runs fully client-side, requiring no backend.

- WASM runtime with gzip-compressed binary.
- Client-side execution, no running server needed.
- Use LocalStorage for editor state persistence
- Website integration via iframe, with full-screen mode and usage guide.
- GitHub Actions step to build, compress, and deploy WASM assets.

### 🔹 Method 1: Go Server (Native Bloblang Runtime)

```sh
bento blobl server

# Open in browser
open http://localhost:4195
```

### 🔹 Method 2: Standalone WASM Playground

```sh
cd website

# Build the WASM playground
yarn build::playground

# Serve it statically
cd static/playground
python3 -m http.server 8000

# Open in browser
open http://localhost:8000
```

### 🔹 Method 3: Bento Docs Integration (WASM Embedded in Docusaurus)

```sh
cd website

yarn build::playground
yarn build
yarn start

# Open in browser
open http://localhost:3000/bento/docs/guides/bloblang/playground
```

### Light Mode:
Screenshot 2025-07-05 at 11 18 18
Screenshot 2025-07-05 at 11 18 04

### Dark Mode:
Screenshot 2025-07-05 at 11 17 36
Screenshot 2025-07-05 at 11 17 52

References issue https://github.com/warpstreamlabs/bento/issues/370


iamramtin:
For deployment, I’ve added a Makefile rule that compiles the WASM binary and moves all static assets (HTML, JS, CSS) into the website folder. It relies only on Go and standard shell utilities. I haven’t been able to test the CI integration myself due to lack of access, but it should be straightforward for a maintainer to verify on a branch.

Let me know if anything needs adjusting or if I've missed something.

iamramtin:
Another note: Running `website/build_playground.sh` will produce the WASM binary and move it along with the static playground files into the website folder. For ease of reviewing, these generated files haven't been committed.

internal/cli/blobl/shared.go
```diff
@@ -0,0 +1,111 @@
...
+		result.MappingError = "Input JSON string cannot be empty"
+		return result, nil
+	}
+
+	if mapping == "" {
+		result.ParseError = "Mapping string cannot be empty"
+		return result, nil
+	}
+
+	exec, err := p.env.NewMapping(mapping)
```
gregfurman: Why the new `BloblangEnvironment` type is we can just call this `NewMapping` using the original mapping type? 

Also, does the `newExecCache` not suffice for creating/using an execution mapping?

internal/config/watcher_wasm.go
```diff
@@ -8,6 +8,37 @@ import (
...
+// Unwrap the underlying error.
+func (e *ErrNoReread) Unwrap() error {
+	return e.wrapped
+}
+
+// Error returns a human readable error string.
+func (e *ErrNoReread) Error() string {
+	return e.wrapped.Error()
+}
+
```
gregfurman: Doesn't look like any of these are used. How come they were added?

internal/cli/blobl/shared.go
```diff
@@ -0,0 +1,111 @@
...
+	if err != nil {
+		result.MappingError = fmt.Sprintf("execution error: %v", err.Error())
+	} else {
+		result.Result = output
+	}
+
+	return result, nil
+}
+
+func (p *BloblangEnvironment) ValidateMapping(mapping string) (*ValidationResult, error) {
```
gregfurman: Is this used?

internal/cli/blobl/shared.go
```diff
@@ -0,0 +1,111 @@
...
+			result.ParseError = fmt.Sprintf("mapping error: %v", err.Error())
+		}
+	}
+
+	return result, nil
+}
+
+func (p *BloblangEnvironment) GetSyntax() (bloblangSyntax, error) {
+	return GenerateBloblangSyntax(p.env)
+}
```
gregfurman: Also unused ☝️

cmd/playground/main.go
```diff
@@ -0,0 +1,95 @@
...
+	// Expose JavaScript globals
+	js.Global().Set("executeBloblang", js.FuncOf(executeBloblang))
+	js.Global().Set("getBloblangSyntax", js.FuncOf(getBloblangSyntax))
+	js.Global().Set("wasmReady", js.ValueOf(true))
+	js.Global().Set("wasmMetadata", map[string]any{
+		"built": time.Now().UTC().Format(time.RFC3339),
+	})
+
+	// Keep the Go runtime alive to handle JS calls
+	select {}
```
gregfurman: Much better way to block than an empty for-loop without an exit condition 💡 

However, this will make it impossible to ever stop the WASM binary running -- since exit signals etc. will be ignored. Can you explain why we need to block forever here and perhaps we can consider a different solution?

internal/cli/blobl/shared.go
```diff
@@ -0,0 +1,111 @@
...
+	env *bloblang.Environment
+}
+
+func NewBloblangEnvironment() *BloblangEnvironment {
+	return &BloblangEnvironment{
+		env: bloblang.GlobalEnvironment(),
+	}
+}
+
+func (p *BloblangEnvironment) ExecuteMapping(input, mapping string) (*ExecutionResult, error) {
```
gregfurman: There's a `(*ExecutionResult, error)` in the method signature but we never actually return an error. Makes me think this `ExecutionResult` abstraction could be redunant -- where we return an (optional) result and an error instead of a struct containing a result and error.

internal/config/watcher_wasm.go
iamramtin: I copied these from `watcher.go` as they're implemented there for non-wasm builds as per the build tag.

It turns out I actually need only the `noReread` function, I'll update accordingly for now. But I think the reason for this is because the build script I wrote lies in the `website` directory as opposed to the root directory. Running `GOOS=js GOARCH=wasm go build ...` in its current state and location produces this error.

```
# github.com/warpstreamlabs/bento/internal/config
../internal/config/reader.go:352:11: undefined: noReread
../internal/config/reader.go:365:10: undefined: noReread
../internal/config/reader.go:381:10: undefined: noReread
../internal/config/resource_reader.go:251:10: undefined: noReread
../internal/config/resource_reader.go:268:10: undefined: noReread
../internal/config/stream_reader.go:209:10: undefined: noReread
../internal/config/stream_reader.go:231:10: undefined: noReread
```

It probably makes sense to rethink this script, as it currently needs to build the WASM binary from the static files in root but it also needs to copy them over to the `website` directory to be properly embedded. We also need to consider convenience of running this both locally and as a part of the GitHub Actions workflow. I'll give it some thought, but open to any ideas

cmd/playground/main.go
iamramtin: From my understanding (and testing), if we dont block to keep the runtime alive the Go functionality gets compiled to WASM and the main() function runs but then exits immediately and the functions we register for javascript are unavailable.

I believe this is the expected behaviour for browser-based WASM implementations. I'm looking into utilsing [TinyGo](https://github.com/tinygo-org/tinygo) as an alternative which seemingly handles this differently (and offers other advantages), if there's a clear advantage, I'll switch over

internal/config/watcher_wasm.go
iamramtin: So I've addressed this in my latest commit by reducing the changes I'd made to `watcher_wasm.go`. But it appears my initial impression wasn't entirely correct.

The playground uses `bloblang.GlobalEnvironment()` which includes **all** registered Bloblang functions, not just the ones needed for basic playground functionality. and `bloblang.GlobalEnvironment()` has functionality which pulls in `github.com/warpstreamlabs/bento/internal/config` and since the environment is being used by the WASM implementation, I think this change is required

internal/cli/blobl/wasm.go
```diff
@@ -0,0 +1,72 @@
...
+// Arguments:
+//   - args[0]: input JSON string
+//   - args[1]: Bloblang mapping string
+//
+// Returns a JS object with:
+//   - "result":        the mapping result (any type, or nil on error)
+//   - "parse_error":   error message if input JSON could not be parsed, else nil
+//   - "mapping_error": error message if mapping failed, else nil
+
+func ExecuteBloblangMapiing() js.Func {
```
gregfurman: nit: spelling error

cmd/tools/playground/main.go
```diff
@@ -0,0 +1,35 @@
+//go:build wasm
+
+package main
```
gregfurman, line 3: Can we embed the `lib/wasm/wasm_exec.js` file using `embed`? They should be in `"$(go env GOROOT)/lib/wasm/wasm_exec.js"`

cmd/tools/playground/main.go
```diff
@@ -0,0 +1,22 @@
...
+
+import (
+	"syscall/js"
+
+	"github.com/warpstreamlabs/bento/internal/cli/blobl"
```
gregfurman, line 8: We should import the `github.com/warpstreamlabs/bento/public/components/pure` since there's a lot of blobl functionality that wont be supported in the editor if omitted (including your playground examples):

```suggestion
	"github.com/warpstreamlabs/bento/internal/cli/blobl"
	_ "github.com/warpstreamlabs/bento/public/components/pure"
```

Also, this will require removing those build tags in `internal/cli/blobl/cli.go` and `internal/cli/blobl/server.go` -- increasing the binary size to 35MB.

internal/cli/blobl/playground/index.html
```diff
@@ -260,15 +255,23 @@
...
+    <script src="./js/wasm_exec.js"></script>
+    {{end}}
+    <!-- BEGIN SERVER MODE -->
+    <!-- Go server injects the Bloblang syntax and initial values into the global JS context -->
+    <script>
+      window.BLOBLANG_SYNTAX = {{.BloblangSyntax}};
+      window.INITIAL_INPUT = {{.InitialInput}};
+      window.INITIAL_MAPPING = {{.InitialMapping}};
+    </script>
+    <!-- END SERVER MODE -->
```
gregfurman:
```suggestion
    {{if .WasmMode}}
    <script src="./js/wasm.js" defer></script>
    <script src="./js/wasm_exec.js"></script>
    {{else}}
    <!-- BEGIN SERVER MODE -->
    <!-- Go server injects the Bloblang syntax and initial values into the global JS context -->
    <script>
      window.BLOBLANG_SYNTAX = {{.BloblangSyntax}};
      window.INITIAL_INPUT = {{.InitialInput}};
      window.INITIAL_MAPPING = {{.InitialMapping}};
    </script>
     {{end}}
    <!-- END SERVER MODE -->
```

cmd/tools/playground/main.go
iamramtin: Ah, my bad, thanks for picking this up. I've fixed it!

internal/cli/blobl/cli.go
```diff
@@ -67,156 +61,56 @@ Find out more about Bloblang at: {{.DocumentationURL}}/guides/bloblang/about`)[1
...
-						Usage:   "an optional path to an input file to load as the initial input to the mapping within the app.",
+						Aliases: []string{"m"},
+						Usage:   "preload a Bloblang mapping file into the playground editor",
 					},
 					&cli.BoolFlag{
 						Name:    "write",
 						Value:   false,
 						Aliases: []string{"w"},
-						Usage:   "when editing a mapping and/or input file write changes made back to the respective source file, if the file does not exist it will be created.",
+						Usage:   "auto-save playground changes back to the source files",
```
gregfurman: nit: Do we want fullstops here?

PR #394, opened 2025-07-06
Azure cloud credentials sql

jem-davies:
Adds `azure` fields to sql components (similar to the aws fields that enable iam auth) that enable auth to postgres running on azure with "entra auth".

### DIFF
--- a/internal/impl/sql/conn_fields.go
+++ b/internal/impl/sql/conn_fields.go
@@ -115,6 +115,43 @@ CREATE TABLE IF NOT EXISTS some_table (
 			Default(false).
 			Version("1.8.0").
 			Advanced(),
+		service.NewObjectField("azure",
+			service.NewBoolField("entra_enabled").
+				Description("An optional field used to generate an entra token to connect to 'Azure Database for PostgreSQL flexible server', This will create a new connection string with the host, user and database from the DSN field - you may need to URL encode the dsn! The [Default Azure Credential Chain](https://learn.microsoft.com/en-gb/azure/developer/go/sdk/authentication/authentication-overview#defaultazurecredential) is used from the Azure SDK.").
+				Optional().
+				Default(false).
+				Version("1.10.0").
+				Advanced(),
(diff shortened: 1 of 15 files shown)


internal/impl/azure/config/config.go
```diff
@@ -0,0 +1,15 @@
+package config
+
+import "github.com/warpstreamlabs/bento/public/service"
+
+// AzureSessionFields defines a re-usable set of config fields for an Azure
+// session that is compatible with the publci service APIs and avoids importing
```
jem-davies:
```suggestion
// session that is compatible with the public service APIs and avoids importing
```

adrianhaj (approved):
LGTM thank you!

website/sidebars.js
```diff
@@ -116,6 +116,7 @@ module.exports = {
           items: [
             'guides/cloud/aws',
             'guides/cloud/gcp',
+            'guides/cloud/azure',
```
jem-davies:
```suggestion
```

adrianhaj:
@jem-davies I tested this on my azure aks cluster with Microsoft Entra and works!

jem-davies:
> @jem-davies I tested this on my azure aks cluster with Microsoft Entra and works!

ahh great thanks a lot! @gregfurman could you give this a 👀  please soon 🙏

internal/impl/sql/conn_fields.go
```diff
@@ -304,6 +347,15 @@ func sqlOpenWithReworks(ctx context.Context, logger *service.Logger, driver, dsn
...
 
+	updatedDSN, err = BuildAzureDsn(dsn, driver)
+	if err != nil {
+		return nil, err
+	}
+
+	if updatedDSN != dsn {
+		logger.Info("Updated dsn with info from Azure")
+	}
+
```
gregfurman, line 358: Can we not rather replace `BuildAwsDsn` with `BuildDsn` and conditionally (depending on whether Azure or AWS) is used we replace it?

Issue #397, opened 2025-07-09
zeromqn exception cannot reconnect


huangpeizhi2018:
# input_zeromqn_01.yaml
# Receive the data sent by output_zeromqn_01.yaml and print it
```
http:
  enabled: false
  debug_endpoints: false

logger:
  level: TRACE
  format: logfmt
  add_timestamp: true
  static_fields:
    '@service': input_zeromqn_01

input:
  broker:
    inputs:
      - zmq4n:
          urls:
            - tcp://0.0.0.0:4501
          bind: true
          socket_type: "PULL"

pipeline:
  processors:
    - mapping: |
        root = this

output:
  drop: {}
  processors:
    - log:
        level: INFO
        message: "raw:: ${! content()}"
```

# output_zeromqn_01.yaml
# Generate data and output to zeromq
```
http:
  enabled: false
  debug_endpoints: false

logger:
  level: INFO
  format: logfmt
  add_timestamp: true
  static_fields:
    '@service': output_zeromqn_01

input:
  generate:
    count: 300
    interval: "2s"
    mapping: |
      root = {"id": counter(),"parentId": "2548433071355506235"}

pipeline:
  processors:
    - mapping: |
        root = this

output:
  retry:
    max_retries: 30
    backoff:
      initial_interval: 500ms
      max_interval: 3s
      max_elapsed_time: 0s
    output:
      zmq4n:
        urls:
          - tcp://localhost:4501
        bind: false
        socket_type: "PUSH"
        poll_timeout: 10m
```

After restarting the input_zeromqn_01 process, the output_zeromqn_01 connection cannot be reestablished.

gregfurman:
Hey. Looks like the problem here has to do with the underlying `go-zeromq` implementation. Seems that underlying connection socket connection gets reaped after 10 failed retries and never re-established. 

In addition, the send message method does not repeatedly fail if no connection exists -- making the failures silent when (in fact) a re-connection is required:
https://github.com/warpstreamlabs/bento/blob/24c8cb584b6fa30b314b091a5d47c2a31deaf05e/internal/impl/zeromq/native_output_zmq4n.go#L178

I've opened a PR to fix this which (when merged) means you can drop the `retry` component and all should work fine 😄

gregfurman:
@huangpeizhi2018 If you use the latest build of Bento, this should be fixed. This will be included and added in `v1.10` 👍 

If possible, could you build and run bento (from HEAD of `main`) and see if all works fine now?

[...]

PR #402, opened 2025-07-14
add(zmq4n): Handle socket and component re-connections

gregfurman:
## Motivation

Fixes https://github.com/warpstreamlabs/bento/issues/397 where the output component fails to re-connect when a target socket has been closed and re-opened.

## Changes

- Add options for configuring internal socket dial reconnection attempts
- Trigger bento stream re-connection attempts when internal retries fail via `ErrNotConnected`.


jem-davies (approved):
LGTM

PR #404, opened 2025-07-14
add MockResourcesOptUseSlogger func 

jem-davies:
add MockResourcesOptUseSlogger func to enable mocking logger when using MockResources

### DIFF
--- a/public/service/resources.go
+++ b/public/service/resources.go
@@ -3,6 +3,7 @@ package service
 import (
 	"context"
 	"io/fs"
+	"log/slog"
 	"time"
 
 	"go.opentelemetry.io/otel/trace"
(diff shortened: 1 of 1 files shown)


public/service/resources.go
```diff
@@ -50,6 +52,14 @@ func MockResourcesOptUseLogger(l *Logger) MockResourcesOptFn {
 	}
 }
 
+func MockResourcesOptUseSlogger(l *slog.Logger) MockResourcesOptFn {
```
gregfurman, line 57: Perhaps just add a doc/comment (since this is exposed in the public SDK)

gregfurman, line 57: Perhaps just add a doc/comment (since this is exposed in the public SDK)

PR #406, opened 2025-07-20
Add option to start new root span with link to remote context

soluchok:


jem-davies:
@soluchok - I see this PR is marked as draft - is it still being worked on? 

The docs can be generated by running `make docs` - it is mentioned on the CONTRIBUTING.md file

gregfurman:
Hey @soluchok 👋 Will give this a look a bit later today. Apologies for the delay.

internal/impl/nats/docs.go
```diff
@@ -24,9 +24,13 @@ NATS component, so that monitoring tools between NATS and bento can stay in sync
 `
 }
 
-func inputTracingDocs() *service.ConfigField {
-	return service.NewExtractTracingSpanMappingField().Version(tracingVersion)
+func inputTracingDocs() []*service.ConfigField {
+	return []*service.ConfigField{
+		service.NewExtractTracingSpanMappingField().Version(tracingVersion),
```
gregfurman, line 29: Can we change `tracingVersion` here to `v1.0.0` and your addition to `v1.13.0` (the upcoming release)? Seems the `tracingVersion` is still referencing a pre-fork version that isn't relevant here 😃

public/service/config_extract_tracing.go
```diff
@@ -27,18 +29,33 @@ func NewExtractTracingSpanMappingField() *ConfigField {
 		Advanced()
 }
 
+func NewRootSpanWithLinkField() *ConfigField {
+	return NewBoolField(nrswlField).
+		Description("EXPERIMENTAL: Starts a new root span with link to parent.").
+		Version("1.0.0").
+		Optional().
```
gregfurman, line 36: **suggestion**: Let's add a `Default` of `false` here to allow no value being set to pass config validations.

public/service/config_extract_tracing.go
```diff
@@ -27,18 +29,33 @@ func NewExtractTracingSpanMappingField() *ConfigField {
...
 	exe, err := p.FieldBloblang(etsField)
 	if err != nil {
 		return nil, err
 	}
-	return &spanInjectBatchInput{inputName: inputName, mgr: p.mgr, mapping: exe, rdr: i}, nil
+
+	newSpan, err := p.FieldBool(nrswlField)
+	if err != nil && !strings.Contains(err.Error(), "was not found in the config") {
+		return nil, err
+	}
```
gregfurman, line 56: Since this is a bool field, if we error out here you _could_ just ignore the error since it'll default to false i.e

```suggestion
	newSpan, _ := p.FieldBool(nrswlField)
	if err != nil {
		return nil, err
	}
```

Properly configuring the config to have a default means this check would probably fall away. Else, since this is optional, we can add a `p.Contains(nrswlField)` and only do this check if the field exists in the parsed config 

see example

https://github.com/warpstreamlabs/bento/blob/f56fc967c81bdf55de02f4597f14f30b7c4218e0/internal/httpclient/auth_oauth2.go#L115-L117

and (probably incorrectly formatted) suggestion 👇 

```suggestion
    var newSpan bool
	if p.Contains(nrswlField) {
		if err != nil {
		   return nil, err
	   }
	}
```

gregfurman:
Hey @soluchok 👋 Do you no longer have have capacity to address my review? I opened a PR in your fork since I don't have permissions to push to your feature branch directly. 

Otherwise, I'm happy to close this PR, cherry-pick your commit, and open another with the new changes included. Lmk!

jem-davies:
#605

PR #407, opened 2025-07-21
csv: enforce expected_headers only when header parsing is enabled (#357)

### LINKED ISSUE #357: CSV Scanner and Input Lint Rule
```
input:
  file:
    paths: [./tmp/csv_expected_headers/data.csv]
    scanner:
      csv:
        parse_header_row: false
        expected_headers: [first_sname, last_name, age]
        expected_number_of_fields: 3
```

Current behaviour of the above config will not check or parse the first row as headers, `parse_header_row: false` that is false but perhaps would be good to add a lint rule to avoid setting `parse_header_row: false` and setting `expected_headers`

rohankumardubey:
This change adds a runtime guard in csvScannerFromParsed that returns an error if `parse_header_row` is set to false while `expected_headers` is non‐empty. It prevents silent misconfiguration where headers won’t be checked but you’ve asked for specific ones.

- Add error in csvScannerFromParsed when !parseHeaderRow && len(expectedHeaders)>0
- Remove unsupported cross‐field LintRule DSL for expected_headers/parse_header_row
- Add TestCSVScannerError_HeaderDisabledWithExpectedHeaders to verify the guard
- Add TestCSVScannerSuccess_HeaderEnabledWithExpectedHeaders to ensure the happy path

Closes #357.

### DIFF
--- a/internal/impl/pure/scanner_csv_test.go
+++ b/internal/impl/pure/scanner_csv_test.go
@@ -192,3 +192,49 @@ a4,b4,c4
 	}, &service.ScannerSourceDetails{})
 	require.ErrorContains(t, err, "wrong number of fields")
 }
+
+// Ensures CSV scanner fails when header parsing is disabled but expected_headers is set.
+func TestCSVScannerError_HeaderDisabledWithExpectedHeaders(t *testing.T) {
+	confSpec := service.NewConfigSpec().Field(service.NewScannerField("test"))
+	pConf, err := confSpec.ParseYAML(`
+test:
+  csv:
+    parse_header_row: false
+    expected_headers:
+      - a
+      - b
+      - c
+`, nil)
+	require.NoError(t, err, "parsing YAML itself should succeed")
+
+	// Now building the scanner must fail
+	_, err = pConf.FieldScanner("test")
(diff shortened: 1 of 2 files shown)


internal/impl/pure/scanner_csv_test.go
```diff
@@ -192,3 +192,49 @@ a4,b4,c4
...
+	pConf, err := confSpec.ParseYAML(`
+test:
+  csv:
+    parse_header_row: false
+    expected_headers:
+      - a
+      - b
+      - c
+`, nil)
+	require.NoError(t, err, "parsing YAML itself should succeed")
```
jem-davies: Hi thanks for the contribution! The intention for the ticket was to use a [LintRule](https://pkg.go.dev/github.com/warpstreamlabs/bento/v4/public/service#ConfigSpec.LintRule) - that way we can catch the misconfiguration at this stage.

rohankumardubey: Gotcha. Added a top‑level LintRule in csvScannerSpec that flags any config where parse_header_row=false and expected_headers is non‑empty.

jem-davies:
@rohankumardubey thanks a lot 😃

PR #410, opened 2025-07-24
feat(aws_dynamodb): add conditional DeleteItem support #408

### LINKED ISSUE #408: Adding DeleteItem functionality to aws_dynamodb
# missing feature
The aws_dyanmodb output module supports PutItem batching. There is no way to batch delete items. This requires support for DeleteRequest to be added to the batch write.

# solution suggestion
A DeleteRequest fed to the batch API takes the Partition Key and Sort Key as input in a Key object, so those would need to be optionally specified along with a parameter to indicate if it is a delete request.

## New fields
- **is_delete**: metadata("delete_this_record") == True # new field - (boolean, default: False) to indicate if the record should be deleted, Bloblang query like in switch check statement
- **partition_key**: myPk # new field - name of the partition key (string, no default)
- **sort_key**: mySk # new field - name of the sort key (string, no default)

## Example Config
```yaml
processors:
  - mapping: |
        
output:
  - aws_dynamodb:
        backoff:
            initial_interval: 1s
            max_elapsed_time: 30s
            max_interval: 5s
        batching:
            count: 1000
            jitter: 0.1
            period: 1s
        json_map_columns:
            data: properties
        max_in_flight: 64
        max_retries: 3
        region: us-east-1
        string_columns:
            myPk: ${!metadata("myPartionKey")}
            mySk: ${!metadata("mySortKey")}
        table: myTable
        is_delete: metadata("delete_this_record") == True # new field - (boolean, default: False) to indicate if the record should be deleted, Bloblang query like in switch check statement
        partition_key: myPk # new field - name of the partition key (string, no default)
        sort_key: mySk # new field - name of the sort key (string, no default)
```

Rohan-flutterint:
- Introduce `is_delete`, `partition_key` and `sort_key` in the aws_dynamodb output spec
- Parse `partition_key`/`sort_key` as column names (FieldString) and reuse the existing `string_columns` expressions to derive key values
- Implement a delete‐branch in `WriteBatch` that issues `DeleteItem` calls before any `BatchWriteItem` or `PutItem`
- Add `TestDynamoDBDeleteRequest` to verify delete‐only behaviour
- Add `TestDynamoDBMixedDeleteAndPut` to cover mixed delete+put batches

closes #408

### DIFF
--- a/internal/impl/aws/output_dynamodb.go
+++ b/internal/impl/aws/output_dynamodb.go
@@ -262,6 +262,9 @@ func newDynamoDBWriter(conf ddboConfig, mgr *service.Resources) (*dynamoDBWriter
 		log:   mgr.Logger(),
 		table: aws.String(conf.Table),
 	}
+	if len(conf.StringColumns) == 0 && len(conf.JSONMapColumns) == 0 {
+		return nil, errors.New("you must provide at least one column")
+	}
 	for k, v := range conf.JSONMapColumns {
 		if v == "." {
 			conf.JSONMapColumns[k] = ""


rohankumardubey:
**TO-DO**
Update documentation with new delete-item fields and example configuration

cthorner (commented):
Wow, that was fast! Does this allow for sort_key and partition_key values from json_map_columns? Valid values are strings, numbers or binary:

https://aws.amazon.com/blogs/database/effective-data-sorting-with-amazon-dynamodb/

jem-davies:
> **TO-DO** Update documentation with new delete-item fields and example configuration

Have you tried to update the docs with `make docs`?

jem-davies:
@rohankumardubey thanks for getting on this issue so quickly with a PR 😎  - I will be looking at getting this in the next release - so will just be addressing the stuff around the docs asap, then doing a review and getting it merged 🚀

internal/impl/aws/output_dynamodb.go
```diff
@@ -141,6 +156,18 @@ This output benefits from sending messages as a batch for improved performance.
...
+				Default("false").
+				Advanced(),
+			service.NewStringField(ddboFieldPartitionKey).
+				Description("The partition key for DeleteItem requests. Required when `is_delete` is true.").
+				Default("").
+				Advanced(),
+			service.NewStringField(ddboFieldSortKey).
+				Description("The sort key for DeleteItem requests.").
+				Default("").
+				Advanced(),
```
gregfurman: These fields are not strictly required. Should we not make them optional rather?

internal/impl/aws/output_dynamodb.go
```diff
@@ -141,6 +156,18 @@ This output benefits from sending messages as a batch for improved performance.
 				Description("The column key to place the TTL value within.").
 				Default("").
 				Advanced(),
+			service.NewInterpolatedStringField(ddboFieldIsDelete).
+				Description("Whether to perform a DeleteItem instead of PutItem.").
+				Default("false").
+				Advanced(),
```
gregfurman: I think we should have a `delete` object rather i.e

```
delete:
  condition: "${ ! <blobl>  }" # interpolated or static
  partition_key:  "${ ! <blobl>  }" # interpolated or static
  sort_key: "${ ! <blobl>  }" # interpolated or static, optional
```

then deletes can be disabled by omitting the entire object (instead of that "false").

internal/impl/aws/output_dynamodb.go
```diff
@@ -141,6 +156,18 @@ This output benefits from sending messages as a batch for improved performance.
 				Description("The column key to place the TTL value within.").
 				Default("").
 				Advanced(),
+			service.NewInterpolatedStringField(ddboFieldIsDelete).
```
gregfurman: Also, this should be a bloblang mapping instead of an interpolated string imo. Like, what if my value doesn't evaluate to `"true"` and I instead have a value of `"foobar"` or something.

We we should instead be passing in a string that we convert to a bloblang executor, then we do a QueryParts on each message and check if true i.e

https://github.com/warpstreamlabs/bento/blob/477a8266d0008f1fbe7a1b9518f2b6cc35c21951/internal/impl/pure/processor_while.go#L114-L117

then:


https://github.com/warpstreamlabs/bento/blob/477a8266d0008f1fbe7a1b9518f2b6cc35c21951/internal/impl/pure/processor_while.go#L132-L139

internal/impl/aws/output_dynamodb.go
```diff
@@ -141,6 +156,18 @@ This output benefits from sending messages as a batch for improved performance.
...
+				Description("Whether to perform a DeleteItem instead of PutItem.").
+				Default("false").
+				Advanced(),
+			service.NewStringField(ddboFieldPartitionKey).
+				Description("The partition key for DeleteItem requests. Required when `is_delete` is true.").
+				Default("").
+				Advanced(),
```
gregfurman: This should be required when the condition is set -- since we should assume that a condition being set _can_ evaluate to true.

internal/impl/aws/output_dynamodb.go
```diff
@@ -195,8 +222,9 @@ func newDynamoDBWriter(conf ddboConfig, mgr *service.Resources) (*dynamoDBWriter
 		log:   mgr.Logger(),
 		table: aws.String(conf.Table),
 	}
-	if len(conf.StringColumns) == 0 && len(conf.JSONMapColumns) == 0 {
-		return nil, errors.New("you must provide at least one column")
+
+	if len(conf.StringColumns) == 0 && len(conf.JSONMapColumns) == 0 && conf.IsDelete == nil {
+		return nil, errors.New("you must provide either columns to write or 'is_delete' field")
```
gregfurman: 🤔 if we want to enforce this, should we not rather use a `LintRule`?

internal/impl/aws/output_dynamodb.go
```diff
@@ -309,21 +318,69 @@ func (d *dynamoDBWriter) WriteBatch(ctx context.Context, b service.MessageBatch)
...
+		mapString := fmt.Sprint(d.conf.StringColumns)
+		fmt.Println("Conf string -> ", mapString)
+		isDel := false
+		if d.conf.IsDelete != nil {
+			var err error
+			isDelStr, err := b.TryInterpolatedString(i, d.conf.IsDelete)
+			if err != nil {
+				return fmt.Errorf("is_delete interpolation error: %w", err)
+			}
+			isDel = isDelStr == "true"
```
gregfurman: Already mentioned above but if we use the mapping approach then this can be simplified to no longer rely on a magic string "true".

internal/impl/aws/output_dynamodb.go
```diff
@@ -309,21 +318,69 @@ func (d *dynamoDBWriter) WriteBatch(ctx context.Context, b service.MessageBatch)
...
+		}
+
+		if isDel {
+			// build key using the names in PartitionKeyField/SortKeyField
+			key := map[string]types.AttributeValue{}
+
+			// pull the value expression from string_columns:
+			exprPK, ok := d.conf.StringColumns[d.conf.PartitionKeyField]
```
gregfurman: nit: can we extract this functionality out into seperate functions? This `WriteBatch` call is super long already lol.

internal/impl/aws/output_dynamodb.go
```diff
@@ -360,6 +418,20 @@ func (d *dynamoDBWriter) WriteBatch(ctx context.Context, b service.MessageBatch)
...
 
+	// Execute deletes first
+	for i, key := range deleteKeys {
+		if _, err := d.client.DeleteItem(ctx, &dynamodb.DeleteItemInput{
+			TableName: d.table,
+			Key:       key,
+		}); err != nil {
+			return fmt.Errorf("delete error at index %d: %w", i, err)
+		}
+	}
```
gregfurman: We shouldn't be doing individual deletes since it defeats the whole purpose of batching our requests. Instead, these `deleteKeys` should be used to create a slice of `types.WriteRequest` with the appropriate `types.DeleteRequest` set so that we batch these requests.

Note, that a PUT and DELETE on the same key will result in an error response:

> You try to perform multiple operations on the same item in the same BatchWriteItem request. For example, you cannot put and delete the same item in the same BatchWriteItem request.

internal/impl/aws/output_dynamodb_test.go
```diff
@@ -495,3 +503,92 @@ string_columns:
 
 	assert.Equal(t, expected, requests)
 }
+
+func TestDynamoDBDeleteRequest(t *testing.T) {
```
gregfurman: Can we add an integration test against LocalStack?

internal/impl/aws/output_dynamodb.go
```diff
@@ -244,48 +272,29 @@ func anyToAttributeValue(root any) types.AttributeValue {
 		for k, v2 := range v {
 			m[k] = anyToAttributeValue(v2)
 		}
-		return &types.AttributeValueMemberM{
-			Value: m,
-		}
+		return &types.AttributeValueMemberM{Value: m}
```
gregfurman: nit: Can we reformat these back to how they originally were? The (already large) diff is being polluted imo

internal/impl/aws/output_dynamodb.go
```diff
@@ -309,21 +318,69 @@ func (d *dynamoDBWriter) WriteBatch(ctx context.Context, b service.MessageBatch)
 	}()
 
 	writeReqs := []types.WriteRequest{}
+	deleteIndices := map[int]bool{}
+	deleteKeys := map[int]map[string]types.AttributeValue{}
+
 	if err := b.WalkWithBatchedErrors(func(i int, p *service.Message) error {
+
+		mapString := fmt.Sprint(d.conf.StringColumns)
+		fmt.Println("Conf string -> ", mapString)
```
gregfurman: Remove the print to stdout
```suggestion
```

jem-davies:
@rohankumardubey - I am currently looking at implementing changes to address @gregfurman's review comments - fyi

internal/impl/aws/output_dynamodb.go
jem-davies: Think it's ok to leave them to just have a default value?

internal/impl/aws/output_dynamodb.go
jem-davies: Agree that we should add a lintRule _as well as_ this - also I think that the final condition is not right because afaik, looking at the implementation you need to provide StringColumns if you are doing deletes anyway: 

```go
exprPK, ok := d.conf.StringColumns[d.conf.PartitionKeyDeleteField
```

Which also pertains to comment/question from @cthorner: 

https://github.com/warpstreamlabs/bento/pull/410#pullrequestreview-3052502813

jem-davies:
Left to do: 

- Add LintRules to ConfigSpec; 1 LintRule to check existence of either json_map_columns or string_columns, and 1 LintRule to check that if we have a delete.condition set, we need a delete.partition_key ✅
- Test on AWS 
- Rewrite documentation to make it clear how new feature works ✅ 
- Check the changes to individualRequestLoop ✅

@gregfurman, going to do ☝️ tomorrow - shouldn't need to make much code changes so you can take another look if you can 🙏

internal/impl/aws/output_dynamodb_test.go
jem-davies: I think we should do a follow up github issue ticket & PR for this.

internal/impl/aws/output_dynamodb.go
```diff
@@ -309,57 +356,49 @@ func (d *dynamoDBWriter) WriteBatch(ctx context.Context, b service.MessageBatch)
...
-		if d.ttl != 0 && d.conf.TTLKey != "" {
-			items[d.conf.TTLKey] = &types.AttributeValueMemberN{
-				Value: strconv.FormatInt(time.Now().Add(d.ttl).Unix(), 10),
-			}
-		}
-		for k, v := range d.conf.StringColumns {
-			s, err := b.TryInterpolatedString(i, v)
+		var err error
+		if d.conf.DeleteConditionExec != nil {
+			result, err := p.BloblangQuery(d.conf.DeleteConditionExec)
```
gregfurman: Wasn't aware of the `BloblangQuery` functionality. Neat!

If you use `p.BloblangQueryValue` we can just do a type cast to a bool instead of the bytes -> bool check i.e

```golang
result, err := p.BloblangQueryValue(d.conf.DeleteConditionExec)
if err != nil {
   return err
}
isDelete, ok := result.(bool)
if !ok {
  return fmt.Errorf("Bloblang expression should evaluate to a bool")
}
if isDelete {
   err = d.addDeleteRequest(i, &b, &writeReqs, p)
}

// ...
```

internal/impl/aws/output_dynamodb.go
```diff
@@ -441,3 +501,108 @@ unprocessedLoop:
 func (d *dynamoDBWriter) Close(context.Context) error {
 	return nil
 }
+
+//------------------------------------------------------------------------------
+
+func (d *dynamoDBWriter) addDeleteRequest(i int, b *service.MessageBatch, writeReqs *[]types.WriteRequest, p *service.Message) error {
```
gregfurman: Don't think we need to pass a pointer to the slice here since slices are technically already pointers to an underlying array in Go

internal/impl/aws/output_dynamodb.go
```diff
@@ -441,3 +501,108 @@ unprocessedLoop:
...
+
+	key := map[string]types.AttributeValue{}
+
+	exprPK, ok := d.conf.StringColumns[d.conf.PartitionKeyDeleteField]
+	if !ok {
+		// let's have a look in json_map_columns then;
```
gregfurman:
```suggestion
```

internal/impl/aws/output_dynamodb.go
```diff
@@ -309,57 +356,49 @@ func (d *dynamoDBWriter) WriteBatch(ctx context.Context, b service.MessageBatch)
...
-						}
-					} else {
-						items[k] = attr
-					}
-				} else {
-					d.log.Warnf("Unable to extract JSON map path '%v' from document: %v", v, err)
+
+			if isDel {
+				err = d.addDeleteRequest(i, &b, &writeReqs, p)
+				if err != nil {
```
gregfurman: If we aren't wrapping errors let's just return the result directly:

```suggestion
			if isDel {
				return d.addDeleteRequest(i, &b, &writeReqs, p)
			}
```

internal/impl/aws/output_dynamodb.go
```diff
@@ -309,57 +356,49 @@ func (d *dynamoDBWriter) WriteBatch(ctx context.Context, b service.MessageBatch)
...
-		items := map[string]types.AttributeValue{}
-		if d.ttl != 0 && d.conf.TTLKey != "" {
-			items[d.conf.TTLKey] = &types.AttributeValueMemberN{
-				Value: strconv.FormatInt(time.Now().Add(d.ttl).Unix(), 10),
-			}
-		}
-		for k, v := range d.conf.StringColumns {
-			s, err := b.TryInterpolatedString(i, v)
+		var err error
+		if d.conf.DeleteConditionExec != nil {
```
gregfurman: nit: can we do a check here if `d.conf.DeleteConditionExec == nil` and do the simple case? 

```suggestion
		if d.conf.DeleteConditionExec == nil {
			return d.addPutRequest(i, &b, &writeReqs, p)
		}
```

Will have to reformat some stuff but you get the point 🫠

internal/impl/aws/output_dynamodb.go
```diff
@@ -309,57 +356,49 @@ func (d *dynamoDBWriter) WriteBatch(ctx context.Context, b service.MessageBatch)
...
-				Item: items,
-			},
-		})
+
+		err = d.addPutRequest(i, &b, &writeReqs, p)
+		if err != nil {
+			return err
+		}
+
 		return nil
```
gregfurman:
```suggestion
		return d.addPutRequest(i, &b, &writeReqs, p)
```

internal/impl/aws/output_dynamodb.go
```diff
@@ -309,57 +356,49 @@ func (d *dynamoDBWriter) WriteBatch(ctx context.Context, b service.MessageBatch)
 	}()
 
 	writeReqs := []types.WriteRequest{}
+
 	if err := b.WalkWithBatchedErrors(func(i int, p *service.Message) error {
```
gregfurman, line 380: super nit: Can we extract this out the `if`? Think this is getting pretty weedy with all these nested statements.

i.e 

```golang
headlineErr := b.WalkWithBatchedErrors(func(i int, p *service.Message) error { ... })
```

internal/impl/aws/output_dynamodb.go
```diff
@@ -373,6 +412,27 @@ func (d *dynamoDBWriter) WriteBatch(ctx context.Context, b service.MessageBatch)
 		for err != nil {
 			batchErr := service.NewBatchError(b, headlineErr)
 			for i, req := range writeReqs {
+				if req.DeleteRequest != nil {
+					if _, iErr := d.client.DeleteItem(ctx, &dynamodb.DeleteItemInput{
+						TableName: d.table,
+						Key:       req.DeleteRequest.Key,
+					}); iErr != nil {
+						d.log.Errorf("Put error: %v\n", iErr)
```
gregfurman: This would be a delete error, right?

internal/impl/aws/output_dynamodb.go
```diff
@@ -373,6 +412,27 @@ func (d *dynamoDBWriter) WriteBatch(ctx context.Context, b service.MessageBatch)
 		for err != nil {
 			batchErr := service.NewBatchError(b, headlineErr)
 			for i, req := range writeReqs {
+				if req.DeleteRequest != nil {
```
gregfurman, line 425: We need to ensure that none of these `req` structs have both a `DeleteRequest` and `PutRequest` defined (which would cause the entire batch to fail AFAIK), otherwise we could end up doing duplicate work.

internal/impl/aws/output_dynamodb.go
```diff
@@ -309,57 +356,49 @@ func (d *dynamoDBWriter) WriteBatch(ctx context.Context, b service.MessageBatch)
...
+			return err
+		}
+
 		return nil
+
 	}); err != nil {
 		return err
 	}
 
+	if len(writeReqs) == 0 {
```
gregfurman: comment(non-actionable): I think we need to be a bit more catering of AWS errors and error-cases. Like, if a batch of requests exceeds 25 requests it'll fail. Then next we try run these individual requests which could end up triggering some rate-limiting, etc.

Perhaps not a now problem but going forward we should consider reworking this whole component since it is perhaps a bit _too_ naive.

From [AWS docs:](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_BatchWriteItem.html)
> 
> If one or more of the following is true, DynamoDB rejects the entire batch write operation:
> 
> - One or more tables specified in the BatchWriteItem request does not exist.
> 
> - Primary key attributes specified on an item in the request do not match those in the corresponding table's primary key schema.
> 
> - You try to perform multiple operations on the same item in the same BatchWriteItem request. For example, you cannot put and delete the same item in the same BatchWriteItem request.
> 
> - Your request contains at least two items with identical hash and range keys (which essentially is two put operations).
> 
> - There are more than 25 requests in the batch.
> 
> - Any individual item in a batch exceeds 400 KB.
> 
> - The total request size exceeds 16 MB.
> 
> - Any individual items with keys exceeding the key length limits. For a partition key, the limit is 2048 bytes and for a sort key, the limit is 1024 bytes.
>

internal/impl/aws/output_dynamodb.go
jem-davies: 😭

rohankumardubey:
@jem-davies Sure, Thanks for addressing the changes. Was in vacation mode : ) .

internal/impl/aws/output_dynamodb.go
jem-davies: yeah that is nicer 🙏

internal/impl/aws/output_dynamodb.go
jem-davies: No I think that it is necessary: https://go.dev/play/p/mDcBz0mZogM

I could have just returned the writeReqs from the two functions but then I think you would be copying the entire slice because go is copy by default when you pass arguments to functions. This way I will be modifying it in place

internal/impl/aws/output_dynamodb.go
jem-davies: done that 🙇‍♂️

internal/impl/aws/output_dynamodb.go
jem-davies: don't think it looks too bad - actually think that other parts are a lot worse...

internal/impl/aws/output_dynamodb.go
jem-davies: Is it necessary to check this? With logic of the closure in WalkWithBatchedErrors, we are just doing if DeleteCondition == true; then append a DeleteRequest; else append a PutRequest?

internal/impl/aws/output_dynamodb.go
jem-davies: Ok so I have changed it to: 

```go

	if err := b.WalkWithBatchedErrors(func(i int, p *service.Message) error {
		if d.conf.DeleteConditionExec == nil {
			return d.addPutRequest(i, &b, &writeReqs, p)
		}

		result, err := p.BloblangQueryValue(d.conf.DeleteConditionExec)
		if err != nil {
			return fmt.Errorf("delete condition exec error: %w", err)
		}
		isDelete, ok := result.(bool)
		if !ok {
			return errors.New("delete.condition bloblang mapping should evaluate to a bool")
		}

		if isDelete {
			return d.addDeleteRequest(i, &b, &writeReqs, p)
		}

		return d.addPutRequest(i, &b, &writeReqs, p)

	}); err != nil {
		return err
	}
```

jem-davies: Guess we could check for d.conf.DeleteConditionExec == nil on component initialisation then have a closureFunc field in dynamoDBWriter which we would alter to either a func that does the delete check or skips it... 

Although I feel just having an if statement in the existing closure is simpler personally ....

internal/impl/aws/output_dynamodb.go
jem-davies: yeah agree - lets raise an issue

jem-davies:
@gregfurman  - responded to your 2nd batch of feedback - please could you take a look 🙇‍♂️ .

I think if there was an already existing localstack int. test, we should be adding a test for this new feature, but because there isn't I feel it should be a separate issue & PR - which we can raise and I will assign and prioritise for next week 👍 - wdyt? 

~~I will manually run a smoke test on real AWS after we happy with all the code changes.~~ EDIT: ended up adding an int. test

internal/impl/aws/output_dynamodb.go
gregfurman: I see. You're doing an append which would require a pointer.

But since we also pass in the index i of the request, we can just set the request with a:

writeReqs[i] = types.WriteRequest{}

Which means we don't need the pointer, right?

jem-davies: yeah I think that would work - provided we use `make()` with the length set to the number of message parts.

jem-davies: is it better though? - feel I was just using pointers to keep the logic similar to what was here before...

jem-davies:
AWS actually provides a docker image for dynamodb: amazon/dynamodb-local:latest

internal/impl/aws/output_dynamodb.go
jem-davies: have changed this now so we do: 

```go

	writeReqs := make([]types.WriteRequest, len(b))
	
	and
	
		writeReqs[i] = types.WriteRequest{
		PutRequest: &types.PutRequest{
			Item: items,
		},
	}

File: internal/impl/aws/output_dynamodb.go (hunk 12 above)
Comment (jem-davies) [reply internal/impl/aws/output_dynamodb.go]: Struggled to get this to work properly - think I was missing `msgWithContext := p.WithContext(ctx)` but then I couldn't find an example in ./internal/impl/ using BloblangQueryValue...

File: internal/impl/aws/output_dynamodb.go [hunk 22]
```diff
@@ -287,6 +352,7 @@ func anyToAttributeValue(root any) types.AttributeValue {
 	return &types.AttributeValueMemberS{
 		Value: fmt.Sprintf("%v", root),
 	}
+
```
gregfurman:
```suggestion
```

internal/impl/aws/output_dynamodb.go
```diff
@@ -195,9 +262,7 @@ func newDynamoDBWriter(conf ddboConfig, mgr *service.Resources) (*dynamoDBWriter
 		log:   mgr.Logger(),
 		table: aws.String(conf.Table),
 	}
-	if len(conf.StringColumns) == 0 && len(conf.JSONMapColumns) == 0 {
-		return nil, errors.New("you must provide at least one column")
-	}
+
```
gregfurman:
```suggestion
```

internal/impl/aws/output_dynamoddb_integration_test.go
```diff
@@ -0,0 +1,236 @@
...
+		if !errors.As(err, &derr) {
+			return err
+		}
+	}
+
+	if dt != nil && dt.Table != nil && dt.Table.TableStatus == types.TableStatusActive {
+		return nil
+	}
+
+	t.Logf("Creating table: %v\n", tableName)
```
gregfurman: nit: all of these log statements are probaly unnecessary

internal/impl/aws/output_dynamoddb_integration_test.go
```diff
@@ -0,0 +1,236 @@
...
+	t.Cleanup(func() {
+		assert.NoError(t, pool.Purge(resource))
+	})
+
+	_ = resource.Expire(900)
+	require.NoError(t, pool.Retry(func() error {
+		return createMusicTable(context.Background(), t, resource.GetPort("8000/tcp"))
+	}))
+
+	fmt.Println(resource.GetPort("8000/tcp"))
```
gregfurman:
```suggestion
```

internal/impl/aws/output_dynamoddb_integration_test.go
gregfurman, line 1: typo:

```
internal/impl/aws/output_dynamodb_integration_test.go => internal/impl/aws/output_dynamodb_integration_test.go
```

internal/impl/aws/output_dynamoddb_integration_test.go
```diff
@@ -0,0 +1,236 @@
...
+
+	// use bento to populate the table
+	builder := service.NewStreamBuilder()
+
+	err = builder.SetYAML(fmt.Sprintf(`
+input:
+  csv:
+    paths: ["./resources/dynamodbMusicTestDataPopTable.csv"]
```
gregfurman: nit: consider using a stream input of some kind here instead of relying on a CSV alone

internal/impl/aws/output_dynamodb.go
```diff
@@ -141,11 +165,54 @@ This output benefits from sending messages as a batch for improved performance.
...
+				Optional().
+				Advanced(),
 			service.NewOutputMaxInFlightField(),
 			service.NewBatchPolicyField(ddboFieldBatching),
 		).
 		Fields(config.SessionFields()...).
-		Fields(retries.CommonRetryBackOffFields(3, "1s", "5s", "30s")...)
+		Fields(retries.CommonRetryBackOffFields(3, "1s", "5s", "30s")...).
+		LintRule(`root = match {
+this.`+ddboFieldStringColumns+`.length() == 0 && this.`+ddboFieldJSONMapColumns+`.length() == 0 => ["at least one of: `+ddboFieldStringColumns+` or `+ddboFieldJSONMapColumns+` must be specified"],
```
gregfurman, line 196: ☠️ A function would've probably been easier here instead of doing this inline but it works haha

gregfurman (approved):
LGTM. Ran tests locally and all looks good. Some small nits but feel free to merge in when ready.

internal/impl/aws/output_dynamoddb_integration_test.go
jem-davies: removed

internal/impl/aws/output_dynamoddb_integration_test.go
jem-davies: 💀

internal/impl/aws/output_dynamoddb_integration_test.go
jem-davies: agree but don't think it's the end of the world

internal/impl/aws/output_dynamoddb_integration_test.go
jem-davies: fixed 😄

internal/impl/aws/output_dynamodb.go
jem-davies: want to keep them in ... - it's ok to add a lintRule as well ... sometimes these checks are useful because it's possible to instantiate the component without using yaml and therefore catching with lintRule won't happen. (although in this case it doesn't seem to be an issue...)

internal/impl/aws/output_dynamodb.go
jem-davies: https://github.com/warpstreamlabs/bento/issues/450

PR #414, opened 2025-07-28
AsyncWriter - log errors when failing to write

brianshih1:
Right now, the `AsyncWriter` loop calls `w.latencyMeasuringWrite(` in 3 different places but only emits one metric when there's an error: `mError.Incr(1)`.

This makes it hard to pinpoint exactly what went wrong, especially when debugging pipelines that are causing duplicate messages. I added some error logs and a metric for better observability.

### DIFF
--- a/internal/component/output/async_writer.go
+++ b/internal/component/output/async_writer.go
@@ -89,13 +89,14 @@ func (w *AsyncWriter) latencyMeasuringWrite(ctx context.Context, msg message.Bat
 func (w *AsyncWriter) loop() {
 	// Metrics paths
 	var (
-		mSent       = w.stats.GetCounter("output_sent")
-		mBatchSent  = w.stats.GetCounter("output_batch_sent")
-		mError      = w.stats.GetCounter("output_error")
-		mLatency    = w.stats.GetTimer("output_latency_ns")
-		mConn       = w.stats.GetCounter("output_connection_up")
-		mFailedConn = w.stats.GetCounter("output_connection_failed")
-		mLostConn   = w.stats.GetCounter("output_connection_lost")
+		mSent           = w.stats.GetCounter("output_sent")
+		mBatchSent      = w.stats.GetCounter("output_batch_sent")
+		mError          = w.stats.GetCounter("output_error")
+		mLatency        = w.stats.GetTimer("output_latency_ns")
+		mConn           = w.stats.GetCounter("output_connection_up")
+		mFailedConn     = w.stats.GetCounter("output_connection_failed")
+		mLostConn       = w.stats.GetCounter("output_connection_lost")
+		mReconnectError = w.stats.GetCounter("output_reconnect_error")
 
 		traceName = "output_" + w.typeStr
 	)
@@ -165,6 +166,9 @@ func (w *AsyncWriter) loop() {
 		// connection, then we gracefully accept defeat.
 		if w.connection.Load().Connected {
 			if latency, err = w.latencyMeasuringWrite(closeLeisureCtx, msg); err != component.ErrNotConnected {
+				if err != nil && w.typeStr != "reject" {
+					w.log.Error("Failed to send message to %v when connection status is already connected: %v\n", w.typeStr, err)
+				}
 				return
 			} else if err != nil {
 				mError.Incr(1)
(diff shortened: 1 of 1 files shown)


internal/component/output/async_writer.go
```diff
@@ -179,10 +183,14 @@ func (w *AsyncWriter) loop() {
...
 			if latency, err = w.latencyMeasuringWrite(closeLeisureCtx, msg); err != component.ErrNotConnected {
+				if err != nil && w.typeStr != "reject" {
+					w.log.Error("Failed to send message to %v when reconnecting: %v\n", w.typeStr, err)
+				}
 				w.connection.Store(component.ConnectionActive(w.mgr))
 				mConn.Incr(1)
 				return
 			} else if err != nil {
+				mReconnectError.Incr(1)
 				mError.Incr(1)
```
jem-davies: Would it be somewhat confusing to increment 2 seperate metric counters for the same error?

jem-davies (commented):
Think that the additional logs are good but there are implications for the new metric imo.

internal/component/output/async_writer.go
```diff
@@ -165,6 +165,9 @@ func (w *AsyncWriter) loop() {
 		// connection, then we gracefully accept defeat.
 		if w.connection.Load().Connected {
 			if latency, err = w.latencyMeasuringWrite(closeLeisureCtx, msg); err != component.ErrNotConnected {
+				if err != nil && w.typeStr != "reject" {
+					w.log.Error("Failed to send message to %v when connection status is already connected: %v\n", w.typeStr, err)
```
jem-davies, line 169: Think we should downgrade the log level to DEBUG?

internal/component/output/async_writer.go
```diff
@@ -179,6 +182,9 @@ func (w *AsyncWriter) loop() {
 				return
 			}
 			if latency, err = w.latencyMeasuringWrite(closeLeisureCtx, msg); err != component.ErrNotConnected {
+				if err != nil && w.typeStr != "reject" {
+					w.log.Error("Failed to send message to %v when reconnecting: %v\n", w.typeStr, err)
```
jem-davies, line 186: Think we should downgrade the log level to DEBUG?

internal/component/output/async_writer.go
```diff
@@ -165,6 +165,9 @@ func (w *AsyncWriter) loop() {
 		// connection, then we gracefully accept defeat.
 		if w.connection.Load().Connected {
 			if latency, err = w.latencyMeasuringWrite(closeLeisureCtx, msg); err != component.ErrNotConnected {
+				if err != nil && w.typeStr != "reject" {
+					w.log.Error("Failed to send message to %v when connection status is already connected: %v\n", w.typeStr, err)
+				}
```
gregfurman, line 170: Does this exact error not get logged already here since the `err` is set by the results `connectLoop`:

https://github.com/warpstreamlabs/bento/blob/ef027e48bd763dff8007acc38856c0c1ad0bc3a3/internal/component/output/async_writer.go#L215-L222


Afterwhich we check again if the `latencyMeasuringWrite` has an error and log accordingly:

https://github.com/warpstreamlabs/bento/blob/ef027e48bd763dff8007acc38856c0c1ad0bc3a3/internal/component/output/async_writer.go#L229-L236

internal/component/output/async_writer.go
```diff
@@ -179,6 +182,9 @@ func (w *AsyncWriter) loop() {
 				return
 			}
 			if latency, err = w.latencyMeasuringWrite(closeLeisureCtx, msg); err != component.ErrNotConnected {
+				if err != nil && w.typeStr != "reject" {
```
gregfurman, line 185: Also, what if the error is `component.ErrTypeClosed`?

internal/component/output/async_writer.go
brianshih1: true. The problem is right now it's hard to tell if an error happened during the retry or just a normal write. So I just wanted some kind of log to help me figure that out.

internal/component/output/async_writer.go
brianshih1: I disagree. This error means that the sink likely returned an actual error (e.g. Kafka leader is down) and the error message would help the customer fix that error. It's the same reason this current error log is not a `Debug` log I think: https://github.com/warpstreamlabs/bento/blob/main/internal/component/output/async_writer.go#L227

internal/component/output/async_writer.go
brianshih1: Maybe I can wrap that error instead? To avoid two similar error logs

internal/component/output/async_writer.go
brianshih1: but yea it does seem like that error is returned and would eventually be logged...

PR #416, opened 2025-07-29
Fix multipart s3 uploads for aws_s3 output destination

Towerthousand:
Starting on v1.29 of the github.com/aws/aws-sdk-go-v2/config package, the default value for RequestChecksumCalculation is RequestChecksumCalculationWhenSupported. This breaks the behavior of the aws_s3 output during multipart uploads, since we're not adding a checksum to the request. We're setting it back to RequestChecksumCalculationWhenRequired.

See https://github.com/aws/aws-sdk-go-v2/discussions/2960#discussioncomment-12077210


Towerthousand:
Happy to add something to the integration test that triggers a multipart upload so that this gets validated during testing.

PR #418, opened 2025-08-01
add SetModularLogger to StreamBuilder

lublak:


public/service/stream_builder.go
```diff
@@ -168,6 +168,12 @@ func (s *StreamBuilder) SetLogger(l *slog.Logger) {
 	s.customLogger = log.NewBentoLogAdapter(l)
 }
 
+// SetModularLogger sets a customer logger via Bento's standard logging interface,
+// allowing you to replace the default Bento logger with your own.
+func (s *StreamBuilder) SetModularLogger(l log.Modular) {
```
gregfurman, line 173: What about `SetStreamLogger`? Seems unfortunate that we've already used `SetLogger(l *slog.Logger)` and enforce a `slog.Logger` type...

Also, if we remove the `Fatal()` method from the `log.Modular` interface then a slog type would implement the `log.Modular` interface -- but this would be a breaking change. Unsure if anyone is using `log.Fatal` (since we certainly aren't).

cc @jem-davies

lublak: @gregfurman i looked around but i also think log.Fatal is not used anywhere so: Is it a breaking change? If its an interface and "SetModularLogger" or "SetStreamLogger" not exists yet. And yes SetLogger would be the better name, but its already used. SetSLogLogger would be here better, but hey: This would be a breaking change and also confusing.

gregfurman: I've given this another look and it seems that in the public API, we don't actually allow for retrieved loggers to call Fatal.

See the Logger struct in public/service/logger.go

So I think it's safe to remove Fatal from the interface.

Otherwise, we should probably define a public interface that a logger can implement instead (without fatal) and leave log.Modular as is.

jem-davies:
Closed for #419

PR #419, opened 2025-08-02
feat(public/stream_builder): Allow custom loggers

gregfurman:
Currently, Bento only allows us to set a stream's logger to a `slog.Logger{}` type. This is restrictive since it means custom implementations of loggers are not allowed (unless they're hackily wrapped around slog...?)

This change allows for any logger implementing a simple log interface to be passed and used as a logger:

```golang
type Leveled interface {
        Error(format string, v ...any)
	Warn(format string, v ...any)
	Info(format string, v ...any)
	Debug(format string, v ...any)
}
```

Moreover, we also create interfaces for loggers with `Fatal` and `Trace` methods, and dynamically call them after a type check at runtime. Else are accounted for with some default behaviour.

TODO:

- [ ] Remove `SetLogger` and allow for https://github.com/warpstreamlabs/bento/pull/418 to rebase against this -- since they were @lublak's changes to begin with.

### DIFF
--- a/public/service/stream_builder.go
+++ b/public/service/stream_builder.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"log/slog"
 	"net/http"
 	"os"
 
@@ -173,9 +174,16 @@ var (
 	_ LeveledLogger = (*slog.Logger)(nil)
 	_ LeveledLogger = (log.Modular)(nil)
 )
-// SetLogger sets a custom logger via Bento's standard logging interface,
+
+// SetLogger sets a slog logger via Bento's standard logging interface,
 // allowing you to replace the default Bento logger with your own.
-func (s *StreamBuilder) SetLogger(l LeveledLogger) {
+func (s *StreamBuilder) SetLogger(l *slog.Logger) {
+	s.customLogger = log.NewBentoLogAdapter(l)
+}
+
+// SetLeveledLogger sets a custom logger via Bento's standard logging interface,
+// allowing you to replace the default Bento logger with your own.
+func (s *StreamBuilder) SetLeveledLogger(l LeveledLogger) {
 	s.customLogger = newAirGapLogger(l)
 }
 


lublak:
@gregfurman thanks for the better pull request :+1: lgtm
For it will help to pass my custom logger into bento :smile:

public/service/stream_builder.go
```diff
@@ -162,10 +161,18 @@ func (s *StreamBuilder) SetPrintLogger(l PrintLogger) {
...
-// SetLogger sets a customer logger via Go's standard logging interface,
+// LeveledLogger is an interface supported by most loggers.
+type LeveledLogger interface {
+	Error(format string, v ...any)
+	Warn(format string, v ...any)
+	Info(format string, v ...any)
+	Debug(format string, v ...any)
+}
+
+// SetLogger sets a customer logger via Bento's standard logging interface,
```
jem-davies:
```suggestion
// SetLogger sets a custom logger via Bento's standard logging interface,
```

was a mistake in the original - unless it should be 'customer'.

public/service/stream_builder.go
```diff
@@ -162,10 +161,18 @@ func (s *StreamBuilder) SetPrintLogger(l PrintLogger) {
...
 // allowing you to replace the default Bento logger with your own.
-func (s *StreamBuilder) SetLogger(l *slog.Logger) {
-	s.customLogger = log.NewBentoLogAdapter(l)
+func (s *StreamBuilder) SetLogger(l LeveledLogger) {
+	s.customLogger = newAirGapLogger(l)
```
jem-davies, line 187: I think we should keep the `SetLogger` as is - and add a new `SetLeveledLogger` for the new functionality. wdyt? 

Just seems safer to me - I get that in existing calls we should be fine because *slog.Logger will satisfy the interface but still.

public/service/stream_builder.go
```diff
@@ -162,10 +161,18 @@ func (s *StreamBuilder) SetPrintLogger(l PrintLogger) {
...
 
-// SetLogger sets a customer logger via Go's standard logging interface,
+// LeveledLogger is an interface supported by most loggers.
+type LeveledLogger interface {
+	Error(format string, v ...any)
+	Warn(format string, v ...any)
+	Info(format string, v ...any)
+	Debug(format string, v ...any)
+}
+
```
jem-davies, line 172:
```suggestion

var (
	_ LeveledLogger = (*slog.Logger)(nil)
	_ LeveledLogger = (log.Modular)(nil)
)
```

suggestion - can make sure the LeveledLogger will accept *slog.Logger & log.Modular like this

public/service/logger.go
```diff
@@ -2,10 +2,116 @@ package service
...
+	if a.l == nil {
+		return
+	}
+	a.l.Warn(format, v...)
+}
+func (a *airGapLogger) Info(format string, v ...any) {
+	if a.l == nil {
+		return
+	}
+	a.l.Info(format, v...)
```
jem-davies, line 65:
```suggestion
	a.l.Info(fmt.Sprintf(format, v...))
```

I gave it a spin with `builder.SetLogger(slog.Default())` on this branch I would get: 

```
2025/09/13 17:08:11 INFO Input type %v is now active !BADKEY=generate
2025/09/13 17:08:11 INFO Output type %v is now active !BADKEY=stdout
```

but using the old `log.NewBentoLogAdapter(l)` I would get: 

```
2025/09/13 17:18:22 INFO Input type generate is now active path=root.input label=""
2025/09/13 17:18:22 INFO Output type stdout is now active path=root.output label=""
``` 

Not sure if we need to add in fmt.Sprint to all the Warn/Info ... etc. or as another suggestion - add a new function rather than altering `func (s *StreamBuilder) SetLogger(l *slog.Logger)` in place.

gregfurman: Weird. OK will give a look at this

gregfurman: I think this is desired behaviour since one is setting a custom logger. Wdyt?

Issue #420, opened 2025-08-04
Add support for service.ParsedConfig.FieldUInt


lublak:
Currently we have
```go
conf.FieldInt();
```

I think it would be useful to have an uint

```go
conf.FieldUInt();
```

There are some configs which uses conf.FieldInt()
Should we than switch to conf.FieldUInt where it makes sense? Like byte_"size" values etc.?

gregfurman:
@lublak Hmm... would it not be easier to add a FieldBytes that can take a byte string? Unsure if specific functionality for a UInt type is ideal especially considering config input is mostly done via yaml etc.

lublak:
A byte string? Right now, I can't really imagine what I would want to do with a byte string 😅.
UInt is very usefull for this rule:

```json
{
   "type" : "integer",
   "minimum" : 0
}
```

For me, it just makes sense for values that must be at least 0. Currently, values below 0 are also undocumented in many places. I think a clear rule would be very useful here.

gregfurman:
Sorry let me rephrase:

Was thinking a string representation of possible byte sizes with units attached. Like a FieldByteSize() where one can pass a scalar like FieldByteSize('10m') for 10mb etc.

What about the ability to attach a rule that allows for easier constraints? Like FieldInt().Min(0).Max(10) instead of needing to rely on LintRules().

I'm not against UInt -- I just think it could be a bit unergonomic for users coming from other languages/paradigms where they never normally have to think about this type of thing 😄 

Wdyt?

jem-davies:
My 2 cents: 

@lublak: Is there a specific need for a .FieldUInt()? 

If the only rationale is "because many fields that are .FieldInt() are naturally non-negative (e.g. byte_size)" - then I would think that the argument to add in .FieldUInt() is a weak one, imho. 

If there is a specific need - then yeah let's add it in... 

It's just that we have been fine without it and therefore it would seem we are ok as-is.

lublak:
@gregfurman , @jem-davies 

i hope its okay to answer you both at the same time!

```
FieldByteSize('10m')
```

this would be very useful for fields with "byte_sizes".

I haven't thought about it that explicitly. But I think that's a good idea too. Especially to make it easier to read
12mb; 1kb; 100b; 2GB

To be honest, my idea was more generally a value explicitly above 0.

```go
FieldInt().Min(0).Max(10)
```

That would be totally fine with me, of course.
[...]

PR #421, opened 2025-08-04
parsed: make FieldUInt return uint and use IGetUInt (#420)

Rohan-flutterint:
parsed: make FieldUInt return uint and use IGetUInt (#420)


jem-davies:
Discussions whether we will want to add FieldUInt on the ticket; https://github.com/warpstreamlabs/bento/issues/420 - imo we don't need them

PR #422, opened 2025-08-06
adds support for binding metrics from multiple StreamBuilders into one shared metrics instance

ecordell:
This has been a source of frustration for us: when using StreamBuilder to create multiple streams, each stream creates its own metrics registry and HTTP server, which means exposing one port per stream if you want metrics from all of them. We wanted a way to have metrics from multiple StreamBuilder-built streams report into one prometheus registry.

This PR introduces a `SharedMetrics` abstraction that can be associated with any given `StreamBuilder`. Stream names can be used to automatically label metrics, similar to how Bento's streams mode works, making it easy to distinguish metrics from different streams in the shared registry.

## Basic Usage
```go
// Create shared metrics setup
sharedMetrics, err := service.NewSharedMetricsSetup(`
prometheus:
  use_histogram_timing: false
`, 9090)

// Start shared metrics server
go sharedMetrics.ListenAndServe()

// Create multiple streams
for i := 0; i < 5; i++ {
    builder := service.NewStreamBuilder()
    sharedMetrics.ConfigureStreamBuilder(builder)
    // ... configure stream normally
    stream, _ := builder.Build()
    go stream.Run(context.Background())
}
// All streams report to single Prometheus endpoint on port 9090
```

## Stream Labeling
Similar to Bento's streams mode, each stream can be given a name that appears in all metrics:

```go
// Create streams with distinct names
streamNames := []string{"orders-processor", "user-events", "notifications"}
for _, name := range streamNames {
    builder := service.NewStreamBuilder()
    builder.SetStreamName(name)  // Add stream labels automatically
    sharedMetrics.ConfigureStreamBuilder(builder)
    // ... configure normally
}
```

**Results in labeled metrics:**
```
input_received{stream="orders-processor",path="root.input"} 5
input_received{stream="user-events",path="root.input"} 12  
input_received{stream="notifications",path="root.input"} 3
```

### DIFF
--- a/public/service/shared_metrics.go
+++ b/public/service/shared_metrics.go
@@ -1,0 +1,201 @@
+package service
+
+import (
+	"context"
+	"fmt"
+	"os"
+
+	"github.com/warpstreamlabs/bento/internal/api"
+	"github.com/warpstreamlabs/bento/internal/component/metrics"
+	"github.com/warpstreamlabs/bento/internal/config"
+	"github.com/warpstreamlabs/bento/internal/docs"
+	"github.com/warpstreamlabs/bento/internal/log"
+	"github.com/warpstreamlabs/bento/internal/manager"
+)
+
+// SharedMetricsSetup provides a way to create shared metrics that can be used
+// across multiple StreamBuilders, allowing them to aggregate metrics to the
+// same destination and expose a single HTTP endpoint.
+//
+// This solves the common problem when building multiple streams programmatically
+// with StreamBuilder APIs: each stream creates its own metrics registry and
+// HTTP server, leading to port conflicts. With SharedMetricsSetup, all streams
+// can share the same metrics registry and expose metrics via a single endpoint.
+//
+// Example usage:
+//
+//	// Create shared metrics setup
+//	sharedMetrics, err := service.NewSharedMetricsSetup(`
+//	prometheus:
+//	  use_histogram_timing: false
+//	`, 9090)
+//	if err != nil {
+//	  log.Fatal(err)
+//	}
+//	defer sharedMetrics.Shutdown(context.Background())
+//
+//	// Start the shared metrics server
+//	go sharedMetrics.ListenAndServe()
+//
+//	// Create multiple streams that share the same metrics with distinct labels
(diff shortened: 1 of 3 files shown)


public/service/shared_metrics_test.go
```diff
@@ -0,0 +1,311 @@
...
+
+	stream, err := builder.Build()
+	require.NoError(t, err)
+
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*2)
+	defer cancel()
+
+	err = stream.Run(ctx)
+	// Context timeout is expected for generate input
+	require.NotEqual(t, err, context.DeadlineExceeded, "Stream should complete before timeout")
```
jem-davies:
```suggestion
	require.NoError(t, err)
```

Confused as the comment reads "Context timeout is expected for generate input" but we then check the error is _Not Equal_ to context.DeadlineExceeded? 

Or should we increase the generate's interval and check the error is a context.DeadlineExceedd for some reason?

jem-davies (commented):
LGTM - 🚀  Thanks! - just one small thing regarding an assertion in one of the tests.

public/service/shared_metrics_test.go
ecordell: Just an old comment from an earlier iteration - sorry about that!

public/service/shared_metrics.go
```diff
@@ -0,0 +1,201 @@
...
+	metrics     *metrics.Namespaced
+	api         *api.Type
+	httpAddress string
+}
+
+// NewSharedMetricsSetup creates a shared metrics setup from a YAML configuration.
+// This allows multiple StreamBuilders to aggregate their metrics to the same
+// destination (e.g., shared Prometheus registry) and expose them via a single
+// HTTP server.
+func NewSharedMetricsSetup(metricsYAML string, httpPort int) (*SharedMetricsSetup, error) {
```
ecordell, line 66: This is adding to the package's public API - are we okay with the naming?

jem-davies: Yes I think so - what else could we have instead?

Issue #426, opened 2025-08-11
Experimental / beta components should require user to 'opt in'.


jem-davies:
I think that the use of experimental/beta components should require the user to explicitly allow them via a cli flag like --allow-experimental / --allow-beta, and also optFunc for the streamBuilder. 

Currently - other than the warning on the documentation page - there is nothing that will treat 'experimental' / 'beta' components differently; which then effectively means that we avoid making breaking changes to experimental & beta components - meaning that they are not as malleable as they should be...

jem-davies:
@gregfurman we have closed 2 PRs that have attempted to fix this issue - although I still think that adding a warning that an experimental / beta component is being used has pragmatic utility. 

I suppose it's this thread where the disagreement is: https://github.com/warpstreamlabs/bento/pull/507#discussion_r2615025167

PR #428, opened 2025-08-11
Added --allow-experimental and --allow-beta CLI flags #426

rohankumardubey:
- Added --allow-experimental and --allow-beta CLI flags and wired them into service startup so unstable components are only loaded when explicitly requested
- Introduced gating in the component environment to block beta/experimental plugins unless they are enabled, with opt-ins available via AllowExperimental/AllowBeta and stream-builder options for programmatic control
- Updated manager creation to respect the environment gating by constructing metrics and tracers through the global environment
- Closes : https://github.com/warpstreamlabs/bento/issues/426

### DIFF
--- a/internal/bundle/buffers.go
+++ b/internal/bundle/buffers.go
@@ -24,7 +24,16 @@ func (e *Environment) BufferAdd(constructor BufferConstructor, spec docs.Compone
 
 // BufferInit attempts to initialise a buffer from a config.
 func (e *Environment) BufferInit(conf buffer.Config, mgr NewManagement) (buffer.Streamed, error) {
-	return e.buffers.Init(conf, mgr)
+	spec, exists := e.buffers.specs[conf.Type]
+	if !exists {
+		return nil, component.ErrInvalidType("buffer", conf.Type)
+	}
+	if err := e.allowStatus(spec.spec); err != nil {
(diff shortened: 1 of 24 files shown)


joeblew999:
Great idea !!

jem-davies:
@rohankumardubey - thanks for the effort with such a quick PR again 🚀  

I had raised this as an issue because it's bad that the only thing to warn users regarding beta/experimental components are the warnings on the doc site - and that it would be cool if we had something like [feature-gates](https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/) from kubernetes. 

The problem with fixing this, is that we will inevitably break the public API, and therefore to maintain semantic versioning, we would need to release it on a Major version update. Also prior to that we will need to review the state of experimental & beta components. 

Maybe we could alter such that we log warnings for now, on a minor release, and mark this for a major release?

jem-davies:
Going to pick this up and alter such that we emit warnings for now

jem-davies:
OK so with https://github.com/warpstreamlabs/bento/pull/507 - we can flag the warnings (and in future potential terminal errors) at the time we Lint the configuration - meaning that we won't instantiate any components before shutting down. i.e. Catching the experimental / beta component earlier than the approach used here.

Issue #423, opened 2025-08-07
Add a "cut" bloblang method for strings & arrays


jem-davies:
It would be a good addition if we had a bloblang function like split but for arrays so that you could have: 

```json
{"array": ["Lex Luthor", "Superman", "", "Lois Lane"]}
```

```
root = this.array.split("")
```

giving: 

```json
["Lex Luthor", "Superman"], ["Lois Lane"]
```

But also we could accept a predicate to determine if an element should be a snipping point.

iamramtin:
Hey! Can I be assigned this issue?

jem-davies:
@iamramtin yeah sure 😄  - it came out of a discord conversation: https://discord.com/channels/1130882542451556482/1247203340689997924/1403055584047272017 if you want context - albeit small... 

And I think @gregfurman was suggesting adding that stuff about a predicate 

Maybe add the option to keep the snipping points as elements somehow ...

gregfurman:
Am thinking this would be operable on any iterable type i.e

For strings:
```coffee
root = this.cut("kafka")

# In:  "my favourite streaming protcol is kafka but don't tell the others..."
# Out: ["my favourite streaming protcol is kafka", "but don't tell the others..." ]
```

For arrays/other iterables:
```coffee
root = this.cut(x -> x == "kafka")

# In:  ["kafka_franz", "kafka", "redis_streams", "gcp_bigquery_select"]
# Out: [["kafka_franz", "kafka"], ["redis_streams", "gcp_bigquery_select"]]
```

You can look at `find_by` for some examples on how this is implemented in code:

```coffee
root.index = this.find_by(v -> v != "bar")
```

But instead the `.cut` should short-circuit on find instead of iterating though each element once a match is found.

iamramtin:
Looking at the examples, I notice these are actually **two different functionalities**:

## Split behaviour:
```
["Lex Luthor", "Superman", "", "Lois Lane"]
root = this.array.split("")
# Out: [["Lex Luthor", "Superman"], ["Lois Lane"]]  // excludes delimiter
```

## Cut behaviour:
```
["kafka_franz", "kafka", "redis_streams", "gcp_bigquery_select"]
root = this.cut(x -> x == "kafka")
# Out: [["kafka_franz", "kafka"], ["redis_streams", "gcp_bigquery_select"]]  // includes delimiter
```

The key difference:
- **`split`**: Excludes the delimiter from results
- **`cut`**: Includes the delimiter in the first segment

[...]

PR #430, opened 2025-08-12
feat: Enhance Bloblang split functionality

iamramtin:
## Enhances the Bloblang split functionality:

1. Extends `split` method to work on arrays in addition to strings
2. Adds `split_by` method enables predicate-based splitting using query expressions
---

#### Static splitting
```
root.words = "hello,world".split(",")
# Out: ["hello", "world"]

root.segments = ["a", "b", "SPLIT", "c", "d"].split("SPLIT")
# Out: [["a","b"], ["c","d"]]
```

#### Predicate splitting
```
root.words = "hello world".split_by(c -> c == " ")
# Out: ["hello", "world"]

root.large_nums = [1, 2, 100, 3, 4, 200, 5].split_by(x -> x > 50)
# Out: [[1,2], [3,4], [5]]

root.authors =  ["George Orwell", "Franz Kafka", "Anton Chekhov"].split_by(x -> x.contains("Kafka")) → [["George Orwell"], ["Anton Chekhov"]]
# Out:  [["George Orwell"], ["Anton Chekhov"]]
```

References issue [#423](https://github.com/warpstreamlabs/bento/issues/423)


internal/bloblang/query/methods_structured.go
```diff
@@ -1733,3 +1733,204 @@ func mapWithout(m map[string]any, paths [][]string) map[string]any {
...
+		"Splits a string or array into segments by splitting on all occurrences of a delimiter value. Returns an array of segments with delimiters excluded.",
+	).InCategory(
+		MethodCategoryObjectAndArray,
+		"",
+		NewExampleSpec("Split string on delimiter",
+			`root = this.split(",")`,
+			`"foo,bar,baz"`,
+			`["foo","bar","baz"]`,
+		),
+		NewExampleSpec("Split array on value",
```
jem-davies:
```suggestion
		NewExampleSpec("Split array of strings by splitting it on a string separator.",
```

Get rid of the old ExampleSpec and move it to another .InCategory 👇

internal/bloblang/query/methods_structured.go
```diff
@@ -1733,3 +1733,204 @@ func mapWithout(m map[string]any, paths [][]string) map[string]any {
...
+			`root = this.villains.split("Kraven The Hunter")`,
+			`{"villains": ["Doctor Octopus", "Electro", "Kraven The Hunter", "Mysterio", "Sandman", "Vulture"]}`,
+			`[["Doctor Octopus","Electro"],["Mysterio","Sandman","Vulture"]]`,
+		),
+	).Param(ParamString("delimiter", "The delimiter to split on.")),
```
jem-davies:
```suggestion
		NewExampleSpec("Split an array of mixed types.",
			`root = this.mixed.split(0)`,
			`{"mixed": [1, "a", 2, "b", 3, "c", 4, "d", 0, "e", 5, "f", 6, "g", 7, "h", 0, "i", 8, "j", 9]}`,
			`[[1,"a",2,"b",3,"c",4,"d"],["e",5,"f",6,"g",7,"h"],["i",8,"j",9]]`,
		),
		NewExampleSpec("Split array of objects by splitting it on a object seperator.",
			`root = this.objects.split({"i": 0, "v": "vs"})`,
			`{"objects": [{"i": 1, "v": "Superman"}, {"i": 0, "v": "vs"}, {"i":2, "v": "Batman"}]}`,
			`[[{"i":1,"v":"Superman"}],[{"i":2,"v":"Batman"}]]`,
		),
	).InCategory(
		MethodCategoryStrings,
		"Split a string value into an array of strings by splitting it on a string separator.",
		NewExampleSpec("",
			`root.new_value = this.value.split(",")`,
			`{"value":"foo,bar,baz"}`,
			`{"new_value":["foo","bar","baz"]}`,
		),
	).Param(ParamAny("delimiter", "The delimiter to split with.")),
```

Add back the .InCategory(MethodCategoryStrings, so it shows in the strings section of the docs & add a couple more examples with different types in the CategoryObjectAndArray. 

Also alter the delimiter to be a ParamAny so that we can accept numbers, objects & strings as delim's

internal/bloblang/query/methods_structured.go
```diff
@@ -1733,3 +1733,204 @@ func mapWithout(m map[string]any, paths [][]string) map[string]any {
...
+	func(args *ParsedParams) (simpleMethod, error) {
+		delim, err := args.FieldString("delimiter")
```
jem-davies:
```suggestion
		delim, err := args.Field("delimiter")
```

internal/bloblang/query/methods_structured.go
```diff
@@ -1733,3 +1733,204 @@ func mapWithout(m map[string]any, paths [][]string) map[string]any {
...
+		if err != nil {
+			return nil, err
+		}
+
+		return func(v any, ctx FunctionContext) (any, error) {
+			switch t := v.(type) {
+			case string:
+				parts := strings.Split(t, delim)
```
jem-davies:
```suggestion
			case string:
				delimStr, ok := delim.(string)
				if !ok {
					return nil, fmt.Errorf("delimiter type: %T must match type of list elements: %T", delim, v)
				}
				parts := strings.Split(t, delimStr)
				parts := strings.Split(t, delim)
```

After changing to ParamAny we need this type assertion.

internal/bloblang/query/methods_structured.go
```diff
@@ -1733,3 +1733,204 @@ func mapWithout(m map[string]any, paths [][]string) map[string]any {
...
+				}
+				return vals, nil
+
+			case []byte:
+				parts := bytes.Split(t, []byte(delim))
+				vals := make([]any, 0, len(parts))
+				for _, b := range parts {
+					vals = append(vals, b)
+				}
+				return vals, nil
```
jem-davies:
```suggestion
```

I think that this is a redundant case because it shouldn't be possible to have a []byte here because our source is ultimately a json blob.

internal/bloblang/query/methods_structured_test.go
```diff
@@ -149,3 +150,382 @@ func TestMethodImmutability(t *testing.T) {
...
+			input:    []byte("foo,bar,baz"),
+			param:    ",",
+			expected: []any{[]byte("foo"), []byte("bar"), []byte("baz")},
+		},
+		{
+			name:     "bytes split empty input",
+			input:    []byte(""),
+			param:    ",",
+			expected: []any{[]byte("")},
+		},
```
jem-davies, line 289:
```suggestion
```

As per the review - don't think we need to worry about []byte 🤔

internal/bloblang/query/methods_structured.go
jem-davies: Ok actually I see the original impl. does have it so probably better left in ...

internal/bloblang/query/methods_structured_test.go
jem-davies: Keep 'em in - because the original impl does deal with []byte - hmm maybe it's for the `content()` func 🤔

internal/bloblang/query/methods_structured.go
```diff
@@ -1733,3 +1733,204 @@ func mapWithout(m map[string]any, paths [][]string) map[string]any {
 	}
 	return newMap
 }
+
+//------------------------------------------------------------------------------
+
+var _ = registerSimpleMethod(
+	NewMethodSpec(
```
jem-davies, line 1740: Could we add a .AtVersion("1.11.0") to this as well?

internal/bloblang/query/methods_structured.go
```diff
@@ -1733,3 +1733,204 @@ func mapWithout(m map[string]any, paths [][]string) map[string]any {
...
+			`root = this.split_by(x -> x.contains("Kafka"))`,
+			`["George Orwell", "Franz Kafka", "Anton Chekhov"]`,
+			`[["George Orwell"],["Anton Chekhov"]]`,
+		),
+		NewExampleSpec("Split array using numeric predicate",
+			`root = this.split_by(x -> x > 5)`,
+			`[1, 2, 10, 3, 4, 20, 7]`,
+			`[[1,2],[3,4],[],[]]`,
+		),
+	).Param(ParamQuery("predicate", "A query that returns true where splits should occur.", false)),
```
jem-davies:
```suggestion
	).Param(ParamQuery("predicate", "A query that returns true where splits should occur.", false)).Experimental().AtVersion("1.11.0"),
```

Add the experimental tag for this one - and the AtVersion again?

internal/bloblang/query/methods_structured_test.go
```diff
@@ -149,3 +150,382 @@ func TestMethodImmutability(t *testing.T) {
...
+			name:  "string split by space characters",
+			input: "hello world test",
+			predicate: arithmetic(
+				NewFieldFunction(""),
+				literalFn(" "),
+				ArithmeticEq,
+			),
+			expected: []any{"hello", "world", "test"},
+		},
+		{
```
jem-davies, line 426:
```suggestion
		},
		{
			name:  "string with emoji split by emoji char",
			input: "❤️ 🧡 💛 💚 💙 💜",
			predicate: arithmetic(
				NewFieldFunction(""),
				literalFn("💛"),
				ArithmeticEq,
			),
			expected: []any{"❤️ 🧡 ", " 💚 💙 💜"},
		},
		{
			name:  "chinese split by space characters",
			input: "你好 世界",
			predicate: arithmetic(
				NewFieldFunction(""),
				literalFn(" "),
				ArithmeticEq,
			),
			expected: []any{"你好", "世界"},
		},
		{
```

The "string with emoji split by emoji" will catch issues with UTF-8 & Runes but it passes as-is 😎

jem-davies (changes requested):
Just a couple of small things for "split_by" - thanks so much for taking these on !

internal/bloblang/query/methods_structured.go
iamramtin: Thinking this may be too verbose

jem-davies: Oh I think that I suggested this so the method still appears in the string section of the method docs.

jem-davies: as it was before I think

iamramtin:
@jem-davies Thanks for your feedback, I have made the requested changes. Let me know if we can :shipit:

jem-davies:
@iamramtin Have taken a look at the changes - it seems that for some of the feedback there has been no response or a change to address it. 

The feedback I think is still outstanding: 

- This PR should alter the existing .split method's docs by adding another .InCategory func to the method spec so that the [existing section in the string section of the methods page remains](https://warpstreamlabs.github.io/bento/docs/guides/bloblang/methods#split) (the delimiter parameter will change type here  but think it's ok...) You can have as many number of .InCategory so I think this is a good case for adding one rather than changing the existing. 
- As my suggestions have altered the delimiter parameter to allow any type, not just strings I think we should add some test cases that have different types for the delimiter parameter. 


Happy to add commits to the PR if you want me to - just let me know

iamramtin:
> This PR should alter the existing .split method's docs by adding another .InCategory func to the method spec 

Hey @jem-davies, I misunderstood what you were asking about adding another `.InCategory` func to `split` method. I have updated it for `split` and `split_by` with appropriate examples both for the string categories as well as objects and arrays.

> As my suggestions have altered the delimiter parameter to allow any type, not just strings I think we should add some test cases that have different types for the delimiter parameter.

I've added more tests, let me know if you're happy, feel free to add more if there are cases I've missed

jem-davies:
@iamramtin great thanks 😄

Issue #82, opened 2024-07-19
StreamBuilder causing data races


nickzelei:
We are using Bento as a part of our Go process and we spawn multiple bento syncs at once.
We instantiate a new bento environment and stream builder for each one.

This seems to be causing data races when running go tests with `-race` flag.

Here is a really simple repo where this can be easily reproduced: https://github.com/nickzelei/benthos-stream-race

Here is the full stack trace:

```
WARNING: DATA RACE
Read at 0x00c0000f7808 by goroutine 17:
  github.com/warpstreamlabs/bento/internal/docs.FieldSpecs.SetDefault()
      /Users/nick/go/pkg/mod/github.com/warpstreamlabs/bento@v1.1.0/internal/docs/field.go:655 +0xb8
  github.com/warpstreamlabs/bento/internal/docs.FieldSpecs.SetDefault()
      /Users/nick/go/pkg/mod/github.com/warpstreamlabs/bento@v1.1.0/internal/docs/field.go:660 +0x15c
  github.com/warpstreamlabs/bento/public/service.NewStreamBuilder()
      /Users/nick/go/pkg/mod/github.com/warpstreamlabs/bento@v1.1.0/public/service/stream_builder.go:80 +0x164
  github.com/warpstreamlabs/bento/public/service.(*Environment).NewStreamBuilder()
      /Users/nick/go/pkg/mod/github.com/warpstreamlabs/bento@v1.1.0/public/service/environment.go:106 +0x4c
  main.sync()
      /Users/nick/code/playground/benthos-stream-race/main.go:51 +0x48
  main.main.func1()
      /Users/nick/code/playground/benthos-stream-race/main.go:38 +0x30
  golang.org/x/sync/errgroup.(*Group).Go.func1()
      /Users/nick/go/pkg/mod/golang.org/x/sync@v0.6.0/errgroup/errgroup.go:78 +0x7c

Previous write at 0x00c0000f7808 by goroutine 18:
  github.com/warpstreamlabs/bento/internal/docs.FieldSpecs.SetDefault()
      /Users/nick/go/pkg/mod/github.com/warpstreamlabs/bento@v1.1.0/internal/docs/field.go:662 +0x214
  github.com/warpstreamlabs/bento/internal/docs.FieldSpecs.SetDefault()
      /Users/nick/go/pkg/mod/github.com/warpstreamlabs/bento@v1.1.0/internal/docs/field.go:660 +0x15c
[...]

PR #432, opened 2025-08-12
fix data race in config/schema.go

miparnisari:
Closes https://github.com/warpstreamlabs/bento/issues/82

Go's data race flags `service.NewStreamBuilder()` as having a data race.

In this PR I'm
- fixing that race
- change the test step in the github workflow to always run `make test-race`
- added a `make help` target that displays some of the targets available


Makefile
```diff
@@ -85,17 +88,17 @@ fmt:
...
-test: $(APPS)
+test: $(APPS) ## Run tests
 	@go test $(GO_FLAGS) -ldflags "$(LD_FLAGS)" -timeout 3m ./...
 	@$(PATHINSTBIN)/bento template lint $(TEMPLATE_FILES)
 	@$(PATHINSTBIN)/bento test ./config/test/...
 
-test-race: $(APPS)
-	@go test $(GO_FLAGS) -ldflags "$(LD_FLAGS)" -timeout 3m -race ./...
+test-race: $(APPS) ## Run tests with -race
+	@go test $(GO_FLAGS) -ldflags "$(LD_FLAGS)" -timeout 3m -v -race ./...
```
jem-davies:
```suggestion
	@go test $(GO_FLAGS) -ldflags "$(LD_FLAGS)" -timeout 3m -race ./...
```

Do we need to alter the verbosity? Won't we get a lot of redundant info when we run via github actions?

miparnisari: I guess we don't need to :)

jem-davies:
Think it's good to alter the test to use the data race detector - but seems to be causing an issue with one of the tests atm - will have to see if we can sort that test out before we merge this ... 

Or we alter this PR to remove the data race detector - and raise a separate issue for it ...

miparnisari:
I don't think the failure seen is related to the race detector. `TestWorkflowUnwrapResourceBranches` has

```go
tCtx, done := context.WithTimeout(context.Background(), time.Second*30)
defer done()

go func() {
	assert.NoError(t, strm.Run(tCtx))
}()
```

it's very likely that that `assert` is misplaced because `Run` will return `context deadline exceeded`. In general, one shouldn't put assertions inside goroutines. I know that there is a linter specifically for this but i can't recall the name.

jem-davies:
> I don't think the failure seen is related to the race detector. `TestWorkflowUnwrapResourceBranches` has
> 
> ```go
> tCtx, done := context.WithTimeout(context.Background(), time.Second*30)
> defer done()
> 
> go func() {
> 	assert.NoError(t, strm.Run(tCtx))
> }()
> ```
> 
> it's very likely that that `assert` is misplaced because `Run` will return `context deadline exceeded`. In general, one shouldn't put assertions inside goroutines. I know that there is a linter specifically for this but i can't recall the name.

But doesn't using the race detector introduce [runtime overheads](https://go.dev/doc/articles/race_detector#Runtime_Overheads)? Couldn't that affect the result of tests?

miparnisari:
The race detector will make the tests run slower, yes, but it shouldn't break a test _if the underlying code is thread safe_.

jem-davies:
@miparnisari sorry I just meant that altering the github action to use the race detector, when running the tests, has caused some to fail - not because it has detected an explicit data race but because the slower execution time has exposed a test that relies on completing within a context with a timeout - is that right 🤔 -

I have made [this commit](https://github.com/warpstreamlabs/bento/pull/441/commits/9810b310da169418b61a77707715e1591ac08dca) on another branch that should work: https://github.com/warpstreamlabs/bento/actions/runs/17055375741/job/48351860815?pr=441

jem-davies:
https://github.com/warpstreamlabs/bento/issues/445

Issue #433, opened 2025-08-14
Create a gRPC plugin in go


akoserwal:
Hi Team,

I have created a plugin https://github.com/akoserwal/bento-grpc-plugin

I want feedback, and also if there is a place where I can add a reference for this plugin?

jem-davies:
Hey that sounds great! - Have you thought about opening a PR to include in Bento?

akoserwal:
Is there a guide that I can follow to submit the PR?

Which folder location in Bento repo?

jem-davies:
[This PR ](https://github.com/warpstreamlabs/bento/pull/118)added a new folder of 'cypher' components - could you use that for reference?

jem-davies:
Still working on adding gRPC components: https://github.com/warpstreamlabs/bento/pull/764

PR #434, opened 2025-08-14
add gRPC support

akoserwal:
closes https://github.com/warpstreamlabs/bento/issues/433

Add gRPC support

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,8 @@
-.PHONY: all serverless deps docker docker-cgo clean docs test test-race test-integration fmt lint install deploy-docs playground
+.PHONY: all serverless deps docker docker-cgo clean docs test test-race test-integration fmt lint install deploy-docs playground proto install-protoc-plugins
 
 TAGS ?=
-
+GO ?= go
+GOPATH_BIN := $(shell $(GO) env GOPATH)/bin
 GOMAXPROCS         ?= 1
 INSTALL_DIR        ?= $(GOPATH)/bin
 WEBSITE_DIR        ?= ./website
@@ -12,7 +13,7 @@ PATHINSTSERVERLESS = $(DEST_DIR)/serverless
 PATHINSTDOCKER     = $(DEST_DIR)/docker
 DOCKER_IMAGE       ?= ghcr.io/warpstreamlabs/bento
 
-VERSION   := $(shell git describe --tags || echo "v0.0.0")
+VERSION   := $(shell git describe --tags 2>/dev/null || echo "v0.0.0")
 VER_CUT   := $(shell echo $(VERSION) | cut -c2-)
 VER_MAJOR := $(shell echo $(VER_CUT) | cut -f1 -d.)
 VER_MINOR := $(shell echo $(VER_CUT) | cut -f2 -d.)
@@ -25,6 +26,11 @@ VER_FLAGS = -X main.Version=$(VERSION) -X main.DateBuilt=$(DATE)
 LD_FLAGS   ?= -w -s
 GO_FLAGS   ?=
 DOCS_FLAGS ?=
+
+# Protobuf generation
+PROTOC           ?= protoc
+PROTO_SRC_DIR    := internal/impl/grpc/proto
+PROTO_GO_OUT_DIR := internal/impl/grpc/pb/proto
 
 APPS = bento
 all: $(APPS)
@@ -115,6 +121,23 @@ docs: $(APPS) $(TOOLS)
 		"$(WEBSITE_DIR)/cookbooks/**/*.md" \
 		"$(WEBSITE_DIR)/docs/**/*.md"
 	@$(PATHINSTBIN)/bento template lint "./config/template_examples/*.yaml"
+
+# -----------------------------------------------------------------------------
+# Protobufs for gRPC plugin
+# -----------------------------------------------------------------------------
+install-protoc-plugins:
+	@echo "Installing protoc plugins..."
+	@GOBIN=$(GOPATH_BIN) $(GO) install google.golang.org/protobuf/cmd/protoc-gen-go@latest
+	@GOBIN=$(GOPATH_BIN) $(GO) install google.golang.org/grpc/cmd/protoc-gen-go-grpc@latest
+
+proto: install-protoc-plugins
+	@mkdir -p $(PROTO_GO_OUT_DIR)
+	@PATH="$(GOPATH_BIN):$$PATH" $(PROTOC) \
+		-I $(PROTO_SRC_DIR) \
+		--go_out=$(PROTO_GO_OUT_DIR) --go_opt=paths=source_relative \
+		--go-grpc_out=$(PROTO_GO_OUT_DIR) --go-grpc_opt=paths=source_relative \
+		$(PROTO_SRC_DIR)/ingest.proto
+
 
 # HACK:(gregfurman): Change misc/wasm/wasm_exec.js => lib/wasm/wasm_exec.js when using Go 1.24
 playground:
--- a/internal/impl/grpc/pb/proto/ingest.pb.go
+++ b/internal/impl/grpc/pb/proto/ingest.pb.go
@@ -2,7 +2,7 @@
 // versions:
 // 	protoc-gen-go v1.36.7
 // 	protoc        v3.19.4
-// source: proto/ingest.proto
+// source: ingest.proto
 
 package pb
 
@@ -31,7 +31,7 @@ type Frame struct {
 
 func (x *Frame) Reset() {
 	*x = Frame{}
-	mi := &file_proto_ingest_proto_msgTypes[0]
+	mi := &file_ingest_proto_msgTypes[0]
 	ms := protoimpl.X.MessageStateOf(protoimpl.Pointer(x))
 	ms.StoreMessageInfo(mi)
 }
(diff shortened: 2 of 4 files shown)


internal/impl/grpc/register.go
```diff
@@ -0,0 +1,169 @@
...
+	}
+	if err := registerGRPCOutput(); err != nil {
+		return err
+	}
+	return nil
+}
+
+func registerGRPCInput() error {
+	spec := service.NewConfigSpec().
+		Version("1.6.0").
```
jem-davies, line 19:
```suggestion
		Version("1.11.0").
```

internal/impl/grpc/register.go
```diff
@@ -0,0 +1,169 @@
...
+				headers:     headers,
+			},
+		}
+		return newGRPCInput(address, consumerID, tlsConf, opts, res)
+	})
+}
+
+func registerGRPCOutput() error {
+	spec := service.NewConfigSpec().
+		Version("1.6.0").
```
jem-davies, line 94:
```suggestion
		Version("1.11.0").
```

Makefile
```diff
@@ -116,6 +122,23 @@ docs: $(APPS) $(TOOLS)
 		"$(WEBSITE_DIR)/docs/**/*.md"
 	@$(PATHINSTBIN)/bento template lint "./config/template_examples/*.yaml"
 
+# -----------------------------------------------------------------------------
+# Protobufs for gRPC plugin
+# -----------------------------------------------------------------------------
+install-protoc-plugins:
+	@echo "Installing protoc plugins..."
+	@GOBIN=$(GOPATH_BIN) $(GO) install google.golang.org/protobuf/cmd/protoc-gen-go@latest
+	@GOBIN=$(GOPATH_BIN) $(GO) install google.golang.org/grpc/cmd/protoc-gen-go-grpc@latest
```
jem-davies, line 131: Think we should pin these deps rather than using the latest tag

akoserwal:
One issue with the plugin is that it requires the gRPC server to implement `pb.UnimplementedBentoIngressServer` to publish and subscribe, which is not feasible in most cases, as it necessitates additional code changes in the application server code. I will replace the generic implementation, which will work without any code changes to the server or receiver client.

jem-davies:
> One issue with the plugin is that it requires the gRPC server to implement `pb.UnimplementedBentoIngressServer` to publish and subscribe, which is not feasible in most cases, as it necessitates additional code changes in the application server code. I will replace the generic implementation, which will work without any code changes to the server or receiver client.

This has been on my want-to-have list for a while! Thanks so much for the effort with the PR. I am currently reading up on gRPC - then I will give the PR a good review. 

From my initial thoughts - and what you pertain to in your comment - would it make sense to follow how Bento works with http, we have a `http_client` and `http_server` inputs & outputs - so like `grpc_client` & `grpc_server`. What we have on this PR would be the `grpc_client` I am thinking. 

When you say you "will replace the generic implementation" - is that a change you wish to make in a follow-up or change on this PR?

akoserwal:
> > One issue with the plugin is that it requires the gRPC server to implement `pb.UnimplementedBentoIngressServer` to publish and subscribe, which is not feasible in most cases, as it necessitates additional code changes in the application server code. I will replace the generic implementation, which will work without any code changes to the server or receiver client.
> 
> This has been on my want-to-have list for a while! Thanks so much for the effort with the PR. I am currently reading up on gRPC - then I will give the PR a good review.
> 
> From my initial thoughts - and what you pertain to in your comment - would it make sense to follow how Bento works with http, we have a `http_client` and `http_server` inputs & outputs - so like `grpc_client` & `grpc_server`. What we have on this PR would be the `grpc_client` I am thinking.
> 
> When you say you "will replace the generic implementation" - is that a change you wish to make in a follow-up or change on this PR?

yes

jem-davies:
> yes

Which?

akoserwal:
> > yes
> 
> Which? sorry, I can create a follow up pr, that will be better as it will new changes.

internal/impl/grpc/grpc_dialopts.go
```diff
@@ -0,0 +1,151 @@
...
+		useTLS: useTLS,
+		cfg:    cfg,
+	}
+}
+
+func (b *dialOptionsBuilder) withTransport() *dialOptionsBuilder {
+	if b.useTLS {
+		// Use default system roots; for advanced CA/mTLS provide a custom config downstream.
+		tlsCfg := &tls.Config{MinVersion: tls.VersionTLS12}
+		b.opts = append(b.opts, grpc.WithTransportCredentials(credentials.NewTLS(tlsCfg)))
```
jem-davies, line 67: This whole tls setup is really confusing - seems like we use `tlsConf, tlsEnabled, _ := conf.FieldTLSToggled("tls")` in the constructor but then here we are just replacing the tlsConf with `&tls.Config{MinVersion: tls.VersionTLS12}`.

internal/impl/grpc/init.go
jem-davies, line 1: I think that it would be better if we altered the PR such that we have `input_grpc_client.go` & `output_grpc_client.go` files that: 

- register the component in `init()`
- define the component spec & constructor

So effectively removing the `internal/impl/grpc/init.go` & `internal/impl/grpc/register.go`. It's just the way the vast majority of plugins are implemented - not that this way is flawed.

internal/impl/grpc/register.go
```diff
@@ -0,0 +1,169 @@
...
+		address, err := conf.FieldString("address")
+		if err != nil {
+			return nil, err
+		}
+		consumerID, err := conf.FieldString("consumer_id")
+		if err != nil {
+			return nil, err
+		}
+		// Read optional auth/opts
+		bearer, _ := conf.FieldString("bearer_token")
```
jem-davies, line 57: We normally would return the errors here.

internal/impl/grpc/register.go
```diff
@@ -0,0 +1,169 @@
...
+		"Consume frames from a gRPC server.",
+		`
+input:
+  grpc:
+    address: 127.0.0.1:50051
+    consumer_id: example-client
+`,
+	)
+
+	return service.RegisterInput("grpc", spec, func(conf *service.ParsedConfig, res *service.Resources) (service.Input, error) {
```
jem-davies, line 47:
```suggestion
	return service.RegisterInput("grpc_client", spec, func(conf *service.ParsedConfig, res *service.Resources) (service.Input, error) {
```

internal/impl/grpc/register.go
```diff
@@ -0,0 +1,169 @@
...
+		"Send messages to a gRPC server via a client-side stream.",
+		`
+output:
+  grpc:
+    address: 127.0.0.1:50051
+    max_in_flight: 64
+`,
+	)
+
+	return service.RegisterOutput("grpc", spec, func(conf *service.ParsedConfig, res *service.Resources) (service.Output, int, error) {
```
jem-davies, line 122:
```suggestion
	return service.RegisterOutput("grpc_client", spec, func(conf *service.ParsedConfig, res *service.Resources) (service.Output, int, error) {
```

internal/impl/grpc/grpc_dialopts.go
```diff
@@ -0,0 +1,151 @@
...
+	authority           string
+	userAgent           string
+	loadBalancingPolicy string
+	maxSendMsgBytes     int
+	maxRecvMsgBytes     int
+	keepalive           keepaliveConfig
+	auth                authConfig
+}
+
+func buildDialOptions(_ context.Context, useTLS bool, cfg clientOpts) ([]grpc.DialOption, error) {
```
jem-davies, line 36: How come we have a ctx in the call signature if we aren't using it?

internal/impl/grpc/grpc_dialopts.go
jem-davies, line 1: Not sure if we need this builder pattern - can't we just create a `[]grpc.DialOption` in the constructor?

internal/impl/grpc/grpc_input.go
```diff
@@ -0,0 +1,187 @@
...
+			cErr := g.conn.Close()
+			if cErr != nil {
+				err = cErr
+			}
+			g.conn = nil
+		}
+	})
+	// Allow time for close
+	_ = ctx
+	time.Sleep(5 * time.Millisecond)
```
jem-davies, line 185: Would need to properly handle context termination, and tidy up any created resources properly.

internal/impl/grpc/register.go
```diff
@@ -0,0 +1,169 @@
...
+	return nil
+}
+
+func registerGRPCInput() error {
+	spec := service.NewConfigSpec().
+		Version("1.6.0").
+		Categories("Services").
+		Summary("Receive messages from a gRPC server-side stream using a simple proto contract.").
+		Description("Subscribes to a server-side stream and emits each received frame as a message. Message body is taken from Frame.data and metadata from Frame.metadata. Supports TLS, bearer token auth, custom per-RPC headers, keepalive parameters, and client options such as authority, user agent, and load balancing policy.").
+		Field(service.NewStringField("address").Description("Server address to connect to, e.g. host:port").Default("127.0.0.1:50051")).
```
jem-davies, line 23: Would need to add descriptions to the fields, and more description for the components themselves ideally. Would want to talk about the Frame schema etc...

jem-davies (changes requested):
I think that I would want to resolve some of the obvious issues - then spend more time looking at the implementation. 

It has been on my radar to do some gRPC components for Bento - I think that we might want to think about if this approach is the right one - and if we could do something more dynamic so we don't have to use protoc to generate go source at compile time, then have the user stuck with a vague protobuf schema. If we could take the approach like https://github.com/fullstorydev/grpcurl so it's more dynamic. 

I am thinking that I will but effort into writing a suite of gRPC components. That will maybe will work with the 4 types of request/response setups: unary RPCs, Server streaming RPCs, Client streaming RPCs and Bidirectional streaming RPCs.

akoserwal:
@jem-davies I have a separate implementation on the way, I will have a pr open for it. I would suggest wait until I have new Pr open with changes.

jem-davies:
> @jem-davies I have a separate implementation on the way, I will have a pr open for it. I would suggest wait until I have new Pr open with changes.

Ok - going to mark this PR as draft

jem-davies:
Closed for #458

PR #453, opened 2025-08-27
refactor(sql/conn): Rework SQL connection builder for AWS and Azure

gregfurman:
## Motivation

There is a race condition when initialising multiple sql components as the global `var AWSGetCredentialsGeneratorFn` can be called from multiple goroutines on connect.

## Changes

- Rework the `sql/azure` and `sql/aws` subpackages to instead use `sync.Values` and no longer make unsafe-references a global variable.

### DIFF
--- a/internal/impl/sql/conn_fields.go
+++ b/internal/impl/sql/conn_fields.go
@@ -221,7 +221,10 @@ func (c *connSettings) apply(ctx context.Context, db *sql.DB, log *service.Logge
 	})
 }
 
-func getDsnBuilder(conf *service.ParsedConfig) (DSNBuilder, error) {
+func getDsnBuilder(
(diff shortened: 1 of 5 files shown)


internal/impl/sql/conn_fields.go
```diff
@@ -347,15 +364,6 @@ func sqlOpenWithReworks(ctx context.Context, logger *service.Logger, driver, dsn
 		logger.Info("Updated dsn with info from AWS")
```
jem-davies: Will need either to build a different log string based on where the dsn info has come from or make this log more generic

gregfurman: Ahh nice catch. Will change!

PR #454, opened 2025-08-28
fix(kafka_franz): Validate non-zero seed_brokers attribute

gregfurman:
## Motivation

Handles the edge-case where no `seed_brokers` are set in the Bento config. 

In addition, since the franz-go library allows a new client to be initialised without any `seed_brokers`, the client ends up not connecting to any broker -- resulting in a no reads or writes taking place silently.

## Changes
- Returns an error in the `input/kafka_franz` and `output/kafka_franz` components when no seed brokers are defined.

### DIFF
--- a/internal/impl/kafka/input_kafka_franz.go
+++ b/internal/impl/kafka/input_kafka_franz.go
@@ -22,6 +22,8 @@ import (
 	"github.com/warpstreamlabs/bento/public/service"
 )
 
+var errInvalidSeedBrokerCount = errors.New("you must provide at least one address in 'seed_brokers'")
+
 func franzKafkaInputConfig() *service.ConfigSpec {
 	return service.NewConfigSpec().
 		Beta().
@@ -291,6 +293,10 @@ func newFranzKafkaReaderFromConfig(conf *service.ParsedConfig, res *service.Reso
 	}
 	for _, b := range brokerList {
 		f.seedBrokers = append(f.seedBrokers, strings.Split(b, ",")...)
+	}
+
+	if len(f.seedBrokers) == 0 {
+		return nil, errInvalidSeedBrokerCount
 	}
 
 	if f.autoOffsetReset, err = getOffsetReset(conf); err != nil {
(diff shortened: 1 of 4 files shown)


gregfurman:
>  When we have a seedBrokers like this it seems the current behaviour is to default to 127.0.0.1:9021.

I double checked and this is not actually the behaviour. If we set the `SeedBrokers` when creating a client, even if `seeds` is empty, then the usage of `cfg.seedBrokers[:0]` returns an empty array and effectively resets the `cfg.seedBrokers` to an empty slice -- meaning no connections are made.

See snippet from `franz-go`
```golang
func SeedBrokers(seeds ...string) Opt {
	return clientOpt{func(cfg *cfg) { cfg.seedBrokers = append(cfg.seedBrokers[:0], seeds...) }}
}
```

> Adding a LintRule to check seedBrokers length earlier sound better than what we do a the moment imo.

I was thinking about this approach. It ended up looking a bit too complicated since we allow for multiple elements as well as elements seperated by commas i.e

```yaml
seed_brokers:
  - "broker_1"
  - "broker_2,broker_3"
```

Then we could also end up with edge-cases like:

```yaml
seed_brokers:
  - ${BROKER_A:}
  - ${BROKER_B:}
```

where technically the array is no longer empty (since it'll render to `[ "", "" ]`). 

So an option to flatten this would be:

```coffee
root = this.fold([], item -> item.tally.concat(item.value.split(",")))
```

but then we need to account for cases where there are no spaces between elements (i.e `"broker_1, broker_2"`) 🫠 

So my conclusion was that it'd be easier to do this in Go lol.

internal/impl/kafka/input_kafka_franz.go
```diff
@@ -53,7 +56,7 @@ This input adds the following metadata fields to each message:
 			Description("A list of broker addresses to connect to in order to establish connections. If an item of the list contains commas it will be expanded into multiple addresses.").
 			Example([]string{"localhost:9092"}).
 			Example([]string{"foo:9092", "bar:9092"}).
-			Example([]string{"foo:9092,bar:9092"})).
+			Example([]string{"foo:9092,bar:9092"}).LintRule(``)).
```
jem-davies: has this been left in by mistake?

gregfurman: Ugh yeah sorry. Good catch!

jem-davies (approved):

PR #455, opened 2025-08-31
feat(bloblang/numbers_v2): Add cosine similarity bloblang function

gregfurman:
Adds a new cosine similarity function to bloblang pure methods. 

Allows a user to compare two non-empty and non-null numeric vectors.

### DIFF
--- a/internal/impl/pure/bloblang_numbers_v2.go
+++ b/internal/impl/pure/bloblang_numbers_v2.go
@@ -24,12 +24,12 @@ func init() {
 		Example(
 			"Calculate similarity between vectors",
 			`root.similarity = cosine_similarity([1, 2, 3], [2, 4, 6])`,
-			[2]string{`{}`, `{"similarity": 1.0}`},
+			[2]string{`{}`, `{"similarity": 1}`},
 		).
 		Example(
 			"Orthogonal vectors have zero similarity",
 			`root.similarity = cosine_similarity([1, 0], [0, 1])`,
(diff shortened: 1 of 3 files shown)


internal/impl/pure/bloblang_numbers_v2.go
```diff
@@ -0,0 +1,143 @@
...
+	errVectorInvalidType   = errors.New("vector must be an array of numeric types")
+	errVectorNull          = errors.New("vectors must not be null (all zeros)")
+	errVectorUnequalLength = errors.New("vectors must be equal length")
+	errVectorEmpty         = errors.New("vectors cannot be empty")
+)
+
+func init() {
+	cosineSimilaritySpec := bloblang.NewPluginSpec().
+		Category(query.FunctionCategoryGeneral).
+		Description("Calculates the cosine similarity between two vectors a and b. Vectors must be the same length and not null (all zeros).").
```
jem-davies:
```suggestion
		Description("Calculates the cosine similarity between two vectors a and b. Vectors must be the same length and neither vector can be null (all zeros).").
```

nit - feel this is a bit clearer

gregfurman:
@jem-davies I think it's fine on the functions page tbh. Otherwise, I think a function is more ergnonomic for something like this since we already have other math functions working in the same way (i.e `pow(x, y)`)

jem-davies:
Ok merge it then

jem-davies (approved):

PR #458, opened 2025-09-02
Adding gRPC plugin

akoserwal:
closes https://github.com/warpstreamlabs/bento/issues/433


jem-davies:
@akoserwal - thanks again for the effort - I see that the name of the PR is prefixed 'DRAFT:' but the PR isn't marked as draft. Should this be marked as Draft? Is it ready for review or are more changes coming?

akoserwal:
> @akoserwal - thanks again for the effort - I see that the name of the PR is prefixed 'DRAFT:' but the PR isn't marked as draft. Should this be marked as Draft? Is it ready for review or are more changes coming?

yes, it should be Draft for now

akhenakh:
warning there is a binary in the PR [cmd/tools/grpc_test_server/grpc_test_server](https://github.com/warpstreamlabs/bento/pull/458/files#diff-f3b0cca781c47e1a867a9ffa8f3fd637d6769600f6ca437b1a33186412dc9f1b)

jem-davies (commented):
I think that this approach using [protoreflect](https://github.com/jhump/protoreflect) is much better than the previous PR [add gRPC support](https://github.com/warpstreamlabs/bento/pull/434) - having to have to link code before compilation wouldn't really fit with Bento. 

But then there is still a lot of issues with the PR as it currently stands. 

 I feel like in the last PR (linked above) I gave the idea of using the `http` components as inspiration for the gRPC components - they are quite similar in many ways. But it doesn't seem like much of what we have here is similar to that - and therefore not idiomatic to the project.
 
I have started to implement a `grpc_client` output on https://github.com/warpstreamlabs/bento/pull/546 - which is using the http client as a guide - so we have sync responses & oAuth2 ...
 
 Also a lot of the code pertains to connection_pool & session manager - and I am not sure how important that is going to be.

PR #462, opened 2025-09-04
replace calls to TryInterpolatedX with MessageBatchInterpolationExecutor

jem-davies:


internal/impl/kafka/output_kafka_franz.go
```diff
@@ -375,32 +375,29 @@ func (f *franzKafkaWriter) WriteBatch(ctx context.Context, b service.MessageBatc
...
+	batchInterpolator := f.newInterpolationExecutor(b)
+
 	records := make([]*kgo.Record, 0, len(b))
 	for i, msg := range b {
 		var topic string
-		if topic, err = b.TryInterpolatedString(i, f.topic); err != nil {
-			return fmt.Errorf("topic interpolation error: %w", err)
+		var key []byte
+		var partition int32
+		var hasPartition bool
```
gregfurman: Do we need to define these a priori?

internal/impl/kafka/output_kafka_franz.go
```diff
@@ -375,32 +375,29 @@ func (f *franzKafkaWriter) WriteBatch(ctx context.Context, b service.MessageBatc
...
 
 		record := &kgo.Record{Topic: topic}
 		if record.Value, err = msg.AsBytes(); err != nil {
 			return
 		}
-		if f.key != nil {
-			if record.Key, err = b.TryInterpolatedBytes(i, f.key); err != nil {
-				return fmt.Errorf("key interpolation error: %w", err)
-			}
+		if key != nil {
```
gregfurman: Can we rather have:

```go
value, err := msg.AsBytes()
if err != nil {
   return
}

record := &kgo.Record{
   Topic: topic,
   Key: key,
   Partition: partition,
}
```

This avoids the unnecessary allocation to `record := &kgo.Record{Topic: topic}` if it turns out that `msg.AsBytes` errors out.

internal/impl/kafka/output_kafka_franz.go
```diff
@@ -430,3 +427,54 @@ func (f *franzKafkaWriter) Close(ctx context.Context) error {
 	f.disconnect()
 	return nil
 }
+
+//------------------------------------------------------------------------------
+
+func (f *franzKafkaWriter) newInterpolationExecutor(batch service.MessageBatch) (ie interpolationExecutor) {
```
gregfurman, line 433: Wow nice! I think we should create a service method for something like this that allows us to define and execute multiple of these interpolated expressions against a batch in a single call.

Seems that each time we call `TryInterpolatedString` or `InterpolationExecutor` we create more allocations equal to (or greater than) the size of the batch. Think the speedup is because `InterpolationExecutor` does this less than `TryInterpolatedString`.

Any idea why we see such a speedup? Am thinking that keeping all allocations within this struct method could have saved some memory 🤔

internal/impl/kafka/output_kafka_franz.go
```diff
@@ -430,3 +427,54 @@ func (f *franzKafkaWriter) Close(ctx context.Context) error {
...
+	return ie
+}
+
+type interpolationExecutor struct {
+	topicExecutor     *service.MessageBatchInterpolationExecutor
+	keyExecutor       *service.MessageBatchInterpolationExecutor
+	partitionExecutor *service.MessageBatchInterpolationExecutor
+}
+
+func (ie *interpolationExecutor) exec(i int) (topic string, key []byte, partition int32, hasPartition bool, err error) {
```
gregfurman: nit: would be nice to rather return a `result` or something here -- would help make it more generalisable.

internal/impl/kafka/output_kafka_franz.go
```diff
@@ -375,32 +375,29 @@ func (f *franzKafkaWriter) WriteBatch(ctx context.Context, b service.MessageBatc
...
-			}
-			partInt, err := strconv.Atoi(partStr)
-			if err != nil {
-				return fmt.Errorf("partition parse error: %w", err)
-			}
-			record.Partition = int32(partInt)
+		if hasPartition {
+			record.Partition = partition
 		}
 		_ = f.metaFilter.Walk(msg, func(key, value string) error {
```
gregfurman: Can we also make this `metaFilter.Walk` conditional? There are quite a lot of unnecessary allocations here. Like, a batch of `10_000` records where the `metaFilter == nil` will just allocate a bunch of memory (for each of these `func` types) that wont ever execute. 

i.e
```go
if f.metaFilter.IsEmpty() {
   // do the filtering
}

```

internal/impl/kafka/output_kafka_franz.go
```diff
@@ -430,3 +427,54 @@ func (f *franzKafkaWriter) Close(ctx context.Context) error {
 	f.disconnect()
 	return nil
 }
+
+//------------------------------------------------------------------------------
+
+func (f *franzKafkaWriter) newInterpolationExecutor(batch service.MessageBatch) (ie interpolationExecutor) {
+	ie.topicExecutor = batch.InterpolationExecutor(f.topic)
```
gregfurman, line 434: FYI: Every one of these will create another slice of `*service.Message` with a capacity of the current batch (as mentioned above):

```go
func (b MessageBatch) InterpolationExecutor(i *InterpolatedString) *MessageBatchInterpolationExecutor {
	msg := make(message.Batch, len(b)) // <--- more allocations...
	for i, m := range b {
		msg[i] = m.part
	}

	return &MessageBatchInterpolationExecutor{
		oldBatch: msg,
		i:        i.expr,
	}
}
```

internal/impl/kafka/output_kafka_franz.go
```diff
@@ -375,32 +375,29 @@ func (f *franzKafkaWriter) WriteBatch(ctx context.Context, b service.MessageBatc
 		return service.ErrNotConnected
 	}
 
+	batchInterpolator := f.newInterpolationExecutor(b)
+
 	records := make([]*kgo.Record, 0, len(b))
 	for i, msg := range b {
```
gregfurman, line 381: Can we rather have a `for i := range b {`? This could avoid the unnecessary allocation of `msg` until we actually need it (ie when extracting `value`).

internal/impl/kafka/output_kafka_franz.go
jem-davies: Yeah I think that we need to raise a github issue to remove all the TryInterpolatedString/Bytes calls throughout the codebase that are executed inside a loop of a message batch - and when we do that we can tackle the generalisation -wdyt?

internal/impl/kafka/output_kafka_franz.go
jem-davies:
```go
if !f.metaFilter.IsEmpty() { 
    // do the filtering
}
```

if not empty right?

internal/impl/kafka/output_kafka_franz.go
jem-davies: Not sure if the trade off between using the idiomatic `for i, msg := range b {` and a performance gain is worth it here. 

Would need to investigate how much performance this will give - but this PR is just a step in the right direction - we can iterate on it there as well - imo

internal/impl/kafka/output_kafka_franz.go
gregfurman: Yes sorry I meant if not empty 😄

jem-davies:
@gregfurman have pushed more commit to address the feedback - thanks 🙏

internal/impl/kafka/output_kafka_franz.go
```diff
@@ -375,40 +375,37 @@ func (f *franzKafkaWriter) WriteBatch(ctx context.Context, b service.MessageBatc
...
-		var topic string
-		if topic, err = b.TryInterpolatedString(i, f.topic); err != nil {
-			return fmt.Errorf("topic interpolation error: %w", err)
-		}
 
-		record := &kgo.Record{Topic: topic}
-		if record.Value, err = msg.AsBytes(); err != nil {
-			return
+		topic, key, partition, execErr := batchInterpolator.exec(i)
```
gregfurman, line 383: nit: this is why I don't like these named types in go 😞  if we remove that `(err error)` then we can just use `err` here and `berr`.

gregfurman:
@jem-davies Can you fix the PR title before merging?

Issue #469, opened 2025-09-15
Bento does not refresh IAM AWS token in version 1.10.1


adrianhaj:
**Description**
When using Bento with IAM Roles for RDS and the following configuration:
```
input:
  label: "storage_input"
  aws_s3:
    bucket: "some-bucket"
    region: "eu-west-1"
    scanner:
      lines: {}
    delete_objects: true
    sqs:
      url: https://some-example.sqs
buffer:
  system_window:
    timestamp_mapping: root = now()
    size: 1m0s
pipeline:
  processors:
    - switch:
        - check: metadata("s3_key").contains("request_") && content().length() > 0
          processors:
            - group_by_value:
                value: ${! json("_workspaceId") } }
            - archive:
                format: json_array
            - mapping: root.data = this
            - http:
                url: http://localhost:8000/api/v1/process
                verb: POST
                headers:
                  Content-Type: application/json
                timeout: 10m0s
            - processors:
                - mapping: root = this.results
                - unarchive:
                    format: json_array
                - mapping: 'meta grouping = "request"'
        - check: metadata("s3_key").contains("response_") && content().length() > 0
          processors:
            - group_by_value:
                value: ${! json("_workspaceId") } }
            - archive:
                format: json_array
            - mapping: root.data = this
            - http:
                url: http://localhost:8000/api/v1/processResponse
                verb: POST
                headers:
                  Content-Type: application/json
                timeout: 10m0s
            - processors:
                - mapping: root = this.results
                - unarchive:
                    format: json_array
                - mapping: 'meta grouping = "response"'
output:
  switch:
    cases:
      - check: meta("grouping") == "request"
        output:
          label: request_aggregator_output_database
          sql_raw:
            driver: postgres
            dsn: postgresql://test_user@test-rds.dadasdasd.eu-west-1.rds.amazonaws.com:5432/some_database?sslmode=require
            args_mapping: root = [this.timestamp,this.column_name,this.value_counts]
            query: INSERT INTO ${! this.table_name }(timestamp, column_name, value_counts) VALUES ($1, $2, $3);
            unsafe_dynamic_query: true
            iam_enabled: true
            batching:
              count: 50
              period: 40s
      - check: meta("grouping") == "response"
        output:
          label: postgresql://test_user@test-rds.dadasdasd.eu-west-1.rds.amazonaws.com:5432/some_database?sslmode=require
          sql_raw:
            driver: postgres
[...]

PR #470, opened 2025-09-15
add ping to sql output components

jem-davies:
PR adds a 'Ping' check to the WriteBatch function of the sql output components - so if a connection cannot be established we return a service.ErrNotConnected. Bento will then go into a retry loop exec'ing the Connect function, it's necessary to use the Connect func because we might need to rebuild the DSN (i.e. RDS IAM etc.)

### DIFF
--- a/internal/impl/sql/output_sql_insert.go
+++ b/internal/impl/sql/output_sql_insert.go
@@ -232,6 +232,13 @@ func (s *sqlInsertOutput) WriteBatch(ctx context.Context, batch service.MessageB
 	s.dbMut.RLock()
 	defer s.dbMut.RUnlock()
 
+	if s.driver != "trino" {
+		if err := s.db.PingContext(ctx); err != nil {
+			s.db = nil
+			return service.ErrNotConnected
+		}
+	}
+
 	insertBuilder := s.builder
 
 	var tx *sql.Tx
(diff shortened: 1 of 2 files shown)


jem-davies:
FIx for https://github.com/warpstreamlabs/bento/issues/469

internal/impl/sql/output_sql_raw.go
```diff
@@ -208,6 +208,13 @@ func (s *sqlRawOutput) WriteBatch(ctx context.Context, batch service.MessageBatc
 	s.dbMut.RLock()
 	defer s.dbMut.RUnlock()
 
+	if s.driver != "trino" {
```
gregfurman, line 211: How come we're excluding Trino?

gregfurman (approved):

internal/impl/sql/output_sql_insert.go
```diff
@@ -232,6 +232,13 @@ func (s *sqlInsertOutput) WriteBatch(ctx context.Context, batch service.MessageB
 	s.dbMut.RLock()
 	defer s.dbMut.RUnlock()
 
+	if s.driver != "trino" {
```
gregfurman, line 235: Also, do the docs need updating?

internal/impl/sql/output_sql_raw.go
jem-davies: was breaking the int test ...

internal/impl/sql/output_sql_insert.go
jem-davies: hmm don't think so

Issue #474, opened 2025-09-19
azure_blob_storage input does not delete objects when using delete_objects: true with targets_input


adrianhaj:
**Description**
When using the [azure_blob_storage input](https://warpstreamlabs.github.io/bento/docs/components/inputs/azure_blob_storage)
 with delete_objects: true, objects are not being deleted from Azure Blob Storage if the input is connected to a targets_input (e.g., MQTT).

Steps to Reproduce:

1. Configure azure_blob_storage input with:
```
input:
  label: azure_input
  azure_blob_storage:
    storage_account: ${AZURE_STORAGE_ACCOUNT}
    storage_access_key: ${AZURE_STORAGE_KEY}
    container: my-container
    delete_objects: true
targets_input: 
  mqtt:
    urls: [ ${MQTT_URL} ]
    topic: "test"

```
2. Upload an object to the container.

3. Object is consumed and forwarded to the mqtt target.

4. Object remains in the Azure Blob Storage container instead of being deleted.

**Expected Behavior**
With `delete_objects: true`, objects should be removed from the container once they are successfully consumed and forwarded.

**Actual Behavior**
Objects remain in the container and are not deleted.

**Environment**
Bento version: 1.10.2
Deployment: k8s
Azure Storage account type: Azure Blob Storage

**Additional Context**
Works fine when consuming directly from azure_blob_storage without targets_input.
Issue seems specific to targets_input pipeline routing.

adrianhaj:
Fixed in this pr: https://github.com/warpstreamlabs/bento/pull/475

PR #475, opened 2025-09-19
fix: fixing issue when using input target for azure blob storage

adrianhaj:
Fix related to https://github.com/warpstreamlabs/bento/issues/474


adrianhaj:
@jem-davies @gregfurman could you please review it?

PR #480, opened 2025-09-19
exclude huggingface components arm build

jem-davies:


gregfurman:
Should this not only be excluded from the public packages?

jem-davies:
> Should this not only be excluded from the public packages?

altered 🙏

gregfurman:
Should be good. We should alert the folks upstream about this issue

gregfurman (approved):

PR #481, opened 2025-09-22
Adding support for azure service bus as input with Azure Default Credentials support

adrianhaj:


adrianhaj:
@jem-davies @gregfurman could I ask you for review this changes? I want to ask if it will be possible to release 1.11.1 with this changes ?

jem-davies:
> @jem-davies @gregfurman could I ask you for review this changes? I want to ask if it will be possible to release 1.11.1 with this changes ?

Thanks for the contribution 😎  - I would suggest considering implementing a plugin, while this PR is in review. 

We don't typically add components in a patch release, and there are a few other PRs to review as well...

internal/impl/azure/input_service_bus_queue.go
```diff
@@ -0,0 +1,610 @@
...
+	}
+
+	return conf, nil
+}
+
+func sbqSpec() *service.ConfigSpec {
+	return service.NewConfigSpec().
+		Beta().
+		Categories("Services", "Azure").
+		Version("1.0.0").
```
gregfurman:
```suggestion
		Version("1.13.0").
```

internal/impl/azure/input_service_bus_queue.go
```diff
@@ -0,0 +1,610 @@
...
+			return nil
+		}
+
+		if res != nil {
+			errStr := res.Error()
+			for _, p := range a.conf.nackRejectPatterns {
+				if p.MatchString(errStr) {
+					a.m.RLock()
+					receiver := a.receiver
+					a.m.RUnlock()
```
gregfurman: Think you can just retrieve the pointer to this once outside the loop instead of a getting a new reference each time, unless you're concerned it's going to be set to `nil` while an `ackFunc` is still running.

internal/impl/azure/input_service_bus_queue.go
```diff
@@ -0,0 +1,610 @@
...
+					if receiver != nil {
+						if err := receiver.DeadLetterMessage(actx, msg, nil); err != nil {
+							// If dead lettering fails, abandon the message instead
+							select {
```
gregfurman: Why can we not just run `receiver.AbandonMessage` with a fire-and-forget if dead lettering fails here?

internal/impl/azure/input_service_bus_queue.go
```diff
@@ -0,0 +1,610 @@
...
+
+		if res != nil {
+			errStr := res.Error()
+			for _, p := range a.conf.nackRejectPatterns {
+				if p.MatchString(errStr) {
+					a.m.RLock()
+					receiver := a.receiver
+					a.m.RUnlock()
+					if receiver != nil {
+						if err := receiver.DeadLetterMessage(actx, msg, nil); err != nil {
```
gregfurman: **question**: These requests will automatically cancel when the `receiver.Close` is called, right?

internal/impl/azure/input_service_bus_queue.go
```diff
@@ -0,0 +1,610 @@
...
+				}
+			}
+			select {
+			case a.nackMessagesChan <- msg:
+			case <-actx.Done():
+				return actx.Err()
+			case <-a.closeSignal.SoftStopChan():
+			}
+			return nil
+		}
```
gregfurman: **suggestion(non-blocking):** Perhaps consider extracting out some of the nested logic where possible.

```suggestion
		if res == nil {
			select {
			case a.ackMessagesChan <- msg:
			case <-actx.Done():
				return actx.Err()
			case <-a.closeSignal.SoftStopChan():
			}
			return nil
		}

		errStr := res.Error()
		for _, p := range a.conf.nackRejectPatterns {
			if !p.MatchString(errStr) {
				continue
			}

			a.m.RLock()
			receiver := a.receiver
			a.m.RUnlock()
			if receiver == nil {
				return nil
			}

			if err := receiver.DeadLetterMessage(actx, msg, nil); err != nil {
				// If dead lettering fails, abandon the message instead
				return actx.Err()
			}
			select {
			case a.nackMessagesChan <- msg:
			case <-actx.Done():
				return actx.Err()
			case <-a.closeSignal.SoftStopChan():
			}
			return nil
		}

		return nil
	}
```

gregfurman:
Hey! @adrianhaj 👋  Thanks for the super neat contribution 🤩 I'm busy reviewing this PR, and should be done by a bit later today.

internal/impl/azure/input_service_bus_queue.go
```diff
@@ -0,0 +1,610 @@
...
+	}
+
+	a.receiver, err = a.client.NewReceiverForQueue(a.conf.queueName, receiverOptions)
+	if err != nil {
+		return fmt.Errorf("failed to create Service Bus receiver: %w", err)
+	}
+
+	var wg sync.WaitGroup
+	wg.Add(2)
+	go a.readLoop(&wg)
```
gregfurman: We could use the new `wg.Go()` here instead of needing to pass in a pointer to the waitgroup

internal/impl/azure/input_service_bus_queue.go
```diff
@@ -0,0 +1,610 @@
...
+	go a.ackLoop(&wg)
+	go func() {
+		wg.Wait()
+		a.closeSignal.TriggerHasStopped()
+	}()
```
gregfurman, line 233: **suggestion:** These loop methods could be refactored to take in a ctx that is cancelled when a soft stop signal is called.

https://github.com/warpstreamlabs/bento/blob/7b8149aeef45fbc1308c74f996da0be334c1d6a8/internal/impl/azure/input_service_bus_queue.go#L290-L291

```suggestion
    softStopCtx, cancel := a.closeSignal.SoftStopCtx(context.Background())
	var wg sync.WaitGroup
	wg.Go(func() { a.readLoop(softStopCtx) })
	wg.Go(func() { a.ackLoop(softStopCtx) })
	go func() {
		wg.Wait()
		a.closeSignal.TriggerHasStopped()
	}()
```

Alternatively, we could look into using an `errgroup`:

```suggestion
	softStopCtx, cancel := a.closeSignal.SoftStopCtx(context.Background())
	group, ctx := errgroup.WithContext(softStopCtx)

	group.Go(func() error { a.readLoop(ctx); return nil })
	group.Go(func() error { a.ackLoop(ctx); return nil })

	go func() {
		group.Wait()
		cancel() // cancel the loops
		a.closeSignal.TriggerHasStopped()
	}()
```

internal/impl/azure/input_service_bus_queue.go
```diff
@@ -0,0 +1,610 @@
...
+	defer done()
+
+	exponentialBackOff := backoff.NewExponentialBackOff()
+	exponentialBackOff.InitialInterval = 10 * time.Millisecond
+	exponentialBackOff.MaxInterval = time.Minute
+	exponentialBackOff.MaxElapsedTime = 0
+
+	for {
+		select {
+		case <-a.closeSignal.SoftStopChan():
```
gregfurman: **suggestion:** If `closeAtLeisureCtx` is cancelled can select on the cancelled context, right?
```suggestion
		case <-closeAtLeisureCtx.Done():
```

internal/impl/azure/input_service_bus_queue.go
```diff
@@ -0,0 +1,610 @@
...
+func (a *azureServiceBusQueueReader) readLoop(wg *sync.WaitGroup) {
+	defer wg.Done()
+
+	closeAtLeisureCtx, done := a.closeSignal.SoftStopCtx(context.Background())
+	defer done()
+
+	exponentialBackOff := backoff.NewExponentialBackOff()
+	exponentialBackOff.InitialInterval = 10 * time.Millisecond
+	exponentialBackOff.MaxInterval = time.Minute
+	exponentialBackOff.MaxElapsedTime = 0
```
gregfurman, line 242: **suggestion(non-blocking):** The `backoff.WithContext()` could be used here as well.

internal/impl/azure/input_service_bus_queue.go
```diff
@@ -0,0 +1,610 @@
...
+	for {
+		select {
+		case msg := <-a.ackMessagesChan:
+			a.m.RLock()
+			receiver := a.receiver
+			a.m.RUnlock()
+
+			if receiver != nil {
+				// Use the leisure context first, fall back to hard stop context if cancelled
+				ctx := closeAtLeisureCtx
```
gregfurman: FYI: I believe the soft stop context will be cancelled when the hard stop is called. So this call could be unnecessary.

internal/impl/azure/input_service_bus_queue.go
```diff
@@ -0,0 +1,610 @@
...
+
+func (a *azureServiceBusQueueReader) disconnect() error {
+	a.m.Lock()
+	defer a.m.Unlock()
+
+	// Stop all active lock renewals
+	a.stopAllLockRenewals()
+
+	if a.receiver != nil {
+		if err := a.receiver.Close(context.Background()); err != nil {
```
gregfurman: **suggestion**: Maybe this context could instead be derived from a`a.closeSignal.HardStopCtx()`or  `a.closeSignal.HasStoppedCtx()`. That way if we do a hard stop, we can make sure these requests are canceleld (and not blocking) the close of the component.

Could also be a nice use-case for `errgroup.WithContext` 👀

internal/impl/azure/input_service_bus_queue.go
```diff
@@ -0,0 +1,610 @@
...
+	select {
+	case <-ctx.Done():
+		return ctx.Err()
+	case <-a.closeSignal.HasStoppedChan():
+	}
+	return a.disconnect()
+}
+
+func (a *azureServiceBusQueueReader) disconnect() error {
+	a.m.Lock()
```
gregfurman, line 529: nit: Let's make sure this doesn't deadlock anywhere! 🔒

internal/impl/azure/input_service_bus_queue.go
```diff
@@ -0,0 +1,610 @@
...
+
+				if receiver == nil {
+					return
+				}
+
+				ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+				if err := receiver.RenewMessageLock(ctx, msg, nil); err != nil {
+					a.log.Debugf("Failed to renew lock for message %s: %v", lockToken, err)
+					cancel()
+					return
```
gregfurman: nit: you can just use `defer cancel()` to ensure this is always called

internal/impl/azure/input_service_bus_queue.go
```diff
@@ -0,0 +1,610 @@
...
+	default:
+		close(stopChan)
+	}
+
+	a.locksMutex.Lock()
+	delete(a.activeLocks, lockToken)
+	a.locksMutex.Unlock()
+}
+
+func (a *azureServiceBusQueueReader) stopAllLockRenewals() {
```
gregfurman, line 633: **non-blocking**: I'm going to do a follow-up comment on this, but. think we can re-work this lock tracking using a [`singleflight.Group`](https://pkg.go.dev/golang.org/x/sync/singleflight). Perhaps I'll open a seperate PR targeting this branch on your forked repo and we can see whether you think it works.

gregfurman (commented):
Thank you for the fantastic PR! Really appreciate your usage of shutdown signallers to ensure this component cleans up gracefully 🤩 

Genuinely a lot of fun to review, so kudos on the contribution 😄 Have left some comments and suggestions. Nothing really blocking but would like to hear your feedback on some of what I've raised.

adrianhaj:
> Thank you for the fantastic PR! Really appreciate your usage of shutdown signallers to ensure this component cleans up gracefully 🤩
> 
> Genuinely a lot of fun to review, so kudos on the contribution 😄 Have left some comments and suggestions. Nothing really blocking but would like to hear your feedback on some of what I've raised.

Thank you for review :) I will try to apply all changes as soon as possible :)

internal/impl/azure/input_service_bus_queue.go
adrianhaj: Fixed. Thanks for finding this :)

internal/impl/azure/input_service_bus_queue.go
adrianhaj: Thanks, applied

internal/impl/azure/input_service_bus_queue.go
adrianhaj: You are right, changed.

internal/impl/azure/input_service_bus_queue.go
adrianhaj: Thanks for pointing this.

internal/impl/azure/input_service_bus_queue.go
adrianhaj: Fixed :)

internal/impl/azure/input_service_bus_queue.go
adrianhaj: Done

internal/impl/azure/input_service_bus_queue.go
adrianhaj: Code was refactored to apply this comment.

internal/impl/azure/input_service_bus_queue.go
adrianhaj: Thanks for finding this

internal/impl/azure/input_service_bus_queue.go
adrianhaj: I will be grateful if you can 🙏

internal/impl/azure/input_service_bus_queue.go
adrianhaj: Code refactored.

internal/impl/azure/input_service_bus_queue.go
adrianhaj: Thanks for pointing this.

adrianhaj:
> > Thank you for the fantastic PR! Really appreciate your usage of shutdown signallers to ensure this component cleans up gracefully 🤩
> > Genuinely a lot of fun to review, so kudos on the contribution 😄 Have left some comments and suggestions. Nothing really blocking but would like to hear your feedback on some of what I've raised.
> 
> Thank you for review :) All changes applied. Please check it if any more comments I will be happy to contribute :)

internal/impl/azure/input_service_bus_queue.go
```diff
@@ -0,0 +1,648 @@
...
+
+	for {
+		select {
+		case msg := <-a.ackMessagesChan:
+			a.m.RLock()
+			receiver := a.receiver
+			a.m.RUnlock()
+
+			if receiver != nil {
+				// Use the soft stop context first, fall back to hard stop context if cancelled
```
gregfurman:
```suggestion
```

internal/impl/azure/input_service_bus_queue.go
```diff
@@ -0,0 +1,648 @@
...
+		ReceiveMode: azservicebus.ReceiveModePeekLock,
+	}
+
+	a.receiver, err = a.client.NewReceiverForQueue(a.conf.queueName, receiverOptions)
+	if err != nil {
+		return fmt.Errorf("failed to create Service Bus receiver: %w", err)
+	}
+
+	softStopCtx, cancel := a.closeSignal.SoftStopCtx(context.Background())
+	defer cancel()
```
gregfurman: Calling `defer cancel()` means that the context will cancel when the `Connect()` method returns -- which is not what you want since we expect it to be cancelled by a soft-stop signal.

We should probably ignore it entirely tbh i.e

```
softStopCtx, _ := a.closeSignal.SoftStopCtx(context.Background())
```

internal/impl/azure/input_service_bus_queue.go
```diff
@@ -0,0 +1,648 @@
...
+	}
+
+	softStopCtx, cancel := a.closeSignal.SoftStopCtx(ctx)
+
+	var wg sync.WaitGroup
+	wg.Go(func() { a.readLoop(softStopCtx) })
+	wg.Go(func() { a.ackLoop(softStopCtx) })
+	go func() {
+		wg.Wait()
+		cancel()
```
gregfurman: Was about to comment on this usage of `cancel` 🙂 We can also just ignore it since we don't ever expect to cancel early.

internal/impl/azure/input_service_bus_queue.go
```diff
@@ -0,0 +1,648 @@
...
+			pConf, err := sbqConfigFromParsed(conf)
+			if err != nil {
+				return nil, err
+			}
+			return newAzureServiceBusQueueReader(pConf, mgr)
+		})
+	if err != nil {
+		panic(err)
+	}
+}
```
gregfurman: Can we place this `init` at the top of the file rather?

internal/impl/azure/input_service_bus_queue_test.go
```diff
@@ -0,0 +1,252 @@
...
+	conf := service.NewConfigSpec().
+		Field(service.NewStringField("connection_string").Default("")).
+		Field(service.NewStringField("namespace").Default("")).
+		Field(service.NewStringField("queue").Default("test-queue")).
+		Field(service.NewInputMaxInFlightField().Default(10)).
+		Field(service.NewBoolField("auto_ack").Default(false)).
+		Field(service.NewStringListField("nack_reject_patterns").Default([]any{})).
+		Field(service.NewBoolField("renew_lock").Default(true))
+
+	parsed, err := conf.ParseYAML(`
```
gregfurman: This isn't explicitly testing the config spec you have defined for the service. Instead, you should replace all of this with a direct reference to your defined config spec in `sbqSpec` i.e

```suggestion
parsed, err := sbqSpec().ParseYAML(`
```

internal/impl/azure/input_service_bus_queue_test.go
```diff
@@ -0,0 +1,252 @@
...
+	assert.Nil(t, ackFn)
+	assert.Equal(t, service.ErrNotConnected, err)
+}
+
+func TestServiceBusQueueInit(t *testing.T) {
+	// Test that the input is properly registered
+	// This test ensures the init() function works correctly
+
+	// We can't easily test init() directly, but we can test that
+	// the registration doesn't panic and creates the right input type
```
gregfurman: FYI: you can test this case with a a `_test` namespaced package i.e

```golang
# internal/impl/azure/input_service_bus_queue_test_package.go
package azure_test

import	(
    "testing"

    _ "github.com/warpstreamlabs/bento/public/components/azure" # this import will ensure `init` runs
)

func TestServiceBusQueueInit(t *testing.T) {
 # ...
}
```

But you should instead look at adding an integration test for this case, which also checks if the registration occurs correctly in addition to whether the component can be opened/closed, read data etc.

See [internal/impl/azure/integration_test.go](https://github.com/warpstreamlabs/bento/blob/0a303f85ee73922922038386610fe95e86d41cfe/internal/impl/azure/integration_test.go).

adrianhaj:
Hi @gregfurman should I change something else that you can review it again ?

internal/impl/azure/input_service_bus_queue.go
```diff
@@ -0,0 +1,648 @@
...
+			conf.nackRejectPatterns = append(conf.nackRejectPatterns, r)
+		}
+	}
+
+	return conf, nil
+}
+
+func sbqSpec() *service.ConfigSpec {
+	return service.NewConfigSpec().
+		Beta().
```
gregfurman: Can you comment out this Beta() tag? I think without integration tests we should rather have the component tagged as experimental.

adrianhaj: done

internal/impl/azure/input_service_bus_queue.go
adrianhaj: done

internal/impl/azure/input_service_bus_queue.go
adrianhaj: done. Thanks for comment

internal/impl/azure/input_service_bus_queue.go
adrianhaj: Thank you for comment. Code changed :)

internal/impl/azure/input_service_bus_queue_test.go
adrianhaj: Test refactored. Integration test added.

adrianhaj:
> Hey! Thanks for the ping here. I had written this review like last week but never posted it (maybe GitHub had a moment).
> 
> Overall, the PR looks good to ship! Think the tests can be worked on a bit though -- as we don't actually test whether the reading functionality of the component is working as intended. Adding an integration test is obviously the gold standard but we sometimes go so far as to create mock HTTP servers.
> 
> Think that could be out of scope for this PR though since it's already quite big. Once this is merged, let's open an issue on the backlog for adding integration tests to this 🙂
> 
> Let me know if you have any Qs on my final feedback. Once addressed, I don't see any blockers to merging :shipit:

Thanks, all changes applied :) I was able to add integration tests. Please check :)

adrianhaj:
> LGTM! Once CI passes I'll merge this in :shipit:
> 
> FYI I've rebased your branch and changed the integration test to instead use `mssql-server` instead of `azure-sql-edge` since the latter was problematic on ARM64.
> 
> Otherwise, thanks for the superb PR and addressing all of my feedback! 🎉

Thank you!

PR #487, opened 2025-10-06
feat(metrics): use default Prometheus registry

### LINKED ISSUE #486: Use default Prometheus registry for metrics
### Problem

When using the StreamBuilder API, there is no way to reuse the default Prometheus registry from the main process to expose metrics under the same endpoint as the parent process. This adds additional overhead for monitoring by requiring separate scrape configs for each stream, as well as exposing multiple ports rather than reusing the Prometheus port from the parent process (`9090`).

### Proposed solution

Add a `use_default_registry` boolean field to the `prometheus` metric component config. Setting this field to true would pull the registry from `prometheus.DefaultRegisterer` rather than initializing a new one via `prometheus.NewRegistry()`.

### Prior work

The ability to share a metrics config for multiple streams was introduced in https://github.com/warpstreamlabs/bento/pull/422.

alecmerdler:
Adds a `use_default_registry` field to the Prometheus metrics config.

This solves a problem when using the StreamBuilder API, where the parent Go application exposes metrics on one port, but the Bento stream exposes its metrics on a different port. To simplify monitoring setup by only needing to scrape a single port, setting this field to true will cause Bento to use the default Prometheus registry rather than creating its own.

Resolves #486

### DIFF
--- a/internal/impl/prometheus/metrics_prometheus.go
+++ b/internal/impl/prometheus/metrics_prometheus.go
@@ -32,6 +32,7 @@ const (
 	pmFieldPushInterval                = "push_interval"
 	pmFieldPushJobName                 = "push_job_name"
 	pmFieldFileOutputPath              = "file_output_path"
+	pmFieldDefaultRegistryPath         = "use_default_registry"
 )
 
 func ConfigSpec() *service.ConfigSpec {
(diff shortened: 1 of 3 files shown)


jem-davies (approved):

jem-davies:
@alecmerdler - need to run `make docs`

alecmerdler:
@jem-davies Thanks, re-ran and pushed a new commit.

alecmerdler:
@jem-davies Thank you for the quick review! Love this project and I'm looking forward to contributing more in the future.

Issue #355, opened 2025-06-01
Flaky test: TestHTTPServerOutputSSEHeartbeat


gregfurman:
```
--- FAIL: TestHTTPServerOutputSSEHeartbeat (11.31s)
    output_http_server_test.go:531: Client timed out
```

https://github.com/warpstreamlabs/bento/actions/runs/15377659715/job/43264380460?pr=352#step:6:92

jem-davies:
still fails frequently

PR #488, opened 2025-10-07
test: Increase channel buffer and add ready signal to fix `TestHTTPServerOutputSSEHeartbeat`

HankStat:
Fixes [#355](https://github.com/warpstreamlabs/bento/issues/355)

### Problem
1. `receivedHeartbeats` channel deadlock: Due to the small buffer size (5) for the `receivedHeartbeats channel`, a deadlock happened when the channel filled up before the client's exit condition (2 messages && 3 heartbeats) was met, creating a circular wait and causing the test to time out.
2. Message race condition: The test used a fixed `time.Sleep(1 * time.Second)` to guess when the client goroutine was ready to receive data. On a busy system, the client might not be ready after 1 second, causing the `TestHTTPServerOutputSSEHeartbeat` test to send messages before the client was listening. This resulted in missed messages and the test failing with a `didn't receive enough messages` error.

### Solution
1. Increase channel buffer size: The buffer size for `receivedHeartbeats` has been increased to 50. This provides more than enough capacity to hold all events the client could produce during its 5-second runtime. This ensures the client goroutine will never block on a channel send, completely eliminating the deadlock.
2. Add channel to ensure synchronization: The `time.Sleep` has been replaced with a `clientReady` channel. After the client is ready and connected, it closes the channel to signal it is ready to read data. The `TestHTTPServerOutputSSEHeartbeat` test waits for the channel to be closed before sending any messages. This guarantees that messages are sent only when the client is ready to receive them.

### Verification
The test now correctly and consistently passes on my local machine after running it 200 times using a loop to simulate the flaky behavior. In addition, after running 200 times, the TestHTTPServerOutputSSEHeartbeat test took an average of 1.61 seconds to finish.
```bash
go test -v -race ./internal/impl/io -run ^TestHTTPServerOutputSSEHeartbeat$ -count=200
```
```
make test
make lint
make fmt
```

### DIFF
--- a/internal/impl/io/output_http_server_test.go
+++ b/internal/impl/io/output_http_server_test.go
@@ -366,8 +366,10 @@ http_server:
 	clientDone := make(chan struct{})
 	clientErrors := make(chan error, 1)
 	receivedMessages := make(chan string, 10)
-	receivedHeartbeats := make(chan struct{}, 5)
-
+	// Use larger buffer to prevent the client from blocking when sending heartbeats
+	receivedHeartbeats := make(chan struct{}, 50)
+	// Use channel to wait for client ready
+	clientReady := make(chan struct{}, 1)
 	go func() {
 		defer close(clientDone)
 
@@ -408,7 +410,8 @@ http_server:
 			clientErrors <- fmt.Errorf("unexpected content type: %s", contentType)
 			return
 		}
-
+		// Client is ready, close the channel
+		close(clientReady)
 		// Read the SSE stream
 		buffer := make([]byte, 4096)
 		currentData := ""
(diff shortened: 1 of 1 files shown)


internal/impl/io/output_http_server_test.go
```diff
@@ -366,8 +366,10 @@ http_server:
 	clientDone := make(chan struct{})
 	clientErrors := make(chan error, 1)
 	receivedMessages := make(chan string, 10)
-	receivedHeartbeats := make(chan struct{}, 5)
-
+	// Use larger buffer to prevent the client from blocking when sending heartbeats
+	receivedHeartbeats := make(chan struct{}, 50)
```
gregfurman, line 370: question: Wondering if these heartbeats even need to be buffered... 👀

internal/impl/io/output_http_server_test.go
```diff
@@ -366,8 +366,10 @@ http_server:
...
+	// Use channel to wait for client ready
+	clientReady := make(chan struct{})
```
gregfurman: nit: Consider making this buffered with a length of `1`

HankStat: Thanks for the suggestion. I've amended the commit to use a buffered channel as you recommended.

internal/impl/io/output_http_server_test.go
HankStat: Yes, the buffer is essential here. The test has no receiver for the `receivedHeartbeats` channel during the client's runtime, so an unbuffered channel would cause the sender to block indefinitely and create a deadlock.

PR #494, opened 2025-10-09
feat: expose readiness from StreamBuilder API

### LINKED ISSUE #493: Expose readiness from StreamBuilder API
### Problem

When using the StreamBuilder API, the only way to programmatically check if the stream is ready is by hitting the `/ready` endpoint on the HTTP server. However, there are situations where the HTTP server should be disabled, leaving the caller unable to determine if the stream is ready (which affects downstream health checks).

### Proposed solution

The internal `stream.Type` struct has an `IsReady()` method. We can expose that on the external `service.Stream` interface.

alecmerdler:
Adds an `IsReady()` method to the public `Stream` interface. This allows callers using the Go `StreamBuilder` API to programmatically determine if the stream is active without needing to run the HTTP server.

Resolves #493

### DIFF
--- a/public/service/stream.go
+++ b/public/service/stream.go
@@ -181,14 +181,13 @@ func (s *Stream) Stop(ctx context.Context) (err error) {
 }
 
 // IsReady returns a boolean indicating whether both the input and output layers
-// of the stream are connected. If the stream has not yet been run, an error is
-// returned.
-func (s *Stream) IsReady() (bool, error) {
+// of the stream are connected.
+func (s *Stream) IsReady() bool {
 	s.strmMut.Lock()
 	strm := s.strm
 	s.strmMut.Unlock()
 	if strm == nil {
-		return false, errors.New("stream has not been run yet")
+		return false
 	}
-	return strm.IsReady(), nil
+	return strm.IsReady()
 }
--- a/public/service/stream_builder_test.go
+++ b/public/service/stream_builder_test.go
@@ -62,15 +62,13 @@ generate:
 	strm, err := b.Build()
 	require.NoError(t, err)
 
-	ready, err := strm.IsReady()
-	require.ErrorContains(t, err, "stream has not been run yet")
+	ready := strm.IsReady()
 	require.False(t, ready)
 
 	var wg sync.WaitGroup
 	wg.Go(func() {
 		require.Eventually(t, func() bool {
-			ready, err = strm.IsReady()
-			require.NoError(t, err)
+			ready = strm.IsReady()
 			return ready
 		}, time.Second*5, time.Millisecond*10)
 


public/service/stream.go
```diff
@@ -179,3 +179,16 @@ func (s *Stream) Stop(ctx context.Context) (err error) {
 	err = closeHTTP(ctx)
 	return
 }
+
+// IsReady returns a boolean indicating whether both the input and output layers
+// of the stream are connected. If the stream has not yet been run, an error is
+// returned.
+func (s *Stream) IsReady() (bool, error) {
```
alecmerdler: The docs for `Stream` [indicate that it exposes status checks](https://github.com/warpstreamlabs/bento/pull/494/files#diff-7f758ddc4d467adcca6fce08625c7a48f3c81b7bd26479b7e57269b2ff77f9cfR20-R21), but there was no method for this until now.

public/service/stream_builder_test.go
```diff
@@ -52,6 +52,35 @@ func TestStreamBuilderDefault(t *testing.T) {
...
+
+	strm, err := b.Build()
+	require.NoError(t, err)
+
+	ready, err := strm.IsReady()
+	require.ErrorContains(t, err, "stream has not been run yet")
+	require.False(t, ready)
+
+	var wg sync.WaitGroup
+	wg.Go(func() {
```
gregfurman, line 69: **praise:** Loving the usage of the new `wg.Go` functionality!

public/service/stream.go
gregfurman: Perhaps this was referring to the internal API that exposed `StreamStatus` via REST?

https://github.com/warpstreamlabs/bento/blob/d554ee222c1136e23fb449deafeebf0a6e2ae0b7/internal/stream/manager/api.go#L30-L34

gregfurman: **question**: Unsure if we should be returning an error here. in addition to a `bool`. What cases would an `error` provide more value to the caller as opposed to just the flag?

public/service/stream.go
```diff
@@ -179,3 +179,16 @@ func (s *Stream) Stop(ctx context.Context) (err error) {
...
+	s.strmMut.Lock()
+	strm := s.strm
+	s.strmMut.Unlock()
+	if strm == nil {
+		return false, errors.New("stream has not been run yet")
+	}
+	return strm.IsReady(), nil
```
gregfurman: Also, thoughts on a `RunWithin` method similar to `StopWithin`?

public/service/stream.go
alecmerdler: I don't feel strongly either way, the motivation was to indicate to the caller whether `Run()` has been called. If it hasn't been called, then `ready` will forever return `false`.

I've gone ahead and removed the `error` return value to simplify using this method.

public/service/stream.go
alecmerdler: A `RunWithin()` method would be very useful, if it could be guaranteed!

PR #495, opened 2025-10-09
enhancement: Allow headers config for websocket input handshake

exside:
WARNING: First time contributing here, so please forgive me if I messed up something, lmk and I'll do my best to make it proper! I did run `make lint` and `make fmt` but both threw errors like these:

```
make lint
make: golangci-lint: No such file or directory
make: *** [lint] Error 1
```

and

```
make fmt
xargs: goimports: No such file or directory
make: *** [fmt] Error 127
```

The `make test` does run, but I aborted it because it tests so much unrelated stuff. I did run my added tests directly in the `/impl/io` folder where it passed.

What this PR does is simply add the ability to configure `headers` for the initial handshake request when opening a WebSocket connection via the websocket input like this, values allow for interpolations: 

```yml
input:
  websocket:
    url: wss://some.stream.com
    headers:
      Sec-WebSocket-Protocol: graphql-ws
      User-Agent: ${! uuid_v4() }
      X-Client-ID: ${CLIENT_ID}
```

I tried to implement it the same way like `headers` for the `http_client`.

### DIFF
--- a/internal/impl/io/input_websocket.go
+++ b/internal/impl/io/input_websocket.go
@@ -189,7 +189,6 @@ func (w *websocketReader) Connect(ctx context.Context) error {
 		}
 		headers.Add(k, value)
 	}
-	// w.log.Info("Configured headers: %v", headers)
 
 	err := w.reqSigner(w.mgr.FS(), &http.Request{
 		URL:    w.urlParsed,


internal/impl/io/input_websocket.go
```diff
@@ -169,6 +182,14 @@ func (w *websocketReader) Connect(ctx context.Context) error {
...
 
 	headers := http.Header{}
+	for k, v := range w.headers {
+		value, err := v.TryString(service.NewMessage(nil))
+		if err != nil {
+			return fmt.Errorf(`failed string interpolation on header %q: %w`, k, err)
+		}
+		headers.Add(k, value)
+	}
+	// w.log.Info("Configured headers: %v", headers)
```
gregfurman:
```suggestion
```

internal/impl/io/input_websocket.go
```diff
@@ -169,6 +182,14 @@ func (w *websocketReader) Connect(ctx context.Context) error {
 	}
 
 	headers := http.Header{}
+	for k, v := range w.headers {
+		value, err := v.TryString(service.NewMessage(nil))
```
gregfurman, line 186: **question**: Since this interpolation only ever happens on `Connect`, the headers are effectively statically set unless the component closes and reconnects. 

Is this desirable behaviour? Or should we rather be re-loading these headers from our interpolation functions every time we attempt to read from the websocket source?

internal/impl/io/input_websocket.go
```diff
@@ -35,14 +35,23 @@ func websocketInputSpec() *service.ConfigSpec {
...
 				Example("ws://localhost:4195/get/ws"),
 			service.NewURLField("proxy_url").
 				Description("An optional HTTP proxy URL.").
 				Advanced().Optional(),
+			service.NewInterpolatedStringMapField("headers").
+				Description("A map of custom headers to add to the websocket handshake.").
+				Example(map[string]any{
+					"Sec-WebSocket-Protocol": "graphql-ws",
+					"User-Agent":             `${! uuid_v4() }`,
+					"X-Client-ID":            `${CLIENT_ID}`,
```
gregfurman, line 51: FYI: environment variables can already be used on startup without the need for dynamic bloblang querying. See https://warpstreamlabs.github.io/bento/docs/configuration/interpolation/

exside: Oh, so you mean that's a pretty bad example because env vars do not actually need interpolation because they are handled already earlier?

gregfurman: Yeah they're loaded in once (on startup)

internal/impl/io/input_websocket.go
exside: Can you define `read from the websocket source`? Do you mean onMessage, e.g. whenever a message is read from the websocket? If that's the case I would not be aware of how to send any kind of HTTP headers with those, at least not in a standardized way. So the anser would be "yes" to your question: "Is this desirable behaviour?". 

Thanks for clarifying!

gregfurman: Ahh I see I misunderstood this change. Not super clued up on websockets so excuse my limited understanding 🙂 

So this addition will only add headers on the input handshake request, not on every subsequent onMessage call. 

In that case, would it make more sense to have these headers defined as a string map as opposed to a map of interpolated types? 

Wdyt?

exside: > So this addition will only add headers on the input handshake request, not on every subsequent onMessage call.

yes, that's the idea, e.g. when gorilla opens ws(s):// url it's first a "regular" HTTP request, then a protocol update to WebSockets follows, to my knowledge there's no standardized way of sending something like HTTP headers around in actual websocket messages but as the initial request for the handshake/protocol upgrade is a regular HTTP request it can has HTTP headers like any other which sometimes are expected by the server. Currently the websocket input can't be used with such servers expecting any kind of custom headers as they can't be set.

> In that case, would it make more sense to have these headers defined as a string map as opposed to a map of interpolated types?

I had a look at the http_client input and basically took inspiration from there and its config allows interpolation, so I though it might be the best to keep things conceptually the same? I think it's useful to have interpolation, but maybe there are good arguments against it?

internal/impl/io/input_websocket.go
exside: Got it, I'll adapt the example, thanks for the FYI!

internal/impl/io/input_websocket.go
gregfurman: > yes, that's the idea, e.g. when gorilla opens ws(s):// url [...]

Sounds good 👍 Thanks for the explanation!

> I had a look at the http_client input and basically took inspiration from there and its config allows interpolation,

My concern is that those HTTP components will be repeatedly performing interpolation, whereas the current approach here will only ever perform it once on startup.

So while the approach technically works, the need for that empty `service.NewMessage(nil)` to retrieve the required values indicates that dynamic interpolation is being used conceptually incorrectly.

That is, the fact that we're applying interpolation on a message that is always empty defeats the purpose of dynamically extracting values from some incoming data payload whose values will vary. 

Does that make sense?

See how interpolation is called multiple times throughout the data ingestion process, not just on startup:

https://github.com/warpstreamlabs/bento/blob/5b6056afdef579a23778972b188ecef5a07eddf4/internal/impl/io/input_http_server.go#L621-L627

With all of that said, in lieu of a better solution for dynamically generating config values for `input` components, I'm happy to merge this in and look into a better approach at a later point 🙂

exside: I'm very new to bento/benthos/<insert100namesofotherforks>, so i definitely have to dig a bit around to get the full picture of what you explained above 😬 ! Thanks a lot for the merge!

Issue #354, opened 2025-06-01
Flaky test: TestWorkflowUnwrapResourceBranches


jem-davies:
```
level=info msg="Input type generate is now active" @service=bento label="" path=root.input.batched.child
level=info msg="Sending inproc messages to ID: 82b41a57-cae4-4d76-a99e-8405d1e4dc6b" @service=bento label="" path=root.output
level=info msg="Input type generate is now active" @service=bento label="" path=root.input.batched.child
level=info msg="Sending inproc messages to ID: 02a14312-f5d2-4d60-ae68-7d6dc93338af" @service=bento label="" path=root.output
level=info msg="Receiving inproc messages from ID: 79db3955-126f-4d38-8344-92bcf53a23ea" @service=bento label="" path=root.input
level=info msg="Sending inproc messages to ID: 0be830ce-300a-49d4-ac3e-b2e2bf776f3b" @service=bento label="" path=root.output
panic: Fail in goroutine after TestWorkflowUnwrapResourceBranches has completed

goroutine 95635 [running]:
testing.(*common).Fail(0xc00064ed00)
	/opt/hostedtoolcache/go/1.23.9/x64/src/testing/testing.go:953 +0xcc
testing.(*common).Errorf(0xc00064ed00, {0x10b5c64?, 0x46e8c5?}, {0xc0008b5db0?, 0xf442c0?, 0x101?})
	/opt/hostedtoolcache/go/1.23.9/x64/src/testing/testing.go:1076 +0x5e
github.com/stretchr/testify/assert.Fail({0x12c5360, 0xc00064ed00}, {0xc000ad8[120](https://github.com/warpstreamlabs/bento/actions/runs/15378104666/job/43265352238#step:6:121), 0x2b}, {0x0, 0x0, 0x0})
	/home/runner/go/pkg/mod/github.com/stretchr/testify@v1.10.0/assert/assertions.go:369 +0x370
github.com/stretchr/testify/assert.NoError({0x12c5360, 0xc00064ed00}, {0x12c50e0, 0x19739e0}, {0x0, 0x0, 0x0})
	/home/runner/go/pkg/mod/github.com/stretchr/testify@v1.10.0/assert/assertions.go:1586 +0x125
github.com/warpstreamlabs/bento/internal/impl/pure_test.TestWorkflowUnwrapResourceBranches.func2()
	/home/runner/work/bento/bento/internal/impl/pure/processor_workflow_test.go:1035 +0x47
[...]

PR #503, opened 2025-10-13
test: Prevent goroutine panic in `TestWorkflowUnwrapResourceBranches`

HankStat:
Fixes [#354](https://github.com/warpstreamlabs/bento/issues/354)

### Problem
`TestWorkflowUnwrapResourceBranches` was flaky and would intermittently fail in CI with a `panic: Fail in goroutine after Test... has completed`.
This was caused by a race condition. Background goroutines, the anonymous function passed to `AddConsumerFunc`, were calling assertion methods (`require.NoError`) on the `*testing.T` object. If the main test function returned before these assertions were made, the `*testing.T` object would become invalid, leading to a panic. 

### Solution
All `assert/require` calls were removed from background goroutines to prevent data races on the `*testing.T` object. Instead, errors from asynchronous operations are now captured in variables.

A `sync.WaitGroup` was added to explicitly track and wait for the `ConsumerFunc`'s execution, ensuring this implicitly asynchronous task is complete before the main thread proceeds.

**Why this works**: This solution creates a definitive synchronization point. It forces the main test goroutine to wait until all background work has fully finished. This guarantees assertions are only performed in a safe context where results are ready and the `*testing.T` object is still valid, thereby eliminating the race condition.

### Verification
The test now correctly and consistently passes on my local machine after running it 100 times using a loop to simulate the flaky behavior. In addition, after running 100 times, the `TestWorkflowUnwrapResourceBranches` test took an average of 0.0263 seconds to finish, which is only 0.0015 seconds longer on average than original method.
```bash
go test -v -race ./internal/impl/pure -run ^TestWorkflowUnwrapResourceBranches$ -count=100
```
```
make test
make lint
make fmt
```

### DIFF
--- a/internal/impl/pure/processor_workflow_test.go
+++ b/internal/impl/pure/processor_workflow_test.go
@@ -1017,10 +1017,25 @@ workflow:
 	inFunc, err := strmBuilder.AddProducerFunc()
 	require.NoError(t, err)
 
-	var outValue string
+	var (
+		outValue    string
+		consumerErr error
+		streamErr   error
+		mu          sync.Mutex
+	)
+
+	consumerWg := &sync.WaitGroup{}
+	consumerWg.Add(1)
+
 	require.NoError(t, strmBuilder.AddConsumerFunc(func(ctx context.Context, m *service.Message) error {
+		defer consumerWg.Done()
 		outBytes, err := m.AsBytes()
-		require.NoError(t, err)
+		if err != nil {
+			mu.Lock()
+			consumerErr = err
+			mu.Unlock()
+			return err
+		}
 		outValue = string(outBytes)
 		return nil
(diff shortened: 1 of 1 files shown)


gregfurman:
Hey! Thanks for this. I was unable to re-create on my machine (am running go `v1.25.1` on  `darwin arm64`).

Do you have a CI run or some logs showing this? Want to try re-create on my end.

HankStat:
> Hey! Thanks for this. I was unable to re-create on my machine (am running go `v1.25.1` on `darwin arm64`).
> 
> Do you have a CI run or some logs showing this? Want to try re-create on my end.

Hi, I couldn't reproduce the error on my local machine either. The fix is based on the error log from issue #354

PR #508, opened 2025-10-17
Add Lint Warnings to Public def of Lints & emit Warning Logs when using ServiceBuilder

jem-davies:


public/service/lints.go
```diff
@@ -112,10 +112,29 @@ func convertDocsLintType(d docs.LintType) LintType {
 	return LintCustom
 }
 
+// LintLevel describes the severity level of a linting error.
+// NOTE: These should be kept in sync with ./internal/docs/field.go:726.
```
gregfurman: Maybe make this in in-comment function, since this will be exposed in `public` otherwise.

public/service/stream_builder.go
```diff
@@ -74,6 +74,8 @@ type StreamBuilder struct {
 	env             *Environment
 	lintingDisabled bool
 	envVarLookupFn  func(string) (string, bool)
+
+	lintWarns []Lint
```
gregfurman, line 80: Does it make sense to store/cache lint warnings? I prefer the approach of linting on build/startup (unsure if this is already happening).

gregfurman (changes requested):

public/service/lints.go
jem-davies: It's the same as https://github.com/warpstreamlabs/bento/pull/508/files#diff-cbeda7f53df6b8a29295c69f3ee48389cfcb2fd5f090f03c36c195285f29c82cR12 

Maybe it's not that bad that people will see that - they might check that it's true

public/service/stream_builder.go
jem-davies: Yeah so it's like this because we lint the configs when we call `streambuilder.AddInputYaml` / `.SetYaml` etc. 

But at that time we don't necessarily have a logger until we call `streambuilder.Build()` where if we haven't set a logger via `.SetLogger()` we will get a default, and then we log the cached lintWarns. 

Perhaps there is a better way - this is just the first that came to mind...

public/service/stream_builder.go
```diff
@@ -1102,16 +1109,18 @@ func (s *StreamBuilder) getYAMLNode(b []byte) (*yaml.Node, error) {
...
+	return err
 }
 
-func (s *StreamBuilder) lintYAMLComponent(node *yaml.Node, ctype docs.Type) error {
+func (s *StreamBuilder) lintYAMLComponent(node *yaml.Node, ctype docs.Type) (err error) {
 	if s.lintingDisabled {
 		return nil
 	}
-	return lintsToErr(docs.LintYAML(s.getLintContext(), ctype, node))
+	s.lintWarns, err = lintsToErr(docs.LintYAML(s.getLintContext(), ctype, node))
```
jem-davies: Should be append probably

public/service/stream_builder.go
```diff
@@ -1102,16 +1109,18 @@ func (s *StreamBuilder) getYAMLNode(b []byte) (*yaml.Node, error) {
 	return docs.UnmarshalYAML(b)
 }
 
-func (s *StreamBuilder) lintYAMLSpec(spec docs.FieldSpecs, node *yaml.Node) error {
+func (s *StreamBuilder) lintYAMLSpec(spec docs.FieldSpecs, node *yaml.Node) (err error) {
 	if s.lintingDisabled {
 		return nil
 	}
-	return lintsToErr(spec.LintYAML(s.getLintContext(), node))
+	s.lintWarns, err = lintsToErr(spec.LintYAML(s.getLintContext(), node))
```
jem-davies: Should be append ...

public/service/lints.go
gregfurman: I mean, I'm personally not a fan of that either. I think it serves as more of a note to developers extending bento as opposed to people building piplines 😅 

Like, if I am reading the docs for `public`, it doesn't help me that this `internal` functionality is being mirrored upstream -- whereas for you or me, it makes much more of a difference imo

public/service/stream_builder.go
gregfurman: Ahh I see. Maybe we add an inline TODO or FIXME about this? Then happy to resolve.

public/service/lints.go
jem-davies: Ok have moved them so it won't appear in the go docs

public/service/stream_builder.go
jem-davies: Added a comment for lintWarns to explain what we are doing with that.

PR #510, opened 2025-10-18
feat(buffer_memory): allow buffer spillover

### LINKED ISSUE #509: Spillover mechanism for memory buffer
### Problem

Some inputs do not handle backpressure gracefully, which can lead to increased latency for upstream services. Using the `memory` buffer can alleviate this, but if the buffer fills (from a sudden burst of inputs), then backpressure is applied and upstream services are affected. In certain cases, dropping messages rather than blocking writes to the buffer is preferable when throughput and latency are more important than delivery guarantees.

A more concrete example: using the StreamBuilder API's `AddBatchProducerFunc()` in the request path of an HTTP server to log events to Kafka. When the `memory` buffer is full, the function will block, which will block the request hot path and increase latency. We could "fire and forget" in a goroutine, but a massive burst of traffic will cause a goroutine explosion. The API doesn't provide a clean mechanism for tracking how many items are in the buffer.

### Proposed solution

Add a `spillover` boolean field to the `memory` buffer. When set to true, any incoming message batch to the buffer will be dropped if it will push the buffer over the configured `limit`.

alecmerdler:
Adds the `spillover` boolean field to the `memory` buffer component (default false). When true, any incoming message batch that will put the buffer above its configured `limit` will be dropped.

The motivation for this is to support input sources that do not gracefully handle backpressure and where the desired delivery guarantees are best effort.

Resolves #509

### DIFF
--- a/internal/impl/pure/buffer_memory.go
+++ b/internal/impl/pure/buffer_memory.go
@@ -262,7 +262,7 @@ func (m *memoryBuffer) WriteBatch(ctx context.Context, msgBatch service.MessageB
 
 	for (m.bytes + extraBytes) > m.cap {
 		if m.spilloverEnabled {
-			return component.ErrMessageTooLarge
+			return component.ErrLimitReached
 		}
 
 		m.cond.Wait()
--- a/internal/component/errors.go
+++ b/internal/component/errors.go
@@ -102,4 +102,5 @@ var (
 // Buffer errors.
 var (
 	ErrMessageTooLarge = errors.New("message body larger than buffer space")
+	ErrLimitReached    = errors.New("adding message to buffer will exceed size limit")
 )
--- a/internal/impl/pure/buffer_memory_test.go
+++ b/internal/impl/pure/buffer_memory_test.go
@@ -524,7 +524,7 @@ spillover: true
 	err := block.WriteBatch(ctx, service.MessageBatch{
 		service.NewMessage([]byte("hello")),
 	}, func(ctx context.Context, err error) error { return nil })
-	require.Equal(t, component.ErrMessageTooLarge, err)
+	require.Equal(t, component.ErrLimitReached, err)
 
 	block.EndOfInput()
 


internal/impl/pure/buffer_memory.go
```diff
@@ -250,6 +261,10 @@ func (m *memoryBuffer) WriteBatch(ctx context.Context, msgBatch service.MessageB
 	}
 
 	for (m.bytes + extraBytes) > m.cap {
+		if m.spilloverEnabled {
+			return component.ErrMessageTooLarge
```
alecmerdler: We can introduce a new error type for spillover, if that makes sense.

gregfurman: I think `component.ErrMessageTooLarge` makes sense here 👍

alecmerdler:
@gregfurman @jem-davies Any thoughts on this approach? I'm happy to make any changes.

internal/impl/pure/buffer_memory.go
jem-davies: Yeah I think it would be better to add a different error message - will be easier to debug too.

alecmerdler: Done in https://github.com/warpstreamlabs/bento/pull/510/commits/e36405df98d5e1a2a00ef2ebe0642dd8455485c6 and rebased.

alecmerdler:
Fixed the tests

gregfurman:
@alecmerdler 👋 Thanks for the contribution! 

While I'll give this proper review a bit later today, this PR got thinking whether the `buffer` component could perhaps be a bit too limited, and whether extending it to support conditional buffering could be a useful approach. For example:

```yaml
buffer:
  memory:
    limit: 1000
  drop_on:
    - check: this.errored()
    - check: this == ""
    # etc ...
```

Could be over-engineering but am curious to hear your thoughts 🤔

alecmerdler:
Adding conditional buffering does seem like overkill and starts to mix concerns. What's missing for my use case is having any layer between inputs and buffers where we could drop on backpressure.

Either way, since the `memory` buffer allows configuring the `limit`, I believe it makes sense to expose a parameter that let's you choose what happens when that limit is reached.

gregfurman (approved):
LGTM. Will give some more thought on how we can extend the buffer component but agree adding this `spillover` flag makes sense 👍

PR #515, opened 2025-10-20
promote various component status to Stable

jem-davies:


gregfurman (approved):
LGTM! Thanks for addressing this 👍 

> Stable sets a documentation label on the component indicating that its configuration spec is stable. Plugins are considered experimental by default.

Oh wow I didn't know that. So we need to explicitly set when a component is `Stable`. 

Am unsure if that is desirable behaviour 🤔

Issue #531, opened 2025-10-28
Kinesis Shards not refreshing.


padzor:
### aws_kinesis input not seeing new shards.

- Shards went from 500 to 1000, resulting in kinesis reporting 1500 shards in total (500 inactive, 1000 active) 
- The 500 additional shards were not detected until 24hrs later - the retention period of the shards, at which point it caught up correctly.  This is likely due to the the inactive shards being removed from the system.

## Proposed solutions from @gregfurman 

ListShards returns at most 1000 shards per call. We don't have any pagination here so am unsure whether the Kinesis API strictly adheres to the MaxResults parameter.

Also, the docs state that for ListShards "does not list expired shards.". So after your data retention period is exceeded, the EXPIRED shards are no longer listed, perhaps decreasing the no. of returned shards to an amount small enough for the single ListShards approach to work -- allowing your new shards to be loaded in.

jem-davies:
@padzor / @gregfurman - can we close this now?

gregfurman:
Think we can close with #532 🙂

PR #532, opened 2025-10-28
fix(input/kinesis): Ensure pagination over all Kinesis shards

gregfurman:
## Motivation

Addresses #531. 

The Kinesis API's `ListShards` operation allows for a maximum of 1K shards to be returned at a time, before a caller is required to paginate across results. 

This PR ensures that we paginate over all possible shards of a Kinesis stream when running the `aws_kinesis` input.

## Changes
- Add helper package and corresponding functions to help with AWS results pagination when no sdk functionality exists.
- Use new `Collect` and `TokenIterator` functions to ensure we paginate over all possible results with `ListShards` operation

### DIFF
--- a/internal/impl/aws/helper/iter_test.go
+++ b/internal/impl/aws/helper/iter_test.go
@@ -92,7 +92,7 @@ func TestCollect(t *testing.T) {
 				assert.Equal(t, test.expErrorValue, err)
 			} else {
 				require.NoError(t, err)
-				assert.Equal(t, test.expResult, allFoo)
+				require.ElementsMatch(t, test.expResult, allFoo)
 			}
 		})
 	}


internal/impl/aws/helper/iter.go
```diff
@@ -0,0 +1,44 @@
...
+				return
+			}
+			prev = next
+		}
+	}
+}
+
+// Collect takes an iterator of slices and returns a concatentation of all results.
+func Collect[V any](seq iter.Seq2[[]V, error]) ([]V, error) {
+	var result []V
```
jem-davies, line 36:
```suggestion
	result := []V{}
```

This fixes the test "[TestCollectEmptyPages](https://github.com/gregfurman/bento/pull/2/files#diff-0e1c59791a4772848ae03b4f736b50cb94421237c12ed4c5bbdd787bbd71bf3bR65-R70)" 

ie. If there is nothing returned shouldn't we expect an empty slice rather than a nil value?

jem-davies (dismissed):
Written some unit tests on another branch for this - just had one test case that could do with a change but otherwise LGTM!

internal/impl/aws/helper/iter.go
gregfurman: The `nil` slice is idiomatic here IMO. This way, if results are empty we can go `pages == nil` or `len(pages) == 0` -- which also means 1 fewer assign in memory 🤷 

Have changed the test to use `ElementsMatch` so we don't run into this when testing

jem-davies (approved):

PR #536, opened 2025-10-30
feat: Add support for NSQ with authentication

danrjohnson:
Changes
---------

This adds support for publishing to and consuming from NSQ infrastructure that requires authentication. This support is optional and configured using the normal configuration patterns.

Testing
-------
* Added extra integration tests with authentication enabled that match existing nsq tests
* Because there is no "out of the box" nsq auth server, this adds some basic code for compiling and running an nsq authentication server for some true end to end testing.

### DIFF
--- a/internal/impl/nsq/output.go
+++ b/internal/impl/nsq/output.go
@@ -14,11 +14,11 @@ import (
 )
 
 const (
-	noFieldNSQDAddr  = "nsqd_tcp_address"
-	noFieldTLS       = "tls"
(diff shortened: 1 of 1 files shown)


danrjohnson:
I don't know that the tests that add auth are really necessary. They can be removed since they don't provide a ton of value beyond "Yes this config option works" :^)

jem-davies (approved):
LGTM - thanks 😄

PR #541, opened 2025-11-03
fix sql outputs db nil

jem-davies:
https://warpstreamlab-uwt2030.slack.com/archives/C0768UE3NEM/p1762109271146129

### DIFF
--- a/internal/impl/sql/integration_test.go
+++ b/internal/impl/sql/integration_test.go
@@ -1568,7 +1568,7 @@ func TestIntegrationCheckReconnectLogic(t *testing.T) {
 			"5432/tcp": {
 				{
 					HostIP:   "0.0.0.0",
-					HostPort: string(freePort),
+					HostPort: freePort,
 				},
 			},
 		},
@@ -1656,7 +1656,8 @@ output:
 	var count int
 	db, err = sql.Open("postgres", dsn)
 	require.Eventually(t, func() bool {
-		db.QueryRow("select count(*) from test_data;").Scan(&count)
+		err := db.QueryRow("select count(*) from test_data;").Scan(&count)
+		require.NoError(t, err)
 		return count > 1000
 	}, time.Minute, time.Second)
 
@@ -1678,6 +1679,7 @@ output:
 	// we should now have reconnected and have more rows in the table
 	var countAfterRestart int
 	require.Eventually(t, func() bool {
+		//nolint:errcheck // We expect errors here whilst reconnecting
 		db.QueryRow("select count(*) from test_data;").Scan(&countAfterRestart)
 		return countAfterRestart > count+100
 	}, time.Minute, time.Second)


internal/impl/sql/output_sql_insert.go
```diff
@@ -216,27 +219,46 @@ func (s *sqlInsertOutput) Connect(ctx context.Context) error {
...
+					return
+				case <-ticker.C:
+					pingCtx, cancel := context.WithTimeout(context.Background(), time.Second*5)
+					if s.driver != "trino" {
+						s.dbMut.RLock()
+						db := s.db
+						s.dbMut.RUnlock()
+						if err := db.PingContext(pingCtx); err != nil {
+							s.dbMut.Lock()
+							s.db = nil
```
gregfurman, line 244: ~This set to nil should be in a write lock~ Nvm the diff loaded weirdly on my phone. This is fine!

internal/impl/sql/output_sql_insert.go
```diff
@@ -216,27 +219,46 @@ func (s *sqlInsertOutput) Connect(ctx context.Context) error {
...
-	go func() {
-		<-s.shutSig.HardStopChan()
-
-		s.dbMut.Lock()
-		_ = s.db.Close()
-		s.dbMut.Unlock()
-
-		s.shutSig.TriggerHasStopped()
-	}()
+	s.healthCheckStarted.Do(func() {
```
gregfurman, line 222: So if we close and reconnect, then the health check won't be triggered again.

Should we not just do a DB ping on connect?

internal/impl/sql/output_sql_insert.go
```diff
@@ -216,27 +219,46 @@ func (s *sqlInsertOutput) Connect(ctx context.Context) error {
...
+		go func() {
+			ticker := time.NewTicker(time.Second * 30)
+			defer ticker.Stop()
+
+			for {
+				select {
+				case <-s.shutSig.HardStopChan():
+					s.dbMut.Lock()
```
gregfurman, line 230: This can be a read lock

internal/impl/sql/output_sql_insert.go
jem-davies: So we aren't just doing a ping on connect - because we want to trigger the Connect func if we lose connection - this could be because a token from a cloud service has expired and we want to get a new one. 

But when we call .Connect on subsequent calls after the initial, we won't want to start this goroutine again.

internal/impl/sql/output_sql_insert.go
jem-davies: Possibly though what I have done is move the goroutine that was here before into a select case, so was wanting to leave as it was.

jem-davies:
@gregfurman 

> Can you clarify where the race condition was occurring? 👀

So it was here: 

```go

func (s *sqlInsertOutput) WriteBatch(ctx context.Context, batch service.MessageBatch) error {
	s.dbMut.RLock()
	defer s.dbMut.RUnlock()

	if s.driver != "trino" {
		if err := s.db.PingContext(ctx); err != nil {
			s.db = nil // Writing to s.db, but we only have a readlock 
		}
	}
```

I spent a while trying to get a test going for this behaviour - trying some [go-sqlmock](https://pkg.go.dev/github.com/data-dog/go-sqlmock) but that is limited because you have assert a limited number of expected queries - and using it to test for a race condition perhaps not ideal. Then was struggling to dep. inject something that would have a flaky `s.db.PingContext`

internal/impl/sql/integration_test.go
```diff
@@ -1546,3 +1547,120 @@ func waitForRds(localstackPort string) (err error) {
 	}
 	return errors.New("polling localstack for ready RDS failed")
 }
+
+func TestIntegrationXxx(t *testing.T) {
```
gregfurman:
```suggestion
func TestIntegrationReconnectToDatabase(t *testing.T) {
```

Perhaps we should add an additional test-case to the integration testing suite for the case of an upstream disconnection 👀

internal/impl/sql/output_sql_insert.go
gregfurman: That makes sense. My only concern is that if we `Close()` the component, a hard-stop is triggered. This will break the loop and we wont have access to this `db.Ping()` mechanism again.

So going:

1. Connect()
2. Close()
3. Connect()

means we're without a healthcheck.

jem-davies: Do you think it would be better if we changed `healthCheckStarted sync.Once` -> `healthCheckRunning bool` then in the `Close()` func we could set this to be false, and only start that goroutine if `healthCheckRunning == false`?

internal/impl/sql/integration_test.go
jem-davies: Hey sorry was working on it when you saw this..

jem-davies:
@gregfurman managed to get a test to simulate the issue, you can see the issue if you checkout the commit that adds the test then running it btw.

jem-davies:
closed in favour of : https://github.com/warpstreamlabs/bento/pull/545

PR #555, opened 2025-11-12
chore: upgrade gcp pubsub dep -> /v2

jem-davies:
Updates gcp pubsub dep -> /v2

[Google's migration guide](https://github.com/googleapis/google-cloud-go/blob/main/pubsub/MIGRATING.md)

TODO: 

- [ ] Fix broken integration test
- [ ] Review changes to the config (i.e. should we change field `topic` -> `publisher` to mirror the API?) 

V1 is marked for end of support July 2026...


gregfurman:
@jem-davies Seems this would fix CI in https://github.com/warpstreamlabs/bento/pull/722 👀

jem-davies:
> @jem-davies Seems this would fix CI in #722 👀

@gregfurman 

Looks like so - do you think we can suppress the Lint for now - and I will revisit this?

gregfurman:
> > @jem-davies Seems this would fix CI in #722 👀
> 
> @gregfurman
> 
> Looks like so - do you think we can suppress the Lint for now - and I will revisit this?

Yeah suppose that is fine. Otherwise, we can always add a ticket to the backlog for this if we need some community assistance :octocat:

PR #556, opened 2025-11-14
extend output docs: Add inline output examples to broker documentation

aratz-lasa:
Otherwise it's not obvious how to specify an output


gregfurman:
@aratz-lasa 👋 I can't directly push to your branch because of permissions. I opened up a PR targeting this branch that will change the inline approach to instead use `Example()` for these example use-cases. See https://github.com/warpstreamlabs/bento/pull/557

If you're happy with that, we can merge it with your changes and ship the new doc changes when you're ready :shipit:

PR #558, opened 2025-11-14
feat. gcp_bigquery_write_api auto schema creation

joaolcorreia:
## Summary 

The existing output `gcp_bigquery_write_api` is a very efficient way to stream data into BigQuery. 

When we have events that have been validated, evolving the schema of the BigQuery table is a problem and will cause operational burden of having to match evolving payload to schema if done by hand.

This contribution adds two options `auto_add_missing_columns: true` and `max_schema_update_retries: 3` which enables this output to mutate the Biquery table with the necessary fields (inferring its type) and retry inserting the rows.

When using this feature it is best to let `gcp_bigquery_write_api` infer the data types from an empty table.

> [!IMPORTANT]
> In the JSON payload if you send a string today and tomorrow the same field has a integer it will failed to mutate the table! Make sure you have strong schema validation upstream of this process.

### How it works

1. Try to insert rows
2. On schema error, acquire table-specific lock (other tables can still write)
3. Invalidate stream cache for this table
4. Fetch current schema and detect missing fields
5. Update table schema with missing fields (with ETag retry logic for concurrent updates)
6. Wait for schema propagation (poll Write Stream API with timeout)
7. Invalidate stream cache again (force recreation with new schema)
8. Release table-specific lock
9. Retry the write (up to max_schema_update_retries)

### Key clarifications:

- Per-table locking, not global (other tables aren't locked)
- Stream invalidation happens twice - before update and after propagation verification
- ETag retry logic - handles concurrent schema updates from multiple processes
- Polling with timeout - verifies propagation rather than blind wait
- Automatic retry - the whole write is retried after schema update

### Common use cases

- Stream validated event data into Bigquery, don't worry about the table schema, it will evolve automatically (use responsibly).

### DIFF
--- a/website/docs/components/outputs/gcp_bigquery_write_api.md
+++ b/website/docs/components/outputs/gcp_bigquery_write_api.md
@@ -54,6 +54,8 @@ output:
       grpc: ""
     stream_type: DEFAULT
     message_format: json
+    auto_add_missing_columns: false
+    max_schema_update_retries: 3
     batching:
       count: 0
       byte_size: 0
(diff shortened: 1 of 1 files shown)


joaolcorreia:
@gregfurman I didn't touch watcher ... not sure why test is failing.

jem-davies:
> @gregfurman I didn't touch watcher ... not sure why test is failing.

Some of the tests are flaky... it succeeded after a rerun.

joaolcorreia:
This doesn't work with JSON, too many errors and inconsistent.

PR #566, opened 2025-11-20
enable the setting of additional headers & methods on CORS config

jem-davies:
https://warpstreamlab-uwt2030.slack.com/archives/C0768UE3NEM/p1763519304373079

### DIFF
--- a/internal/httpserver/cors.go
+++ b/internal/httpserver/cors.go
@@ -18,7 +18,7 @@ const (
 )
 
 var defaultHeaders = []string{"Accept", "Accept-Language", "Content-Language", "Origin"}
-var defaultMethods = []string{"GET", "HEAD", "POST", "PUT", "PATCH", "DELETE"}
+var defaultMethods = []string{http.MethodGet, http.MethodHead, http.MethodPost, http.MethodPut, http.MethodPatch, http.MethodDelete}
 
 // CORSConfig contains struct configuration for allowing CORS headers.
 type CORSConfig struct {
@@ -76,12 +76,10 @@ func CORSConfigFromParsed(pConf *docs.ParsedConfig) (conf CORSConfig, err error)
 		return
 	}
 
-	ah, err := pConf.FieldStringList(fieldCORSAllowedHeaders)
+	conf.AllowedHeaders, err = pConf.FieldStringList(fieldCORSAllowedHeaders)
 	if err != nil {
 		return
 	}
-	// conf.AllowedHeaders = buildHeadersFromConfigAndDefaults(ah)
-	conf.AllowedHeaders = ah
 
 	if conf.AllowedMethods, err = pConf.FieldStringList(fieldCORSAllowedMethods); err != nil {
 		return
@@ -89,13 +87,3 @@ func CORSConfigFromParsed(pConf *docs.ParsedConfig) (conf CORSConfig, err error)
 
 	return
 }
-
-// func buildHeadersFromConfigAndDefaults(allowed_headers []string) []string {
-// 	strippedHeaders := []string{}
-// 	for _, ah := range allowed_headers {
-// 		if !slices.Contains(defaultHeaders, ah) {
-// 			strippedHeaders = append(strippedHeaders, ah)
-// 		}
-// 	}
-// 	return append(strippedHeaders, defaultHeaders...)
-// }


internal/httpserver/cors.go
```diff
@@ -49,16 +61,41 @@ func ServerCORSFieldSpec() docs.FieldSpec {
 	return docs.FieldObject(fieldCORS, "Adds Cross-Origin Resource Sharing headers.").WithChildren(
 		docs.FieldBool(fieldCORSEnabled, "Whether to allow CORS requests.").HasDefault(false),
 		docs.FieldString(fieldCORSAllowedOrigins, "An explicit list of origins that are allowed for CORS requests.").Array().HasDefault([]any{}),
+		docs.FieldString(fieldCORSAllowedHeaders, "Appends additional headers to the list of default allowed headers: Accept, Accept-Language, Content-Language & Origin. These default headers are therefore always allowed.").Array().HasDefault([]any{}).AtVersion("1.13.0"),
+		docs.FieldString(fieldCORSAllowedMethods, "Used to explicitly set allowed methods in the Access-Control-Allow-Methods header.").Array().HasDefault(defaultMethods).AtVersion("1.13.0"),
```
gregfurman, line 64: Should we not be checking whether the specified value is contained in `defaultMethods`? Or perhaps have a check that ensures the provided http method for CORS is actually valid.

internal/httpserver/cors.go
```diff
@@ -13,19 +14,28 @@ const (
 	fieldCORS               = "cors"
 	fieldCORSEnabled        = "enabled"
 	fieldCORSAllowedOrigins = "allowed_origins"
+	fieldCORSAllowedHeaders = "allowed_headers"
+	fieldCORSAllowedMethods = "allowed_methods"
 )
 
+var defaultHeaders = []string{"Accept", "Accept-Language", "Content-Language", "Origin"}
```
gregfurman, line 20: nit: Can we convert these to functions instead? i.e 

```golang
func getDefaultHeaders() []string { return []string{...} }
```  

Perhaps then we could consolidate the `buildHeadersFromConfig...` function to be replaced by

```golang
func getDefaultHeaders(extraHeaders ...string) []string { ... }
```

internal/httpserver/cors.go
```diff
@@ -13,19 +14,28 @@ const (
...
+var defaultMethods = []string{"GET", "HEAD", "POST", "PUT", "PATCH", "DELETE"}
```
gregfurman: Perhaps we use the constants that `http` provides:

```suggestion
var defaultMethods = []string{http.MethodGet, http.MethodHead, http.MethodPost, http.MethodPut, http.MethodPatch, http.MethodDelete}
```

Also, there are a couple of more methods i.e `CONNECT`, `OPTIONS`, and `TRACE` that could be worth adding?

internal/httpserver/cors.go
```diff
@@ -49,16 +61,41 @@ func ServerCORSFieldSpec() docs.FieldSpec {
...
+	if conf.AllowedMethods, err = pConf.FieldStringList(fieldCORSAllowedMethods); err != nil {
+		return
+	}
+
 	return
 }
+
+func buildHeadersFromConfigAndDefaults(allowed_headers []string) []string {
+	strippedHeaders := []string{}
+	for _, ah := range allowed_headers {
```
gregfurman: 🧹  Remove the snake case

```suggestion
func buildHeadersFromConfigAndDefaults(allowedHeaders []string) []string {
	strippedHeaders := []string{}
	for _, ah := range allowedHeaders {
```

internal/httpserver/cors.go
gregfurman: Also, think it's worth mentioning that if this is overidden, certain methods will _not_ be allowed. Like, if I want to add in a `CONNECT` method, I would have to re-add in all of the previous methods as well to ensure all is supported.

internal/httpserver/cors.go
```diff
@@ -49,16 +61,41 @@ func ServerCORSFieldSpec() docs.FieldSpec {
 	return docs.FieldObject(fieldCORS, "Adds Cross-Origin Resource Sharing headers.").WithChildren(
 		docs.FieldBool(fieldCORSEnabled, "Whether to allow CORS requests.").HasDefault(false),
 		docs.FieldString(fieldCORSAllowedOrigins, "An explicit list of origins that are allowed for CORS requests.").Array().HasDefault([]any{}),
+		docs.FieldString(fieldCORSAllowedHeaders, "Appends additional headers to the list of default allowed headers: Accept, Accept-Language, Content-Language & Origin. These default headers are therefore always allowed.").Array().HasDefault([]any{}).AtVersion("1.13.0"),
```
gregfurman, line 63: nit: Why do we not add the defaults in here at `HasDefault` instead of explicitly mentioning them in the docs?

Like, we could specify the defaults `[Accept, Accept-Language, Content-Language, Origin]` are always added and any additional headers will just be appended to the list. Think it's nicer way to document what will be used instead of explicitly mentioning it in description. Wdyt?

internal/httpserver/cors.go
```diff
@@ -49,16 +61,41 @@ func ServerCORSFieldSpec() docs.FieldSpec {
...
+	conf.AllowedHeaders = buildHeadersFromConfigAndDefaults(ah)
+
+	if conf.AllowedMethods, err = pConf.FieldStringList(fieldCORSAllowedMethods); err != nil {
+		return
+	}
+
 	return
 }
+
+func buildHeadersFromConfigAndDefaults(allowed_headers []string) []string {
```
gregfurman: I gave the look at the `handlers.AllowedMethods` and it seems there is already header deduplication that takes place, with header values being converted into their canonical form on build. See below:

```golang
func AllowedHeaders(headers []string) CORSOption {
	return func(ch *cors) error {
		for _, v := range headers {
			normalizedHeader := http.CanonicalHeaderKey(strings.TrimSpace(v))
			if normalizedHeader == "" {
				continue
			}

			if !ch.isMatch(normalizedHeader, ch.allowedHeaders) {
				ch.allowedHeaders = append(ch.allowedHeaders, normalizedHeader)
			}
		}

		return nil
	}
}
```

Makes me think we could actually just pass in whatever additonal headers were provided appended with the default headers we define above instead of filtering out like this.

internal/httpserver/cors.go
jem-davies: Getting rid of the function so not sure how relevant this is☝️

internal/httpserver/cors.go
jem-davies: Have altered to the reference but want to leave the list as-is because it's based on the defaults that were before.

internal/httpserver/cors.go
jem-davies: deleted func anyway

internal/httpserver/cors.go
jem-davies: >  Like, if I want to add in a CONNECT method, I would have to re-add in all of the previous methods as well to ensure all is supported

Yeah but that is just how default values work in Bento isn't it? Like with [allowed_verbs](https://warpstreamlabs.github.io/bento/docs/components/inputs/http_server#allowed_verbs) - you would have to add POST to any other that you wanted in the list manually.

internal/httpserver/cors.go
jem-davies: I think I prefer it the way it is.

internal/httpserver/cors.go
jem-davies: > Should we not be checking whether the specified value is contained in `defaultMethods`? 

Why tho? - Prior to this PR you were just getting the default without the option to specify in the config, Now you could specify any list of strings here, if that is a subset or superset of the defaultMethods, it is still valid?

> Or perhaps have a check that ensures the provided http method for CORS is actually valid.

What you mean? - You can have custom methods I think so would be difficult to check?

gregfurman: Since Go's http library defines a set of constant values for supported methods, am unsure of behavior if we specify methods outside those. 

Not a huge deal I guess but thought we could add some validations while we're changing this code.

gregfurman (approved):

PR #570, opened 2025-11-24
feat(bloblang): Improve Bloblang playground with Go/WASM migration

iamramtin:
## Summary

Improves the Bloblang playground by migrating core logic from JavaScript to Go/WASM.

The playground previously duplicated some of Bloblang logic in JavaScript, creating maintenance overhead and potential inconsistencies. By moving to Go/WASM, we:
- Better align it with the rest of the codebase 
- Reuse existing Bloblang environment and validation logic directly
- Eliminate duplication between Go and JS implementations
- Improve type safety and error handling
- Maintain a closer source of truth for Bloblang behavior

### Changes

**Go/WASM Migration:**
- Migrated execution, validation, formatting, autocompletion, and other utilities from JS to Go/WASM
- Split blobl package into focused modules (core, editor, server, wasm, types)
- Simplified WASM initialization and auto-register server/WASM handlers
- Added elementary test suite for core playground functions (verifies playground-specific behavior without duplicating existing Bloblang tests)

**JavaScript Simplification:**
- Extracted BloblangAPI abstraction layer for WASM/server routing
- Removed unused code, dead methods, redundant functionality
- Improved error handling and WASM initialization flow

**Other Improvements:**
- Improved autocomplete with pre-rendered HTML docs and smarter context detection
- Replaced hardcoded values with CSS variables and duplicate rules
- Improved mobile responsive layout and overall UX

#### 1. Run Unit Tests
```bash
make playground-test
```

#### 2. Test Server Mode (Go backend via HTTP)

```
# Build and run playground server
go build -o bento ./cmd/bento/main.go
make playground
./bento blobl playground -p 3001 --no-open
```

Visit [playground](http://localhost:3001) and verify:
- Playground loads without errors
- Execute mappings (try: root.name = this.name.uppercase())
- Format mappings (Cmd/Ctrl+Shift+F or Format button)
- Autocomplete works (type root =  and trigger autocomplete)
- Error messages display correctly for invalid input/mappings
- Theme toggle works (light/dark mode)

#### 3. Test WASM Mode (Client-side execution)

```
# Clean and rebuild
rm -rf website/static/playground
make docs
make playground
cd website && yarn start
```

Visit the [playground](http://localhost:3000/bento/docs/guides/bloblang/playground) in the docs and verify:
- Playground loads in iframe without errors
- All functionality from server mode works (execution, formatting, autocomplete)
- Theme syncs with Docusaurus theme
- Responsive layout works on mobile viewport (resize browser)
- Full-screen mode works (click "Open in Full Screen" button)

### Breaking Changes

None

### DIFF
--- a/internal/cli/blobl/editor.go
+++ b/internal/cli/blobl/editor.go
@@ -196,6 +196,11 @@ func createSpecDocHTML(name string, spec query.BaseSpec, kind SpecKind) string {
 		fmt.Fprintf(&b, `<div class="ace-doc-version">Since: v%s</div>`, spec.Version)
 	}
 
+	if spec.Description != "" {
+		desc := processMarkdownDescription(spec.Description)
+		fmt.Fprintf(&b, `<div class="ace-doc-description">%s</div>`, desc)
+	}
+
 	writeParamsSection(&b, spec)
 
 	b.WriteString(`</div>`)


internal/cli/blobl/core.go
```diff
@@ -1,8 +1,11 @@
+// Package blobl provides core functionality for the Bloblang language playground,
+// including code execution, syntax highlighting, autocompletion, and formatting.
```
gregfurman: This package is not exported, so I think it's safe to remove the doc comments.

```suggestion
```

internal/cli/blobl/core.go
```diff
@@ -14,18 +17,39 @@ import (
 	"github.com/warpstreamlabs/bento/internal/value"
 )
 
-// executionResult represents the result of executing a Bloblang mapping.
-type executionResult struct {
-	Result       any `json:"result"`
-	ParseError   any `json:"parse_error"`
-	MappingError any `json:"mapping_error"`
-}
+// Configuration constants
```
gregfurman:
```suggestion
```

internal/cli/blobl/core.go
```diff
@@ -194,41 +189,145 @@ func evaluateMapping(env *bloblang.Environment, input, mapping string) *executio
...
 	}
 
-	return syntax, nil
+	// Get cached syntax data
+	syntaxData, err := getOrGenerateSyntax(env)
+	if err != nil {
+		return AutocompletionResponse{
+			Completions: []CompletionItem{},
+			Success:     false,
+			Error:       fmt.Sprintf("Failed to get syntax data: %v", err),
```
gregfurman:
```suggestion
			Error:       fmt.Sprintf("Failed to get syntax data: %s", err),
```

Makefile
```diff
@@ -128,3 +128,6 @@ playground:
 		cp "$$(go env GOROOT)/misc/wasm/wasm_exec.js" "$(WEBSITE_DIR)/static/playground/js/wasm_exec.js"; \
 	fi
 	@GOOS=js GOARCH=wasm go build -o "$(WEBSITE_DIR)/static/playground/playground.wasm" "cmd/tools/playground/main.go"
+
+playground-test:
+	@go test ./internal/cli/blobl -timeout 3m
```
gregfurman: Wont this just run as a part of the normal test suite? Why do we need to add an extra make command?

iamramtin:
Hey @gregfurman @jem-davies just a friendly ping to see if we can get this merged in?

gregfurman:
@iamramtin hey! Thanks for the ping and sorry for delay, will give this a priority review 👍

internal/cli/blobl/wasm.go
```diff
@@ -30,35 +49,127 @@ func ExecuteBloblangMapping() js.Func {
...
-// If marshaling fails, returns a JS string with the error message.
+// exportAutocomplete returns a js.Func that provides autocompletion suggestions for Bloblang code.
+//
+// Arguments:
+//   - args[0]: JSON string containing AutocompletionRequest with line, column, prefix, beforeCursor
+//
+// Returns a JS object with:
+//   - "completions": array of completion items with caption, value, meta, type, score, docHTML
+//   - "success":     true if autocompletion succeeded, false otherwise
+//   - "error":       error message if autocompletion failed, else nil
```
gregfurman: I don't think this level of verbosity is necessary

```suggestion
```

internal/cli/blobl/wasm.go
```diff
@@ -30,35 +49,127 @@ func ExecuteBloblangMapping() js.Func {
...
+
+// exportFormat returns a js.Func that formats Bloblang mappings with proper indentation.
+//
+// Arguments:
+//   - args[0]: Bloblang mapping string to format
+//
+// Returns a JS object with:
+//   - "formatted": the formatted mapping string
+//   - "success":   true if formatting succeeded, false otherwise
+//   - "error":     error message if formatting failed, else nil
```
gregfurman: This can be removed as well:

```suggestion
```

internal/cli/blobl/wasm.go
```diff
@@ -30,35 +49,127 @@ func ExecuteBloblangMapping() js.Func {
...
+func exportFormat() js.Func {
```
gregfurman: I think this JS function pattern should be re-considered if this level of documentation is required to understand what's going on. 

If we decide to go with the global env approach, what about something instead like:

```go
func format(_ js.Value, args []js.Value) any {
   if len(args) != 1 || args[0].Type() != js.TypeString {
      // ...
   }
   // ...
   return toJS(response)
}
```

or better yet, we could pass in the bloblang env as a param when constructing the function:

```go
func formatHandler(env *bloblang.Environment) js.Func {
   // ...
}
```

internal/cli/blobl/wasm.go
```diff
@@ -4,13 +4,32 @@ package blobl
...
 	"github.com/warpstreamlabs/bento/internal/bloblang"
 )
 
-// ExecuteBloblangMapping returns a js.Func that executes the Bloblang mapping functionality on input JSON.
-// This function is intended to be exposed to JavaScript via WebAssembly.
+// getWasmEnvironment returns a Bloblang environment suitable for browser/WASM context.
+// Excludes functions that depend on system resources (env vars, filesystem).
+var getWasmEnvironment = sync.OnceValue(func() *bloblang.Environment {
+	return bloblang.GlobalEnvironment().WithoutFunctions("env", "file")
+})
```
gregfurman: Already mentioned, but why not we replace this with an `init` and run the `InitializeWASM` logic as well as all JS function registration when the package gets initialised i.e

```go
func init() {
   env := bloblang.GlobalEnvironment().WithoutFunctions("env", "file")
   InitializeWASM()
   // .. pass 'env' around
}
```

Then we can import

internal/cli/blobl/types.go
```diff
@@ -0,0 +1,134 @@
...
+type completionCache struct {
+	syntax *bloblangSyntax
+	env    *bloblang.Environment
+}
+
+// FormatMappingResponse represents the standardized response for mapping formatting
+type FormatMappingResponse struct {
+	Success   bool   `json:"success"`
+	Formatted string `json:"formatted"`
+	Error     string `json:"error,omitempty"`
```
gregfurman: I think we can use the globally available `Error` function from the JS environment i.e

```go
var errorResp FormatMappingResponse
err := js.Global().Get("Error").New(errorResp.Error)
panic(err) // or maybe ignore if error
```

Then we can treat the error case as a JS-level error and handle with a try-catch instead of this error payload approach.

Also, that way we don't need to pass around an `error` attribute and can focus on payload responses.

wdyt?

internal/cli/blobl/types.go
```diff
@@ -0,0 +1,134 @@
...
+type Spec interface {
+	GetDescription() string
+	GetStatus() query.Status
+	GetVersion() string
+	GetParams() query.Params
+}
+
+// FunctionSpecWrapper wraps query.FunctionSpec to implement Spec interface
+type FunctionSpecWrapper struct {
+	query.FunctionSpec
```
gregfurman: I don't think the entire struct needs to be embedded here since you're just using some attributes. Consider

```suggestion
	spec query.FunctionSpec
```

and then 

```go
func (f FunctionSpecWrapper) GetDescription() string { return f.spec.Description }
```

internal/cli/blobl/types.go
```diff
@@ -0,0 +1,134 @@
...
+
+// AutocompletionResponse represents the response with completion suggestions
+type AutocompletionResponse struct {
+	Completions []CompletionItem `json:"completions"`
+	Success     bool             `json:"success"`
+	Error       string           `json:"error,omitempty"`
+}
+
+// Spec represents a unified interface for function and method specs
+type Spec interface {
```
gregfurman: Why not use a dedicated `Spec` struct here instead with the attributes we need? Feels like 2 more structs + an interface is a bit of overkill.

internal/cli/blobl/wasm/wasm.go
gregfurman, line 1: Wdyt of placing all of these WASM playground related functions in their own package i.e `cli/blobl/wasm` or `/playground`?

gregfurman (commented):
I'll do another pass but have left some comments on structure and the pattern(s) used. Lmk thoughts!

internal/cli/blobl/core.go
iamramtin: Why should we use `%s` here for error types?

gregfurman: Ahh sorry. Meant to say `err.Error()`, since that would extract the string form of the error:

```suggestion
			Error:       fmt.Sprintf("Failed to get syntax data: %s", err.Error()),
```

internal/cli/blobl/types.go
iamramtin: Good call, I think this explicitness is much better 👍

internal/cli/blobl/types.go
iamramtin: How would you feel about me adding a `BaseSpec` to internal/bloblang/query/docs.go and having `FunctionSpec` and `MethodSpec` embed those shared attributes instead of duplicating all this logic in this file?

iamramtin: See commit [407ffa62b](https://github.com/warpstreamlabs/bento/pull/570/changes/407ffa62b2f24127e673be7880a495d812ae86c7) and LMK what you think, this is non-breaking and fully compatible and should let me not have to redeclare these shared attributes

iamramtin: And follow up [1da4e310a](https://github.com/warpstreamlabs/bento/pull/570/changes/1da4e310a511ce8b48454f0ea56a7ec9cc65cbb4) to see how it affects the current code

internal/cli/blobl/core.go
iamramtin: Are these not effectively the same thing when the error is not nil?

internal/cli/blobl/types.go
iamramtin: See [24c6fa92f2a9](https://github.com/warpstreamlabs/bento/pull/570/changes/24c6fa92f2a9509ba51b195d6881b654355cd128), lmk wyt!

internal/cli/blobl/wasm/wasm.go
iamramtin: See [2ee4747ac](https://github.com/warpstreamlabs/bento/pull/570/changes/2ee4747ac53c36ec0cf1b1d539ddfbc7e0dfa654)

iamramtin:
@gregfurman addressed comments, LMK WYT 🏁

internal/cli/blobl/core_test.go
```diff
@@ -0,0 +1,431 @@
...
+
+			hasResult := result.Result != nil
+			if hasResult != tt.expectResult {
+				t.Errorf("ExecuteBloblangMapping() hasResult = %v, expectResult = %v", hasResult, tt.expectResult)
+			}
+		})
+	}
+}
+
+// Test validate (Bloblang validation)
```
gregfurman:
```suggestion
```

internal/cli/blobl/core_test.go
```diff
@@ -0,0 +1,431 @@
...
+		},
+		{
+			name:    "empty mapping",
+			mapping: "",
+			valid:   false,
+		},
+		{
+			name:    "invalid syntax",
+			mapping: `root.bad =`,
+			valid:   false,
```
gregfurman: `valid` is implied to be `false` if empty, so it can just be omitted

internal/cli/blobl/core_test.go
```diff
@@ -0,0 +1,431 @@
...
+		},
+		{
+			name:    "unknown function",
+			mapping: `root = fake_function()`,
+			valid:   false,
+		},
```
gregfurman, line 182: Perhaps add some more negative test cases i.e 
```suggestion
		},
   	{
			name:    "invalid formatting",
			mapping: `root=fake_function()`,
			valid:   false, // should this be false?
		},
   	{
			name:    "invalid root name",
			mapping: `ROOT = this`,
			valid:   false,
		},
```

internal/cli/blobl/core_test.go
```diff
@@ -0,0 +1,431 @@
...
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			valid, err := ValidateBloblangMapping(env, tt.mapping)
```
gregfurman: Since `valid` will always be `false` on `error` and `true` when no error is present, I can't really see a need for checking (and returning) both values.

internal/cli/blobl/core.go
```diff
@@ -194,41 +174,105 @@ func evaluateMapping(env *bloblang.Environment, input, mapping string) *executio
...
+		return false, errors.New("mapping cannot be empty")
+	}
+
+	_, err := env.NewMapping(mapping)
+	if err != nil {
+		return false, fmt.Errorf("could not parse mapping: %v", err)
+	}
+
+	return true, nil
+}
```
gregfurman: Is this extra utility function necessary? Can we not just use the result of the underlying `NewMapping`?

```golang
_, err := env.NewMapping(mapping)
if err != nil {
   // ...
}
```

internal/cli/blobl/core_test.go
```diff
@@ -0,0 +1,431 @@
...
+			if !tt.valid && err == nil {
+				t.Errorf(
+					"ValidateBloblangMapping() expected error, got nil",
+				)
+			}
+		})
+	}
+}
+
+// Test syntax (Syntax metadata generation)
```
gregfurman:
```suggestion
```

internal/cli/blobl/core_test.go
```diff
@@ -0,0 +1,431 @@
...
+	}
+
+	// Check highlighting rules generated
+	if len(syntax.Rules) == 0 {
+		t.Error("GenerateBloblangSyntax() returned no syntax rules")
+	}
+
+	// Check that DocHTML is pre-generated for functions
+	foundDocHTML := false
+	for _, fn := range syntax.Functions {
```
gregfurman: Do we expect this `DocHTML` to be non-empty for all functions?

internal/cli/blobl/core_test.go
```diff
@@ -0,0 +1,431 @@
...
+			}
+		})
+	}
+}
+
+// Test syntax (Syntax metadata generation)
+func TestGenerateBloblangSyntax(t *testing.T) {
+	env := bloblang.GlobalEnvironment()
+
+	syntax, err := GenerateBloblangSyntax(env)
```
gregfurman: Would it not be easier to register a dummy function and method to a  new bloblang enironment, and then assert that it's `DocHTML` is equal to some pre-expected value? 

Think that'd be a nice way of testing this GenerateBloblangSyntax functionality in a more maintainable + verifiable way.

internal/cli/blobl/core_test.go
```diff
@@ -0,0 +1,431 @@
...
+					"formatted output mismatch\nexpected:\n%s\n\ngot:\n%s",
+					tt.expected,
+					formatted,
+				)
+			}
+		})
+	}
+}
+
+// Test autocomplete (Autocompletion)
```
gregfurman:
```suggestion
```

internal/cli/blobl/editor.go
```diff
@@ -0,0 +1,591 @@
...
+		{"import", "Import external mapping", 800},
+		{"meta", "Access message metadata", 820},
+		{"deleted", "Delete the current field", 750},
+	}
+
+	var completions []CompletionItem
+
+	for _, keyword := range keywords {
+		docHTML := fmt.Sprintf(`
+			<div class="ace-doc">
```
gregfurman: Wont this contain all this extra tab spacing?

```suggestion
<div class="ace-doc">
```

Perhaps place this into a `const` like


```golang
const itemHTMLFormat = `
<div class="ace-doc">
<div class="ace-doc-header">
`
```

Then when you use it you can just remove the first `\n` character:

```golang
docHTML := fmt.Sprintf(itemHTMLFormat[1:], ...)
```

internal/cli/blobl/editor.go
```diff
@@ -0,0 +1,591 @@
...
+				true, // isMethod
+			))
+		}
+	}
+
+	return completions
+}
+
+// buildCompletionItem creates a CompletionItem from a spec wrapper.
+func buildCompletionItem(name string, spec query.BaseSpec, category, itemType string, isMethod bool) CompletionItem {
```
gregfurman: nit: `itemType` is always `"method"` or `"function"` so why do we need to also check `isMethod`?

internal/cli/blobl/editor.go
```diff
@@ -0,0 +1,591 @@
...
+	case map[string]functionSpecWithHTML:
+		for name, spec := range s {
+			completions = append(completions, buildCompletionItem(
+				name,
+				spec.BaseSpec,
+				spec.Category,
+				"function",
+				false, // isMethod
+			))
+		}
```
gregfurman: Perhaps remove the `buildCompletionItem` in favour of inlining this:

```suggestion
	case map[string]functionSpecWithHTML:
		for name, spec := range s {
			category := "general"
			if spec.Category != "" {
				category = spec.Category
			}

			item := CompletionItem{
				Caption:     name,
				Meta:        category,
				Type:        "function",
				Score:       getSpecScore(spec.Status),
				Description: spec.Description,
				DocHTML:     createSpecDocHTML(name, spec.BaseSpec, false),
			}
			completions = append(completions, item)
		}
```

internal/cli/blobl/editor.go
```diff
@@ -0,0 +1,591 @@
...
+			Description: keyword.description,
+			DocHTML:     docHTML,
+		})
+	}
+
+	return completions
+}
+
+// getCompletions converts function/method specs with HTML into autocompletion items.
+func getCompletions(specs any) []CompletionItem {
```
gregfurman: suggestion(non-blocking): Why not just have two functions, one for `Method` and one for `Function`?

internal/cli/blobl/editor.go
```diff
@@ -0,0 +1,591 @@
...
+	switch status {
+	case query.StatusStable:
+		return 1000
+	case query.StatusBeta:
+		return 800
+	case query.StatusExperimental:
+		return 600
+	}
+	return 500
+}
```
gregfurman: I like this priority idea. I just think the magic number approach is not very maintainable.

Why not rather define some constant enums for this?


```suggestion
type statusPriority int

const (
	Unknown statusPriority = iota // this will be 0
	Experimental
	Beta
	Stable // this will be 3
)

func getSpecScore(status query.Status) statusPriority {
	// Prioritize by status (stable > beta > experimental)
	switch status {
	case query.StatusStable:
		return Stable
	case query.StatusBeta:
		return Beta
	case query.StatusExperimental:
		return Experimental
	}
	return Unknown
}
```

internal/cli/blobl/editor.go
```diff
@@ -0,0 +1,591 @@
...
+	html.WriteString(fmt.Sprintf(`<div class="ace-doc" data-docs-url="%s" data-function-name="%s" data-is-method="%t">`, docUrl, name, isMethod))
+
+	// Header with signature and status
+	html.WriteString(fmt.Sprintf(`
+		<div class="ace-doc-header clickable-header" title="Click to view documentation">
+			<div class="ace-doc-signature">
+				<strong>%s</strong>
+				<span class="ace-status-%s">%s</span>
+			</div>
+		</div>`, signature, status, status))
```
gregfurman: nit: FYI you can use the some string formatting magic here to remove the repeated reference


```suggestion
				<span class="ace-status-%[2]s">%[2]s</span>
			</div>
		</div>`, signature, status))
```

internal/cli/blobl/editor.go
```diff
@@ -0,0 +1,591 @@
...
+	}
+	return 500
+}
+
+func hasSpecParameters(spec query.BaseSpec) bool {
+	return len(spec.Params.Definitions) > 0 || spec.Params.Variadic
+}
+
+// createSpecDocHTML creates HTML documentation for Bloblang functions and methods
+func createSpecDocHTML(name string, spec query.BaseSpec, isMethod bool) string {
```
gregfurman: Thoughts on using HTML templating here instead? We can create a template with all of this and use go templating magic to sub in all these values when the template is rendered.

See https://pkg.go.dev/html/template

Currently think this approach is a bit tough to understand and could potentially be a bit brittle

internal/cli/blobl/editor.go
```diff
@@ -0,0 +1,591 @@
...
+	case value.TArray:
+		return "array"
+	case value.TObject:
+		return "object"
+	case value.TTimestamp:
+		return "timestamp"
+	default:
+		return "any"
+	}
+}
```
gregfurman, line 282: Is this necessary?

The underlying `value.Type` is just an alias to a `string` so running `string(param.ValueType)` would convert the type into it's string equivalent


```suggestion
func getParamTypeString(param query.ParamDefinition) string {
   return string(param.ValueType)
}
```

internal/cli/blobl/editor.go
```diff
@@ -0,0 +1,591 @@
...
+// generateDocumentationLink creates the documentation URL
+func generateDocumentationLink(name string, isMethod bool) string {
+	if isMethod {
+		return fmt.Sprintf("%s/methods#%s", docsBaseURL, name)
+	}
+	return fmt.Sprintf("%s/functions#%s", docsBaseURL, name)
+}
+
+// formatBloblang formats Bloblang code with indentation and consistent spacing
+func formatBloblang(originalMapping string) (string, error) {
```
gregfurman: ok wow I can't pretend to say I understand much of what is happening here re: all of this parsing. How did we do it previously in the playground?

internal/cli/blobl/server.go
```diff
@@ -57,225 +39,284 @@ func init() {
...
+		http.Error(w, "Invalid request body", http.StatusBadRequest)
+		return
+	}
 
-	if !c.Bool("no-open") {
-		u, err := url.Parse("http://localhost:" + port)
-		if err != nil {
-			return fmt.Errorf("failed to parse URL: %w", err)
-		}
-		openBrowserAt(u.String())
```
gregfurman, line 258: Why the refactor here?

internal/cli/blobl/server.go
gregfurman, line 1: Would it be possible to revert this refactor? I'm finding it a bit tough to see what was added versus what was refactored

internal/bloblang/query/docs.go
```diff
@@ -54,6 +54,33 @@ var (
 	StatusHidden       Status = "hidden"
 )
 
+// ------------------------------------------------------------------------------
+
+// BaseSpec contains fields shared by function and method specifications.
+type BaseSpec struct {
```
gregfurman, line 60: Thoughts on just having a custom struct inside the `blobl` package that maps `FunctionSpec` or `MethodSpec` to a common single type? That way, the `internal/bloblang` functionality does not need to be touched at all and we can keep this unified approach logic where its needed

internal/cli/blobl/core_test.go
iamramtin: `ROOT = this` would be valid because it would be assigning `this` to a variable called `ROOT` rather than Bloblang special `root` key, i.e.

```
{
  "ROOT": {
    "name": "bento",
    "stars": 1500
  }
}
```


instead of 

```
{
  "name": "bento",
  "stars": 1500
}
```

internal/cli/blobl/server.go
iamramtin: I understand the difficulty in reviewing this as a mixed change, but  I'd prefer not to revert as most of the diff is structural rather than behavioural

The refactor separates server lifecycle, routing, and handlers into explicit types and methods, removes large inline closures, and centralises the state into a server type. This was necessary to make this file more idiomatic when adding the new endpoints cleanly without further growing what was previously essentially just a single function

Functionally, behaviuor is the same,  main change is turning implicit structure into explicit structure so future additions don't compound the layout

I’m happy to walk through the structure or split follow-ups if that helps

PR #575, opened 2025-11-26
feat(buffer_memory): spillover and active metrics

### LINKED ISSUE #574: Memory buffer metrics for tracking spillover
It's currently difficult to monitor if events are spilling over when using the `memory` buffer; it can be somewhat accomplished using `input_received - buffer_received`, but this isn't easily discoverable and isn't robust. A dedicated `buffer_spillover` metric would be useful.

alecmerdler:
Adds two new metrics to the `memory` buffer which track the active number of bytes in the buffer and number of bytes that have been spilled due to reaching maximum capacity.

Resolves #574

### DIFF
--- a/internal/impl/pure/buffer_memory.go
+++ b/internal/impl/pure/buffer_memory.go
@@ -44,13 +44,21 @@ This buffer intentionally weakens the delivery guarantees of the pipeline and th
 
 ## Batching
 
-It is possible to batch up messages sent from this buffer using a [batch policy](/docs/configuration/batching#batch-policy).`).
+It is possible to batch up messages sent from this buffer using a [batch policy](/docs/configuration/batching#batch-policy).
+
+## Metrics 
+
+- ` + "`buffer_active`" + ` Gauge metric tracking the current number of bytes in the buffer.
+- ` + "`buffer_spillover`" + ` Counter metric tracking the total number of bytes dropped because of spillover.
+
+`).
 		Field(service.NewIntField("limit").
 			Description(`The maximum buffer size (in bytes) to allow before applying backpressure upstream.`).
 			Default(524288000)).
 		Field(service.NewBoolField("spillover").
 			Description("Whether to drop incoming messages that will exceed the buffer limit.").
 			Advanced().
+			Version("1.13.0").
 			Default(false)).
 		Field(service.NewInternalField(bs))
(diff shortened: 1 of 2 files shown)


internal/impl/pure/buffer_memory.go
```diff
@@ -262,6 +268,7 @@ func (m *memoryBuffer) WriteBatch(ctx context.Context, msgBatch service.MessageB
 
 	for (m.bytes + extraBytes) > m.cap {
 		if m.spilloverEnabled {
+			m.spilloverBytes.Incr(int64(extraBytes))
```
gregfurman, line 279: nit: This `Incr` definitely looks like it'd benefit from generics given this unnecessary conversion 🤔

internal/impl/pure/buffer_memory.go
```diff
@@ -118,14 +118,19 @@ type memoryBuffer struct {
...
 
-func newMemoryBuffer(capacity int, spilloverEnabled bool, batcher *service.Batcher) *memoryBuffer {
+func newMemoryBuffer(capacity int, spilloverEnabled bool, batcher *service.Batcher, res *service.Resources) *memoryBuffer {
 	return &memoryBuffer{
 		cap:              capacity,
 		spilloverEnabled: spilloverEnabled,
 		cond:             sync.NewCond(&sync.Mutex{}),
 		batcher:          batcher,
+		activeBytes:      res.Metrics().NewGauge("buffer_active"),
+		spilloverBytes:   res.Metrics().NewCounter("buffer_spillover"),
```
gregfurman, line 141: Can you add a note in the docs that these metrics will now be tracked?

gregfurman (commented):
Looks good! I think we should just mention in the docs that the `active_bytes` and `buffer_spillover` are now going to be tracked. Once done, happy to approve and ship :shipit:

internal/impl/pure/buffer_memory.go
jem-davies: @alecmerdler - was going to add but we don't have access so raised a PR: https://github.com/alecmerdler/bento/pull/1

alecmerdler: Thanks! I merged it and pulled in the changes to this branch now.

PR #577, opened 2025-11-27
chore/fix(sql): remove lib/pq in favor of pgx

maxtheaxe:
Draws on [this issue](https://github.com/redpanda-data/connect/issues/1987) (which I ran into myself), in which `sql_insert` doesn't actually support Postgres array columns. I was left trying to manually escape convoluted strings using regex, and handcraft array literals, which is really brittle.

This is a different approach to [the _other_ PR](https://github.com/redpanda-data/connect/pull/3674), in which I add **pgx** as an alternate driver. In this case, as discussed [here](https://warpstreamlab-uwt2030.slack.com/archives/C0768UE3NEM/p1760534919589189), I'm replacing **lib/pq** directly, and adding a shim that retains support for referencing the new postgres driver with `postgres`. Given that `pgx/v4` was already used here, I stuck with that, though obviously you may choose to upgrade to `v5`.

It doesn't redefine any settings otherwise, just reuses everything from the original postgres driver.

---

I've run `go test` on `internal/impl/sql` and built a docker image, and everything seems to be fully working, at least for my somewhat narrow use case (that in the issue referenced above).

That being said, I would like to state for the record that this was LLM-assisted, though obviously the change is pretty straightforward. I figure it doesn't hurt to mention in case I've missed something obvious.

### DIFF
--- a/internal/impl/sql/conn_fields.go
+++ b/internal/impl/sql/conn_fields.go
@@ -163,6 +163,26 @@ CREATE TABLE IF NOT EXISTS some_table (
 
 }
 
+var warnPostgresV2Once sync.Once
+
+func warnIfPostgresV2(driver string, log *service.Logger) {
+	if driver != "postgres_v2" || log == nil {
+		return
+	}
+	warnPostgresV2Once.Do(func() {
+		log.With("driver", driver).Warn("driver 'postgres_v2' (jackc/pgx) is temporary and will become default 'postgres' in next minor release")
+	})
+}
+
+func getDriver(conf *service.ParsedConfig, log *service.Logger) (string, error) {
+	driver, err := conf.FieldString("driver")
+	if err != nil {
+		return "", err
+	}
+	warnIfPostgresV2(driver, log)
+	return driver, nil
+}
+
 func rawQueryField() *service.ConfigField {
 	return service.NewStringField("query").
 		Description("The query to execute. The style of placeholder to use depends on the driver, some drivers require question marks (`?`) whereas others expect incrementing dollar signs (`$1`, `$2`, and so on) or colons (`:1`, `:2` and so on). The style to use is outlined in this table:" + `
--- a/internal/impl/sql/cache_sql.go
+++ b/internal/impl/sql/cache_sql.go
@@ -117,7 +117,7 @@ func newSQLCacheFromConfig(conf *service.ParsedConfig, mgr *service.Resources) (
 
 	var err error
 
-	if s.driver, err = conf.FieldString("driver"); err != nil {
+	if s.driver, err = getDriver(conf, s.logger); err != nil {
 		return nil, err
 	}
 
--- a/internal/impl/sql/input_sql_raw.go
+++ b/internal/impl/sql/input_sql_raw.go
@@ -101,7 +101,7 @@ func newSQLRawInputFromConfig(conf *service.ParsedConfig, mgr *service.Resources
 
 	var err error
 
-	if s.driver, err = conf.FieldString("driver"); err != nil {
+	if s.driver, err = getDriver(conf, s.logger); err != nil {
 		return nil, err
 	}
 
--- a/internal/impl/sql/input_sql_select.go
+++ b/internal/impl/sql/input_sql_select.go
@@ -122,7 +122,7 @@ func newSQLSelectInputFromConfig(conf *service.ParsedConfig, mgr *service.Resour
 
 	var err error
 
-	if s.driver, err = conf.FieldString("driver"); err != nil {
+	if s.driver, err = getDriver(conf, s.logger); err != nil {
 		return nil, err
 	}
 
--- a/internal/impl/sql/output_sql_deprecated.go
+++ b/internal/impl/sql/output_sql_deprecated.go
@@ -56,7 +56,7 @@ func init() {
 //------------------------------------------------------------------------------
 
 func newSQLDeprecatedOutputFromConfig(conf *service.ParsedConfig, mgr *service.Resources) (*sqlRawOutput, error) {
-	driverStr, err := conf.FieldString("driver")
+	driverStr, err := getDriver(conf, mgr.Logger())
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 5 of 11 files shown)


internal/driver/postgrespgx/driver.go
```diff
@@ -0,0 +1,18 @@
+package postgrespgx
```
gregfurman, line 1: Why do we need this `driver` approach? Can we not just directly swap out `lib/pq` usage with `"github.com/jackc/pgx/v4/stdlib"`?

maxtheaxe: Unfortunately, the `pgx` driver is just not called `postgres`, it's called `pgx` or `pgx/vX`:

https://github.com/jackc/pgx/blob/ae204a414cbdd6fdb503808221dda22335cf4fd9/stdlib/sql.go#L107-L117
```
func init() {
	pgxDriver = &Driver{
		configs: make(map[string]*pgx.ConnConfig),
	}


	// if pgx driver was already registered by different pgx major version then we
	// skip registration under the default name.
	if !slices.Contains(sql.Drivers(), "pgx") {
		sql.Register("pgx", pgxDriver)
	}
	sql.Register("pgx/v5", pgxDriver)

	// extra stuff
}
```

maxtheaxe: In the [_other_ PR I refer to the driver directly](https://github.com/redpanda-data/connect/pull/3674/files) as `pgx`, if you want to see that option.

gregfurman: Ahh I see. So by removing `lib/pq` we lose the `postgres` plugin, meaning we have ensure it's registered if we want the `sql.Open("postgres", "<conn>")` to work. 

I've done some reading on this and it seems that even the `lib/pq` maintainers suggest using the `pgx` driver 🤔 

See [lib/pq/README.md#status](https://github.com/lib/pq?tab=readme-ov-file#status):
> For users that require new features or reliable resolution of reported bugs, we recommend using [pgx](https://github.com/jackc/pgx) which is under active development.

Here's what I think we should do:

1. Register `pgx` as normal, and add an additonal `postgres_v2` driver to our SQL components, which acts as an alias to the underlying `pgx` driver.
   - `driver: postgres => lib/pq::postgres`
   - `driver: postgres_v2 => github.com/jackc/pgx/v4/stdlib::pgx`

```yaml
input:
  sql_select:
    driver: postgres_v2 # will use pgx
```

3. We can then add an integration test using this `pgx` driver (aliased in config by `postgres_v2`) in our `integration_test.go`.

 https://github.com/warpstreamlabs/bento/blob/e93b7b16ffac606bd80de781cbd0617670c11149/internal/impl/sql/integration_test.go#L731

4. Then, in the next release, we can change the default `postgres` to point to `pgx` and deprecate `lib/pq` support:
   - `driver: postgres_legacy => lib/pq::postgres`
   - `driver: postgres => github.com/jackc/pgx/v4/stdlib::pgx`
   - `driver: postgres_v2 => github.com/jackc/pgx/v4/stdlib::pgx` 

   Note: specifying this `driver: postgres_v2` should probably log a warning saying it's being removed in next minor release, else we just keep it around if it's not too much extra complexity.

internal/impl/sql/conn_fields.go
```diff
@@ -163,6 +163,26 @@ CREATE TABLE IF NOT EXISTS some_table (
...
+func warnIfPostgresV2(driver string, log *service.Logger) {
+	if driver != "postgres_v2" || log == nil {
+		return
+	}
+	warnPostgresV2Once.Do(func() {
+		log.With("driver", driver).Warn("driver 'postgres_v2' (jackc/pgx) is temporary and will become default 'postgres' in next minor release")
+	})
+}
+
+func getDriver(conf *service.ParsedConfig, log *service.Logger) (string, error) {
```
gregfurman, line 177: Let's log this in `sqlOpenWithReworks` rather? Then we wont need to use the `getDriver` in each component

https://github.com/warpstreamlabs/bento/blob/6640566a3a8a1acca639c3dd9176b30e61fbdd73/internal/impl/sql/conn_fields.go#L366

internal/driver/postgrespgx/driver.go
```diff
@@ -0,0 +1,18 @@
...
+
+import (
```
gregfurman, line 3: Can we not register `pgx` as a driver in `public`?

https://github.com/warpstreamlabs/bento/blob/ecb62f331da60bb2af9d97dd1c990346c41250b6/public/components/sql/package.go#L11-L12

internal/impl/sql/conn_fields.go
```diff
@@ -163,6 +163,26 @@ CREATE TABLE IF NOT EXISTS some_table (
 
 }
 
+var warnPostgresV2Once sync.Once
+
+func warnIfPostgresV2(driver string, log *service.Logger) {
+	if driver != "postgres_v2" || log == nil {
```
gregfurman, line 169: Can we add `postgres_v2` as a driver in the conn_fields? It should be in 

https://github.com/warpstreamlabs/bento/blob/6640566a3a8a1acca639c3dd9176b30e61fbdd73/internal/impl/sql/conn_fields.go#L19-L20

gregfurman:
Hey @maxtheaxe 👋 Do you still have capacity for this? Otherwise happy for myself (or @jem-davies) to push it over-the-line.

Lmk!

maxtheaxe:
Hey @gregfurman. Sorry for the delay here—been busy. I will try to find time this weekend. If I can't, it's all yours. Apologies for the trouble.

gregfurman:
@maxtheaxe No worries at all! This isn't a blocker for anyone atm, so I just wanted to make sure that you're still keen to finish it off. 

TBH I think the library change you proposed is a super neat addition and would love to see it actualised 🙂

maxtheaxe:
I'm definitely still game, it just isn't critical to what I'm doing at work. We're launching very shortly, so I'll do it on my own time. I appreciate you following up here to check in!

maxtheaxe:
Hey @gregfurman. Sadly, I think I'm probably letting either pride or selfishness get in the way of a good feature here. I still don't have time, but if it becomes a blocker, please don't hesitate to finish without me. If all else fails, eventually I will get to it.

To be clear, I'm not saying I don't still _want_ to do it, I'm just recognizing how long it's dragged on at this point.

gregfurman:
@maxtheaxe Thanks for letting me know! I'll place this back into draft and you can pick it back up when you're ready 🙂

PR #579, opened 2025-11-28
feat(bloblang): Add file_rel_json function for cached JSON parsing

triddell:
## Summary

This PR adds a new `file_rel_json()` Bloblang function that caches both the file read and JSON parse operations, significantly improving performance for large lookup tables accessed on every message.

## Problem

The existing `file_rel().parse_json()` pattern caches the file read but parses the JSON on every invocation. For a 302KB routing table accessed on ~131k records, this causes:
- **130,900x JSON parse operations** (~277μs overhead per record)
- **15.3x slower performance** compared to native Go (38.7s vs 2.5s)

## Solution

The new `file_rel_json()` function:
- Reads and parses the JSON file once during function initialization
- Caches the parsed Go object (not just bytes)
- Returns the cached result on every invocation
- Uses the same `Static()` optimization as other file functions

## Performance Impact

Expected improvement: **10-15x faster** for JSON lookup use cases, bringing performance close to native Go implementation.

## Example Usage

```bloblang
# Before (slow - parses JSON on every message)
let routing = file_rel("routing-table.json").parse_json()
root.target = $routing.get(this.id).or("default")

# After (fast - JSON parsed once at startup)
let routing = file_rel_json("routing-table.json")
root.target = $routing.get(this.id).or("default")
```

## Implementation Details

- Added to `internal/impl/io/bloblang.go`
- Follows existing patterns from `file_rel()` and `env()` functions
- Uses `args.ImportFile()` for relative path resolution
- Marked as `Static()` for proper caching behavior
- Includes test example following project conventions

## Testing

- All existing tests pass
- New function includes test example that validates the cached JSON parsing
- Builds successfully

## Benchmark Context

This optimization was identified through benchmarking Bento against native Go for enrichment workloads. The bottleneck was not the file read (which was already cached) but the repeated JSON parsing on every record.

### DIFF
--- a/website/docs/guides/bloblang/functions.md
+++ b/website/docs/guides/bloblang/functions.md
@@ -595,6 +595,24 @@ root.doc = file_rel(path: env("BENTO_TEST_BLOBLANG_FILE"), no_cache: true).parse
 # Out: {"doc":{"foo":"bar"}}
 ```
 
+### `file_rel_json`
+
+Reads a JSON file and returns the parsed object. The file is read and parsed only once, with the result cached for all subsequent calls. Relative paths are resolved from the directory of the mapping. This is significantly more efficient than `file_rel().parse_json()` which parses the JSON on every invocation.
+
+#### Parameters
+
+**`path`** &lt;string&gt; The path of the target JSON file.  
+
+#### Examples
+
+
+```coffee
+root.doc = file_rel_json(env("BENTO_TEST_BLOBLANG_FILE"))
+
+# In:  {}
+# Out: {"doc":{"foo":"bar"}}
+```
+
 ### `hostname`
 
 Returns a string matching the hostname of the machine running Bento.


internal/impl/io/bloblang.go
```diff
@@ -183,4 +185,42 @@ func init() {
...
+
+			var parsedJSON any
+			if err := json.Unmarshal(fileBytes, &parsedJSON); err != nil {
+				return nil, fmt.Errorf("failed to parse JSON from %s: %w", path, err)
+			}
+
+			// Return a function that always returns the cached parsed result
+			return func() (any, error) {
+				return parsedJSON, nil
+			}, nil
```
gregfurman, line 221: Can we use the same approach as `file_rel` and allow for caching to be optional?

i.e 

https://github.com/warpstreamlabs/bento/blob/e93b7b16ffac606bd80de781cbd0617670c11149/internal/impl/io/bloblang.go#L164-L166

and 

https://github.com/warpstreamlabs/bento/blob/e93b7b16ffac606bd80de781cbd0617670c11149/internal/impl/io/bloblang.go#L175-L181

jem-davies: I think we could merge it as-is and later add the option `no_cache` because the `file_rel`'s option is defaulting to `false`? 

Edit - more time to think about potentially doing a 'reload' if the file's metadata indicates it's been altered / SHA hash - instead of what would be the equivalent to what we already can do with the `file_rel` function - if that makes sense?

PR #585, opened 2025-12-01
couchbase: add counter operations increment and decrement

sapk:
This PR add atomic counter operations support for Couchbase processor.

This is backward compatible as they are new operations.

There is also append, prepend and touch operation that could be added later.

### DIFF
--- a/website/docs/components/processors/couchbase.md
+++ b/website/docs/components/processors/couchbase.md
@@ -183,13 +183,13 @@ Default: `"get"`
 
 | Option | Summary |
 |---|---|
-| `decrement` | decrement a counter by the value in content. |
-| `get` | fetch a document. |
-| `increment` | increment a counter by the value in content. |
-| `insert` | insert a new document. |
-| `remove` | delete a document. |
-| `replace` | replace the contents of a document. |
-| `upsert` | creates a new document if it does not exist, if it does exist then it updates it. |
+| `decrement` | Decrement a counter by the value in content, if it does not exist then it creates a counter with an initial value equal to the negative of the value in content. If the initial value is less than or equal to 0, a document not found error is returned. |
+| `get` | Fetch a document. |
+| `increment` | Increment a counter by the value in content, if it does not exist then it creates a counter with an initial value equal to the value in content. If the initial value is less than or equal to 0, a document not found error is returned. |
+| `insert` | Insert a new document. |
+| `remove` | Delete a document. |
+| `replace` | Replace the contents of a document. |
+| `upsert` | Creates a new document if it does not exist, if it does exist then it updates it. |
 
 
 ### `cas_enabled`


internal/impl/couchbase/couchbase.go
```diff
@@ -75,3 +86,33 @@ func upsert(key string, data []byte, cas gocb.Cas) gocb.BulkOp {
...
+	if len(data) > 0 {
+		if d, err := strconv.ParseInt(string(data), 10, 64); err == nil {
+			delta = d
+		}
+	}
+
+	return &gocb.DecrementOp{
+		ID:      key,
+		Delta:   delta,
+		Initial: -delta, // Default initial to -delta
```
jem-davies, line 138: You can't have a counter < 0? Is the intention here then - that if the counter doesn't exist, the processor will error?

internal/impl/couchbase/processor_test.go
```diff
@@ -82,6 +82,26 @@ couchbase:
...
+			name: "decrement without content",
+			config: `
+couchbase:
+  url: 'url'
+  bucket: 'bucket'
+  id: '${! json("id") }'
+  operation: 'decrement'
 `,
 		},
 	}
```
jem-davies, line 131: So if we do an increment without a content - what do we want to happen? - at the moment I am seeing that the counter is not incremented unless a content field is supplied.

internal/impl/couchbase/processor.go
```diff
@@ -40,6 +40,9 @@ func ProcessorConfig() *service.ConfigSpec {
 			string(client.OperationRemove):  "delete a document.",
 			string(client.OperationReplace): "replace the contents of a document.",
 			string(client.OperationUpsert):  "creates a new document if it does not exist, if it does exist then it updates it.",
+
+			string(client.OperationIncrement): "increment a counter.",
+			string(client.OperationDecrement): "decrement a counter.",
```
jem-davies:
```suggestion
			string(client.OperationIncrement): "increment a counter.",
			string(client.OperationDecrement): "decrement a counter.",
```

nit

internal/impl/couchbase/processor_test.go
jem-davies: So if I do: 

```yaml
    - couchbase:
        url: couchbase://localhost:11210
        username: bento
        password: password
        bucket: bento-test
        id: ${! json("id") }
        operation: increment
    - catch:
      - log:
          message: "Processing failed due to: ${!error()}"
          level: ERROR
```

Where the message is: `{"id":4}` - which doesn't exist - I get the error: 

```
ERRO Processing failed due to: couchbase operator failed: document not found
```

But if I include the content: 

```
...
        id: ${! json("id") }
        operation: increment
        content: 1
...
```

Then the id 4 is created and initialised at 1. 

Which makes sense. 

But then if I run the first config again - without the content - I just get the value of the counter at that time. 

Because Go's default value for an int will be 0 - so I don't actually increment. Seems like a content should always be supplied with the increment / decrement operation?

internal/impl/couchbase/couchbase.go
sapk: I didn't put more thinking that it could allow to work if delta is negative (setting to a positive value in this case). But I could add a test to validate the case to know if it trigger an error or not.

internal/impl/couchbase/processor_test.go
sapk: It felt to be the safest to be an noop if nothing is set or it is an invalid number but we can set the default value to one [here](https://github.com/warpstreamlabs/bento/pull/585/files/4d5e3f0541d0dd27332c7126c79d2f034eeea1b6#diff-7ea89c630a38188ccfd4371131b75c9a307c1c9fb5e69c8b0a7a9f3f97bd4570R91).
I could maybe improve this part by returning an error in case of invalid input (empty or not an number).

internal/impl/couchbase/couchbase.go
sapk: I did some testing and calling decrement on a non existing counter with a initial value <= 0 trigger the document not found error. 
So setting the initial value to zero doesn't help to initialize to zero the counter.  
At least with -delta, in the strange corner-case, if the value is negative it pass.

internal/impl/couchbase/processor_test.go
sapk: I added to make the content mandatory and validate that it is an integer. 
So now if not set it will error. 
If set explicitly to zero it retrieve the counter unchanged.

We could set the default value to 1 if content is not set or zero to be an noop by default.
Let me know if that what you want or if we stay with no default value.

internal/impl/couchbase/processor.go
```diff
@@ -37,14 +37,16 @@ func ProcessorConfig() *service.ConfigSpec {
...
-			string(client.OperationRemove):  "delete a document.",
-			string(client.OperationReplace): "replace the contents of a document.",
-			string(client.OperationUpsert):  "creates a new document if it does not exist, if it does exist then it updates it.",
+			string(client.OperationGet):       "fetch a document.",
+			string(client.OperationInsert):    "insert a new document.",
+			string(client.OperationRemove):    "delete a document.",
+			string(client.OperationReplace):   "replace the contents of a document.",
+			string(client.OperationUpsert):    "creates a new document if it does not exist, if it does exist then it updates it.",
+			string(client.OperationIncrement): "increment a counter by the value in content.",
+			string(client.OperationDecrement): "decrement a counter by the value in content.",
```
jem-davies:
```suggestion
			string(client.OperationGet):       "Fetch a document.",
			string(client.OperationInsert):    "Insert a new document.",
			string(client.OperationRemove):    "Delete a document.",
			string(client.OperationReplace):   "Replace the contents of a document.",
			string(client.OperationUpsert):    "Creates a new document if it does not exist, if it does exist then it updates it.",
			string(client.OperationIncrement): "Increment a counter by the value in content, if it does not exist then it creates a counter with an initial value equal to the value in content. If the initial value is less than or equal to 0, a document not found error is returned.",
			string(client.OperationDecrement): "Decrement a counter by the value in content, if it does not exist then it creates a counter with an initial value equal to the negative of the value in content. If the initial value is less than or equal to 0, a document not found error is returned.",
```

sapk: It is better explain that way than in the gocb pkg itself.

PR #595, opened 2025-12-03
fix: remove client.Disconnect from mongodb output

### LINKED ISSUE #576: mongodb output is stuck in infinite ping loop when started before replica set is ready - never recovers
Hello,

I've encountered a bug where MongoDB output gets stuck in an infinite loop. The problem is unfortunately somewhat difficult to reproduce, but it reliably occurs when Bento is running in a Kubernetes cluster that also has a MongoDB with a replica count greater than 2. If Bento is started while the MongoDB isn't ready, Bento disconnects and then gets stuck in a ping loop instead of attempting to establish a new connection. When the MongoDB subsequently becomes ready, Bento doesn't detect this and simply remains stuck.

The logs look like this:

```bash
ERRO Failed to connect to mongodb: ping failed: server selection error: server selection timeout, current topology: { Type: ReplicaSetNoPrimary, Servers: [{ Addr: mongodb-0.mongodb-svc.analytic.svc.cluster.local:27017, Type: Unknown, Last error: dial tcp: lookup mongodb-0.mongodb-svc.analytic.svc.cluster.local: no such host }, { Addr: mongodb-1.mongodb-svc.analytic.svc.cluster.local:27017, Type: Unknown, Last error: dial tcp: lookup mongodb-1.mongodb-svc.analytic.svc.cluster.local: no such host }, ] }  @service=bento label="" path=root.output
ERRO Failed to connect to mongodb: ping failed: client is disconnected  @service=bento label="" path=root.output
ERRO Failed to connect to mongodb: ping failed: client is disconnected  @service=bento label="" path=root.output
ERRO Failed to connect to mongodb: ping failed: client is disconnected  @service=bento label="" path=root.output
ERRO Failed to connect to mongodb: ping failed: client is disconnected  @service=bento label="" path=root.output
ERRO Failed to connect to mongodb: ping failed: client is disconnected  @service=bento label="" path=root.output
ERRO Failed to connect to mongodb: ping failed: client is disconnected  @service=bento label="" path=root.output
ERRO Failed to connect to mongodb: ping failed: client is disconnected  @service=bento label="" path=root.output
ERRO Failed to connect to mongodb: ping failed: client is disconnected  @service=bento label="" path=root.output
ERRO Failed to connect to mongodb: ping failed: client is disconnected  @service=bento label="" path=root.output
ERRO Failed to connect to mongodb: ping failed: client is disconnected  @service=bento label="" path=root.output
ERRO Failed to connect to mongodb: ping failed: client is disconnected  @service=bento label="" path=root.output
ERRO Failed to connect to mongodb: ping failed: client is disconnected  @service=bento label="" path=root.output
ERRO Failed to connect to mongodb: ping failed: client is disconnected  @service=bento label="" path=root.output
```

I already think that I have found the error in code. 

I'd really like to solve this too. Would a contribution be welcome? I have experience developing Go applications, so quality shouldn't be a problem. Of course, I don't want to put in the work if nobody's interested.

Greetings,
Dominik

mladBlum:
Fixes #576 .

Please review my change carefully, because its my first contact with the bento code at all.

What and Why:  

So bento currently keeps looping in "ping failed: client is disconnected" if client.Ping once throws an error. Bento will never recover from this by its own. Even if the ping works again.

What I figured out is, that if client.Ping throws an error we do a m.client.Disconnect. That's fatal because after that, a reconnect needs to be done, but in this loop there is nothing like this.
The disconnect should not be done at all. Just return the ping error and let the retry loop try again.

I tested the code and now even the errormessage is more appropriate on whats going on and why ping is failing.

I am not sure if the mongodb input is also affected by this error. Probably not because there is no disconnect if the ping fails.

Greetings,
Dominik

### DIFF
--- a/internal/impl/mongodb/output.go
+++ b/internal/impl/mongodb/output.go
@@ -107,7 +107,6 @@ func (m *outputWriter) Connect(ctx context.Context) error {
 	defer m.mu.Unlock()
 
 	if err := m.client.Ping(ctx, nil); err != nil {
-		_ = m.client.Disconnect(ctx)
 		return fmt.Errorf("ping failed: %v", err)
 	}
 	return nil


mladBlum:
@gregfurman Thank you! Yes and I totally aggree with you. The fact that the connect function is called "connect" and not "connects" is very strange in itself.

PR #606, opened 2025-12-05
feat: Support for generic registrations against environment

gregfurman:
## Motivation

Supports registering arbitrary constructors against the bento environment. 

This has two main outcomes:
1. Allows us to bridge the gap between the bloblang environment and bento resources (see companion PR)
2. Provides an alternative approach to our (frankly hacky) `bundle` based approach for customising the entire blobl environment on startup.

## Changes

- Creates a `ConstructorSet` which contains a slice of all constructors registered. These are run in order of registration and after `resource` configurations have been initialised.
- Adds a `LimitedResources` interface, intended for usage with these new constructors, which constrains what operations should be possible when registering a custom managed constructor against the env.

### DIFF
--- a/internal/bundle/environment.go
+++ b/internal/bundle/environment.go
@@ -18,20 +18,23 @@ type Environment struct {
 	tracers    *TracerSet
 
 	scanners *ScannerSet
+
+	constructors *ConstructorSet
 }
 
 // NewEnvironment creates an empty environment.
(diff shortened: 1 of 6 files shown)


jem-davies:
I have taken a look at #607

jem-davies (approved):

PR #607, opened 2025-12-05
add: Bloblang functions for operating on a resource cache

gregfurman:
## Motivation

This PR adds functionality to directly call a `cache` resource from bloblang. It leverages the (proposed) generic environment constructors to ensure that 

Requires https://github.com/warpstreamlabs/bento/pull/606

## Changes

- Adds `cache_get`, `cache_set`, `cache_add`, and `cache_delete` functions to our bloblang environment that can be used against a live cache resource.

### DIFF
--- a/internal/impl/pure/bloblang_cache.go
+++ b/internal/impl/pure/bloblang_cache.go
@@ -24,10 +24,12 @@ func init() {
 		panic(err)
 	}
 
-	service.RegisterManagedConstructor(func(mgr service.LimitedResources) error {
+	if err := service.RegisterManagedConstructor(func(mgr service.LimitedResources) error {
 		container.Store(mgr)
 		return nil
-	})
+	}); err != nil {
+		panic(err)
+	}
 
 }
 
@@ -101,10 +103,7 @@ func registerBloblangFunctions(getMgr func() service.LimitedResources) error {
 				}); err != nil {
 					return nil, err
 				}
-				if cerr != nil {
-					return nil, cerr
-				}
-				return nil, nil
+				return nil, cerr
 			}, nil
 		}); err != nil {
 		return err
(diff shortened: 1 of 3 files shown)


public/service/resources.go
```diff
@@ -16,6 +16,28 @@ import (
 	"github.com/warpstreamlabs/bento/internal/manager/mock"
 )
 
+var _ LimitedResources = (*Resources)(nil)
+
+// LimitedResources provides constrained access a subset of resources
```
jem-davies:
```suggestion
// LimitedResources provides constrained access to a subset of resources
```

internal/impl/pure/bloblang_cache.go
```diff
@@ -0,0 +1,189 @@
...
+				ctx := context.Background()
+				if err := getMgr().AccessCache(ctx, resource, func(c service.Cache) {
+					cerr = c.Add(ctx, key, []byte(value), nil)
+				}); err != nil {
+					return nil, err
+				}
+				if cerr != nil {
+					return nil, cerr
+				}
+				return value, nil
```
jem-davies: In Set & Delete we return nil - but here we return the value - is that intentional?

internal/impl/pure/bloblang_cache.go
```diff
@@ -0,0 +1,189 @@
...
+
+	getManager := func() service.LimitedResources {
+		return container.Load().(service.LimitedResources)
+	}
+
+	if err := registerBloblangFunctions(getManager); err != nil {
+		panic(err)
+	}
+
+	service.RegisterManagedConstructor(func(mgr service.LimitedResources) error {
```
jem-davies: (from lint output) - we are ignoring an error here

internal/impl/pure/bloblang_cache.go
gregfurman: It was intentional... but I think you're correct to point out this discrepency. Also the cache definition doesn't have `Add` return anything so I suppose for consistency this should be removed.

PR #608, opened 2025-12-05
feat: allow configuring AWS expiry window

gitphill:
Add `expiry_window` as a confiuration option under `aws.credentials`.

Allow the credentials to trigger refreshing prior to the credentials actually expiring. This is beneficial so race conditions with expiring credentials do not cause requests to fail. Should be a duration, valid time units are 'ns', 'us' (or 'µs'), 'ms', 's', 'm', 'h'. For example '10s' would refresh credentials ten seconds before expiration.

### DIFF
--- a/internal/impl/aws/config/config.go
+++ b/internal/impl/aws/config/config.go
@@ -37,8 +37,8 @@ func SessionFields() []*service.ConfigField {
 			service.NewStringField("role_external_id").
 				Description("An external ID to provide when assuming a role.").
 				Default("").Advanced(),
-			service.NewStringField("expiry_window").
-				Description("Allow the credentials to trigger refreshing prior to the credentials actually expiring. This is beneficial so race conditions with expiring credentials do not cause requests to fail. Should be a duration, valid time units are 'ns', 'us' (or 'µs'), 'ms', 's', 'm', 'h'. For example '10s' would refresh credentials ten seconds before expiration.").
+			service.NewDurationField("expiry_window").
+				Description("Allow the credentials to trigger refreshing prior to the credentials actually expiring. This is beneficial so race conditions with expiring credentials do not cause requests to fail. For example '10s' would refresh credentials ten seconds before expiration. Setting to a duration of `0` disables the expiry window.").
 				Default("").Advanced()).
 			Advanced().
 			Description("Optional manual configuration of AWS credentials to use. More information can be found [in this document](/docs/guides/cloud/aws)."),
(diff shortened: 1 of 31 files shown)


gitphill:
is this page manually maintained https://warpstreamlabs.github.io/bento/docs/guides/cloud/aws? i did `make docs` which updated most of the md files, but not this one

gitphill:
I put some words around the expiry_window setting into website/docs/guides/cloud/aws.md

internal/impl/aws/config/config.go
```diff
@@ -36,6 +36,9 @@ func SessionFields() []*service.ConfigField {
 				Default("").Advanced(),
 			service.NewStringField("role_external_id").
 				Description("An external ID to provide when assuming a role.").
+				Default("").Advanced(),
+			service.NewStringField("expiry_window").
+				Description("Allow the credentials to trigger refreshing prior to the credentials actually expiring. This is beneficial so race conditions with expiring credentials do not cause requests to fail. Should be a duration, valid time units are 'ns', 'us' (or 'µs'), 'ms', 's', 'm', 'h'. For example '10s' would refresh credentials ten seconds before expiration.").
```
gregfurman:
```suggestion
			service.NewDurationField("expiry_window").
				Description("Allow the credentials to trigger refreshing prior to the credentials actually expiring. This is beneficial so race conditions with expiring credentials do not cause requests to fail. For example '10s' would refresh credentials ten seconds before expiration. Setting to a duration of `0` disables the expiry window.").
```

internal/impl/aws/session.go
```diff
@@ -21,6 +22,16 @@ func int64Field(conf *service.ParsedConfig, path ...string) (int64, error) {
...
 }
 
+func getCredentialsCacheOptions(conf *service.ParsedConfig) func(*aws.CredentialsCacheOptions) {
+	var expiryWindow time.Duration
+	if expiryWindowStr, _ := conf.FieldString("expiry_window"); expiryWindowStr != "" {
+		expiryWindow, _ = time.ParseDuration(expiryWindowStr)
+	}
+	return func(cco *aws.CredentialsCacheOptions) {
+		cco.ExpiryWindow = expiryWindow
+	}
```
gregfurman:
```suggestion
    expiryWindow, _ := conf.FieldDuration("expiry_window")
	return func(cco *aws.CredentialsCacheOptions) {
		cco.ExpiryWindow = expiryWindow
	}
```

internal/impl/aws/config/config.go
gitphill: ah yep lets do that, i didn't know NewDurationField was a thing 🎉

PR #610, opened 2025-12-05
Add strict_schema configuration option for Parquet components

triddell:
# Add `strict_schema` configuration option for Parquet components

## Summary

This PR adds a `strict_schema` boolean configuration option (defaults to `true`) to Bento's parquet input, `parquet_decode` processor, and `parse_parquet()` bloblang function. This allows users to read Parquet files with non-standard schema structures while maintaining backward compatibility with strict validation as the default.

## Problem Statement

Bento currently fails to read certain Parquet files that use non-standard schema structures, specifically files with non-compliant LIST type definitions. This affects files from sources like AWS Security Lake and potentially other data producers that generate Parquet files using older or non-standard conventions.

### Error Example
```
decoding panic: node with logical type LIST is not composed of a repeated .list.element
```

### Root Cause

Bento uses `parquet.GenericReader[any]` from the parquet-go library, which enforces strict Parquet schema validation. While this is correct according to the Parquet specification, some real-world files use alternative LIST representations that are valid but non-standard.

**Non-standard format** (e.g., AWS Security Lake):
```
optional group profiles (LIST) {
  repeated binary array (STRING);
}
```

**Standard format**:
```
optional group profiles (LIST) {
  repeated group list {
    optional binary element (STRING);
  }
}
```

## Solution

Add a `strict_schema` boolean configuration option (defaulting to `true`) that:
- Maintains current strict validation behavior by default (backward compatible)
- When set to `false`, uses a lenient reading approach that bypasses strict schema validation
- Uses the non-deprecated `parquet.RowReader` interface instead of the deprecated `parquet.Reader`

### Architecture

**Strict Mode** (`strict_schema: true`, default):
- Uses `parquet.GenericReader[any]` for strict schema validation
- Existing behavior, fully backward compatible

**Lenient Mode** (`strict_schema: false`):
- Uses `parquet.File.RowGroups()` to iterate through row groups
- Each `RowGroup.Rows()` returns a `parquet.Rows` which implements `parquet.RowReader`
- Manually converts `parquet.Row` to `map[string]any` to avoid schema validation
- Proper resource cleanup with `Close()` on `Rows` readers

#### 1. **internal/impl/parquet/common.go** (NEW)
**Why a separate file?** These lenient reading functions are shared across three components: the parquet input, the parquet_decode processor, and the parse_parquet() bloblang function. Creating a separate file avoids code duplication and ensures consistent behavior across all three.

Shared functions for lenient parquet reading using the `RowReader` interface:

- `readLenient(rdr parquet.RowReader, schema *parquet.Schema, rows []any)` - Reads rows using the RowReader interface and converts them to map[string]any
- `convertParquetValue(vals []parquet.Value, field parquet.Field)` - Converts parquet values to Go types, handling repeated fields (arrays)
- `convertSingleValue(val parquet.Value)` - Converts individual parquet values based on their kind (Boolean, Int32, Int64, Float, Double, ByteArray, etc.)

Key features:
- Uses `parquet.RowReader` interface (not deprecated)
- Manual row-by-row conversion bypasses strict schema validation
- Handles repeated fields as Go slices
- Converts ByteArray to strings for JSON compatibility

#### 2. **internal/impl/parquet/input_parquet.go** (MODIFIED)
Added `strict_schema` configuration to the parquet input component:

- Added `strict_schema` field to configuration spec (default: `true`)
- Updated `openParquetFile` struct to track RowGroups and iteration state:
  ```go
  type openParquetFile struct {
      schema          *parquet.Schema
      handle          fs.File
      rdr             *parquet.GenericReader[any]
      lenientRGs      []parquet.RowGroup
      lenientRGIdx    int
      lenientRowsRdr  parquet.Rows
      strictSchema    bool
  }
  ```
- Modified `getOpenFile()` to create appropriate reader based on `strict_schema` setting
- Updated `ReadBatch()` to handle lenient mode with RowGroup iteration:
  - Opens each RowGroup's Rows reader sequentially
  - Reads batches from each RowGroup
  - Properly closes Rows readers when RowGroups are exhausted
- Proper cleanup in `Close()` method for both reader types

#### 3. **internal/impl/parquet/processor_decode.go** (MODIFIED)
Added `strict_schema` configuration to the `parquet_decode` processor:

- Added `strict_schema` field to configuration spec (default: `true`)
- Added `strictSchema` field to `parquetDecodeProcessor` struct
- Updated `Process()` method with conditional logic:
  - Strict mode: Uses existing `GenericReader[any]` approach
  - Lenient mode: Iterates through `RowGroups()`, using `readLenient()` for each
- Properly closes `Rows` readers with defer after processing each RowGroup

#### 4. **internal/impl/parquet/bloblang.go** (MODIFIED)
Added `strict_schema` parameter to the `parse_parquet()` bloblang function:

- Added `strict_schema` parameter to plugin spec (default: `true`)
- Added example: `root = content().parse_parquet(strict_schema: false)`
- Updated implementation to use RowGroups iteration in lenient mode
- Proper resource cleanup with defer for each RowGroup's Rows reader

#### 5. **internal/impl/parquet/processor_decode_test.go** (MODIFIED)
Fixed existing tests to properly initialize the `strictSchema` field:

- Added `strictSchema: true` to all `parquetDecodeProcessor` instantiations in existing tests
- Ensures existing tests maintain expected default behavior
- No test logic changes, only proper struct initialization

#### 6. **internal/impl/parquet/strict_schema_test.go** (NEW)
Comprehensive test coverage for the new functionality:

**Test: `TestParquetStrictSchemaInput`**
- Sub-test: "strict_schema true should handle standard files"
  - Creates standard parquet file with compliant schema
  - Verifies strict mode reads it successfully
- Sub-test: "strict_schema false should handle non-standard files"
  - Creates non-standard parquet file with `Repeated` fields (non-standard LIST)
  - Verifies lenient mode reads and parses data correctly
  - Validates all field values and array contents

**Test: `TestParquetStrictSchemaProcessor`**
- Sub-test: "strict_schema false should decode non-standard files"
  - Tests processor with non-standard schema
  - Validates correct decoding of repeated fields
- Sub-test: "strict_schema true should decode standard files"
  - Ensures strict mode still works for standard files

**Test: `TestParquetStrictSchemaDefault`**
- Verifies that `strict_schema` defaults to `true` when not specified
- Confirms backward compatibility

#### 7. **internal/impl/parquet/bloblang_test.go** (MODIFIED)
Added test for bloblang function with lenient schema:

**Test: `TestParquetParseBloblangLenientSchema`**
- Creates non-standard parquet file with repeated field
- Tests `parse_parquet(strict_schema: false)`
- Validates correct parsing and JSON serialization

### Parquet Input
```yaml
input:
  parquet:
    paths:
      - /path/to/files/*.parquet
    strict_schema: false  # Allow non-standard schema structures
```

### Parquet Decode Processor
```yaml
input:
  aws_s3:
    bucket: my-bucket
    prefix: data/
  processors:
    - parquet_decode:
        strict_schema: false  # Allow non-standard schema structures
```

### Bloblang Function
```coffeescript
root = content().parse_parquet(strict_schema: false)
```

## Benefits

1. **Backward Compatibility**: Defaults to `true`, maintaining current strict validation behavior
2. **Flexibility**: Users can opt-in to lenient mode when needed for non-standard files
3. **Real-world Compatibility**: Enables reading files from AWS Security Lake and similar sources
4. **Simple Configuration**: Single boolean flag controls the behavior
5. **No Deprecated APIs**: Uses `parquet.RowReader` interface instead of deprecated `parquet.Reader`
6. **Comprehensive Coverage**: Available in input, processor, and bloblang function

## Testing

- **21+ tests passing**, including:
  - 3 new tests specifically for `strict_schema` functionality
  - All existing tests updated and passing
  - Compression tests passing
  - Round-trip encode/decode tests passing
  - Bloblang function tests with lenient schema

- **Manual testing**: Verified with actual AWS Security Lake Parquet files that previously failed. With `strict_schema: false`, these files are successfully read and parsed into valid JSON output.

## Breaking Changes

None. The change is purely additive with `strict_schema: true` as the default, maintaining existing behavior.

## Additional Notes

- The parquet-go library itself supports both strict and lenient reading modes
- This addresses production use cases where users need to consume Parquet files from external sources they don't control
- The implementation avoids using deprecated APIs by using the `parquet.RowReader` interface
- Proper resource management with explicit Close() calls on all readers

### DIFF
--- a/internal/impl/parquet/bloblang.go
+++ b/internal/impl/parquet/bloblang.go
@@ -56,7 +56,7 @@ func init() {
 							break
 						}
 
-						for i := 0; i < n; i++ {
+						for i := range n {
 							result = append(result, rowBuf[i])
 						}
 					}
@@ -71,21 +71,22 @@ func init() {
 					// Use RowGroups to avoid deprecated Reader
 					for _, rg := range inFile.RowGroups() {
 						rows := rg.Rows()
-						defer rows.Close()
 
 						for {
 							n, err := readLenient(rows, schema, rowBuf)
 							if err != nil && !errors.Is(err, io.EOF) {
+								rows.Close()
 								return nil, err
 							}
 							if n == 0 {
 								break
 							}
 
-							for i := 0; i < n; i++ {
+							for i := range n {
 								result = append(result, rowBuf[i])
 							}
 						}
+						rows.Close()
 					}
 				}
 
--- a/internal/impl/parquet/common.go
+++ b/internal/impl/parquet/common.go
@@ -3,6 +3,7 @@ package parquet
 import (
 	"errors"
 	"io"
+	"unicode/utf8"
 
 	"github.com/parquet-go/parquet-go"
 )
@@ -86,9 +87,11 @@ func convertSingleValue(val parquet.Value) any {
 	case parquet.Double:
 		return val.Double()
 	case parquet.ByteArray:
-		// Try to convert to string if it looks like UTF-8
 		b := val.ByteArray()
-		return string(b)
+		if utf8.Valid(b) {
+			return string(b)
+		}
+		return b
 	case parquet.FixedLenByteArray:
 		return val.ByteArray()
 	default:
--- a/internal/impl/parquet/processor_decode.go
+++ b/internal/impl/parquet/processor_decode.go
@@ -138,7 +138,7 @@ func (s *parquetDecodeProcessor) Process(ctx context.Context, msg *service.Messa
 				break
 			}
 
-			for i := 0; i < n; i++ {
+			for i := range n {
 				newMsg := msg.Copy()
 
 				if s.useLegacyListFormat {
@@ -155,18 +155,18 @@ func (s *parquetDecodeProcessor) Process(ctx context.Context, msg *service.Messa
 		// Use RowGroups to avoid deprecated Reader
 		for _, rg := range inFile.RowGroups() {
 			rows := rg.Rows()
-			defer rows.Close()
 
 			for {
 				n, err := readLenient(rows, schema, rowBuf)
 				if err != nil && !errors.Is(err, io.EOF) {
+					rows.Close()
 					return nil, err
 				}
 				if n == 0 {
 					break
 				}
 
-				for i := 0; i < n; i++ {
+				for i := range n {
 					newMsg := msg.Copy()
 
 					if s.useLegacyListFormat {
@@ -177,6 +177,7 @@ func (s *parquetDecodeProcessor) Process(ctx context.Context, msg *service.Messa
 					resBatch = append(resBatch, newMsg)
 				}
 			}
+			rows.Close()
 		}
 	}
 


internal/impl/parquet/bloblang.go
```diff
@@ -17,37 +18,74 @@ func init() {
...
+						for {
+							n, err := readLenient(rows, schema, rowBuf)
+							if err != nil && !errors.Is(err, io.EOF) {
+								return nil, err
+							}
+							if n == 0 {
+								break
+							}
+
+							for i := 0; i < n; i++ {
```
jem-davies:
```suggestion
							for i := range n {
```

nit - my IDE just points this "modernisation" out

internal/impl/parquet/bloblang.go
```diff
@@ -17,37 +18,74 @@ func init() {
...
+					for {
+						n, err := readWithoutPanic(pRdr, rowBuf)
+						if err != nil && !errors.Is(err, io.EOF) {
+							return nil, err
+						}
+						if n == 0 {
+							break
+						}
+
+						for i := 0; i < n; i++ {
```
jem-davies:
```suggestion
						for i := range n {
```

nit - my IDE just points this "modernisation" out

internal/impl/parquet/processor_decode.go
```diff
@@ -113,32 +120,63 @@ func (s *parquetDecodeProcessor) Process(ctx context.Context, msg *service.Messa
...
+			if err != nil && !errors.Is(err, io.EOF) {
+				return nil, err
+			}
+			if n == 0 {
+				break
 			}
 
-			newMsg.SetStructuredMut(rowBuf[i])
-			resBatch = append(resBatch, newMsg)
+			for i := 0; i < n; i++ {
```
jem-davies:
```suggestion
			for i := range n {
```

nit

internal/impl/parquet/processor_decode.go
```diff
@@ -113,32 +120,63 @@ func (s *parquetDecodeProcessor) Process(ctx context.Context, msg *service.Messa
...
+			for {
+				n, err := readLenient(rows, schema, rowBuf)
+				if err != nil && !errors.Is(err, io.EOF) {
+					return nil, err
+				}
+				if n == 0 {
+					break
+				}
+
+				for i := 0; i < n; i++ {
```
jem-davies:
```suggestion
				for i := range n {
```

nit

internal/impl/parquet/processor_decode.go
```diff
@@ -113,32 +120,63 @@ func (s *parquetDecodeProcessor) Process(ctx context.Context, msg *service.Messa
...
+				resBatch = append(resBatch, newMsg)
+			}
+		}
+	} else {
+		schema := inFile.Schema()
+
+		// Use RowGroups to avoid deprecated Reader
+		for _, rg := range inFile.RowGroups() {
+			rows := rg.Rows()
+			defer rows.Close()
```
jem-davies: Should we close at the end of each loop iteration - rather than deferring multiple calls until we exit the func?

internal/impl/parquet/bloblang.go
```diff
@@ -17,37 +18,74 @@ func init() {
...
 					}
 
-					for i := 0; i < n; i++ {
-						result = append(result, rowBuf[i])
+					schema := inFile.Schema()
+
+					// Use RowGroups to avoid deprecated Reader
+					for _, rg := range inFile.RowGroups() {
+						rows := rg.Rows()
+						defer rows.Close()
```
jem-davies: Should we close at the end of each loop iteration - rather than deferring multiple calls until we exit the func?

internal/impl/parquet/common.go
```diff
@@ -0,0 +1,97 @@
...
+	case parquet.Int32:
+		return val.Int32()
+	case parquet.Int64:
+		return val.Int64()
+	case parquet.Float:
+		return val.Float()
+	case parquet.Double:
+		return val.Double()
+	case parquet.ByteArray:
+		// Try to convert to string if it looks like UTF-8
```
jem-davies: Could you check if it is valid UTF-8 with: https://pkg.go.dev/unicode/utf8#Valid ?

triddell:
Thanks for the review! I've addressed all three items:

## Changes Made

### 1. Loop Modernization ✅
Replaced `for i := 0; i < n; i++` with `for i := range n` throughout:
- [bloblang.go:59](internal/impl/parquet/bloblang.go#L59) - strict path (current/default behavior)
- [bloblang.go:85](internal/impl/parquet/bloblang.go#L85) - lenient path
- [processor_decode.go:141](internal/impl/parquet/processor_decode.go#L141) - strict path (current/default behavior)
- [processor_decode.go:169](internal/impl/parquet/processor_decode.go#L169) - lenient path

### 2. Resource Management ✅
Changed from deferred `rows.Close()` to explicit close at end of each row group iteration to avoid accumulating open resources:
- [bloblang.go:73-89](internal/impl/parquet/bloblang.go#L73-L89) - lenient path only
- [processor_decode.go:157-181](internal/impl/parquet/processor_decode.go#L157-L181) - lenient path only

Also added explicit `rows.Close()` on error paths before returning. Note: This only affects the lenient path (`strict_schema: false`) since the strict path doesn't iterate over row groups.

### 3. UTF-8 Validation ✅
Implemented proper UTF-8 validation in [common.go:90-94](internal/impl/parquet/common.go#L90-L94) using `unicode/utf8.Valid()`:
- ByteArray values return as `string` if valid UTF-8
- ByteArray values return as `[]byte` if not valid UTF-8

This only affects the lenient path since `convertSingleValue` is only called by `readLenient`.

All changes build successfully.

PR #617, opened 2025-12-15
couchbase: add output

sapk:
This PR add output for couchbase. It re-use as much existing code. 
It add support for:
- partial fail
- batching and max_in_flight logic.

I was previously using the reject processor with couchbase processor to use as an output but it had the disadvantage of not batching the write.

This is cleaner.

### DIFF
--- a/internal/impl/couchbase/output.go
+++ b/internal/impl/couchbase/output.go
@@ -12,7 +12,7 @@ import (
 func OutputConfig() *service.ConfigSpec {
 	return service.NewConfigSpec().
 		// TODO Stable().
-		Version("1.14.0").
+		Version("1.15.0").
 		Categories("Integration").
 		Summary("Performs operations against Couchbase for each message, allowing you to store data within message payloads.").
 		Description("When inserting, replacing or upserting documents, each must have the `content` property set.\n\n### Concurrent Document Mutations\nTo prevent read/write conflicts, Couchbase returns a [_Compare And Swap_ (CAS)](https://docs.couchbase.com/go-sdk/current/howtos/concurrent-document-mutations.html) value with each accessed document. Bento stores these as key/value pairs in metadata with the `couchbase_cas` field. Note: CAS checks are enabled by default. You can configure this by changing the value of `cas_enabled: false`.").
--- a/website/docs/components/outputs/couchbase.md
+++ b/website/docs/components/outputs/couchbase.md
@@ -20,7 +20,7 @@ This component is experimental and therefore subject to change or removal outsid
 :::
 Performs operations against Couchbase for each message, allowing you to store data within message payloads.
 
-Introduced in version 1.14.0.
+Introduced in version 1.15.0.
 
 
 <Tabs defaultValue="common" values={[


internal/impl/couchbase/couchbase.go
```diff
@@ -1,146 +1,201 @@
...
+	ErrInvalidOperation = errors.New("invalid operation")
+	// ErrContentRequired content field is required.
+	ErrContentRequired = errors.New("content required")
+	// ErrBatchError batch error.
+	ErrBatchError = errors.New("batch error")
+)
 
-	return nil, gocb.Cas(0), errors.New("type not supported")
+// Config export couchbase processor specification.
+func Config() *service.ConfigSpec {
```
gregfurman: Unsure that a common description/summary makes sense to share between input, processor, and output components.

I think you should rather return `[]*service.ConfigFIeld` that can be passed to a `service.ConfigSpec` via the `Fields` method. See how this was done in mongodb:

https://github.com/warpstreamlabs/bento/blob/ac10c9066920efbcbde8167ff1ab2752cb630ef7/internal/impl/mongodb/common.go#L39-L54 

and passed to `Fields`:

https://github.com/warpstreamlabs/bento/blob/ac10c9066920efbcbde8167ff1ab2752cb630ef7/internal/impl/mongodb/cache.go#L17-L22

internal/impl/couchbase/operations.go
gregfurman, line 1: nit: I think `utils.go` or `convert.go` would be more appropriate than `tools.go`

internal/impl/couchbase/couchbase.go
```diff
@@ -1,146 +1,201 @@
...
 	}
 
-	if ttl != nil {
-		op.Expiry = *ttl
+	return ops, nil
+}
+
+// Connect connects to the couchbase cluster
+func (c *Couchbase) Connect(_ context.Context) (err error) {
+	c.couchbaseClient, err = getClient(c.conf, c.mgr)
```
gregfurman: Should we not be passing this `ctx` to the `getClient` to populate those `WaitUntilReady`options?

https://github.com/warpstreamlabs/bento/blob/ac10c9066920efbcbde8167ff1ab2752cb630ef7/internal/impl/couchbase/client.go#L94-L97

I'm concerned this `timeout` being too large will block the pipeline closing if the `Connect` context is cancelled early

i.e

```golang
	err = cluster.Bucket(bucket).WaitUntilReady(timeout, &gocb.WaitUntilReadyOptions{Context: ctx})

```

internal/impl/couchbase/couchbase.go
```diff
@@ -1,146 +1,201 @@
...
+		p.op = increment
+	case client.OperationDecrement:
+		p.op = decrement
+	default:
+		return nil, fmt.Errorf("%w: %s", ErrInvalidOperation, op)
 	}
 
-	if ttl != nil {
-		op.Expiry = *ttl
+	if !p.outputMode {
```
gregfurman, line 118: Just pointing out that this will always be false since I can't see it ever being set (have commented above).

internal/impl/couchbase/couchbase.go
```diff
@@ -1,146 +1,201 @@
...
-	}
+// New returns a Couchbase instance.
+func New(conf *service.ParsedConfig, mgr *service.Resources, outputMode bool) (*Couchbase, error) {
+	var err error
 
-	if ttl != nil {
-		op.Expiry = *ttl
+	p := &Couchbase{
+		conf: conf,
+		mgr:  mgr,
```
gregfurman: I think you meant to set this `outputMode` attribute here since I can't see it ever being set elsewhere:

```suggestion
	p := &Couchbase{
		conf: conf,
		mgr:  mgr,
		outputMode: outputMode,
```

gregfurman (changes requested):

internal/impl/couchbase/couchbase.go
sapk: Thanks for the catch and sorry for the mistake. It must has passed under the test radar because output mode is just to skip useless steps since connect is called in a dedicated method for output and there is not need to update the message. I will fix this when I have access to my pc.

internal/impl/couchbase/couchbase.go
sapk: I didn't passed the context because we don't have one to pass in processor setup. But we could pass a nil one without any issue for processor setup. Good idea.

internal/impl/couchbase/operations.go
sapk: After some thought, I renamed it to `operations.go` as it is helpers to convert to gocb.BulkOp interface ("generic" couchbase operation). Let me know if that's ok.

internal/impl/couchbase/couchbase.go
sapk: done

sapk:
I re-run the integrations and run the build with a valid config and everything still works fine. So it is good for review.

PR #621, opened 2025-12-18
feat: Add WASM-based python processor

gregfurman:
## Motivation

Includes a new `python` processor that runs within a sandboxed WASM environment.

## Changes

- Adds a new `python` processor that runs within a sandboxed WASM environment.
- Adds a `WasmModulePool` in `internal/impl/wasm/wasmpool/pool.go` that uses a Golang `sync.Pool` to execute multiple concurrent guests and have their lifecycles managed by the GC. This allows us to spin up more instances as demand increases and have the Go GC handle the guest lifecycle.
- Adds an `entrypoint.py` that allows python execution `script` to compile and execute -- `stdout` and `stdin` is used to pass data between the guest and host.
- If the `x_wasm` tag is used, the WASM executable and corresponding `entrypoint.py` are embedded at compile time. Using the processor without this tag panics on init.
- Adds a script at `scripts/install.sh` that downloads the `python-3.12.0.wasm` and checks it against the SHA in `runtime/python-3.12.0.wasm.sha256sum`.

## Installation/Usage

Run the following commands to create a config, install the python runtime (and check it against the SHA), and execute a pipeline.

```sh
cat > python.yaml <<'EOF'
pipeline:
  processors:
    - python:
        script: |
          root = sum(this)
EOF

bash internal/impl/python/scripts/install.sh

echo "[11, 6, 20, 5]" | go run -tags "x_wasm" cmd/bento/main.go -c python.yaml --set logger.level=warn
```

The output of the pipeline should be `42`.

## TODO

- [x] Extend the CI (or Makefile) to run the downloader script and build with the `x_wasm` tag -- allowing the `.wasm` to be embedded at compile time and used in the processor.

### DIFF
--- a/internal/impl/python/processor.go
+++ b/internal/impl/python/processor.go
@@ -44,7 +44,6 @@ func init() {
 
 func pythonProcessorSpec() *service.ConfigSpec {
 	return service.NewConfigSpec().
-		Stable().
 		Categories("Mapping").
 		Summary(`
 Executes a Python script against each message in a stream.
--- a/website/docs/components/processors/python.md
+++ b/website/docs/components/processors/python.md
@@ -2,7 +2,7 @@
 title: python
 slug: python
 type: processor
-status: stable
+status: experimental
 categories: ["Mapping"]
 ---
 
@@ -15,6 +15,9 @@ categories: ["Mapping"]
 import Tabs from '@theme/Tabs';
 import TabItem from '@theme/TabItem';
 
+:::caution EXPERIMENTAL
+This component is experimental and therefore subject to change or removal outside of major version releases.
+:::
 
 Executes a Python script against each message in a stream.
 


aronchick:
this is going to be really cool! lmk if you need help

gregfurman:
Thanks @aronchick! If you have time (and are feeling extra generous), I'd appreciate if you could run through the `Installation/Usage` steps locally and try out different configurations. I'm also looking for any feedback on usability since we're planning to expand our WASM offering, so any insights would be super helpful!

I haven't dug too deep into the limitations of the sandboxed runtime and how it responds to accessing non-permissive resources (e.g., filesystem or OS). 

In addition, we'd probably need to soak test the processor before making this GA in Bento since I'm unsure of performance/behavior over longer periods and under differing load profiles.

internal/impl/python/runtime/entrypoint.py
```diff
@@ -0,0 +1,89 @@
...
+        header = sys.stdin.buffer.read(4)
+        if not header: break
+        length = struct.unpack(">I", header)[0]
+        payload = sys.stdin.buffer.read(length)
+
+        status = _SUCCESS_STATUS
+        try:
+            input_data = _get_input(payload)
+            namespace = {"this": input_data, "root": None, "__builtins__": __builtins__, **injected_libs}
+            exec(compiled_code, namespace)
```
gregfurman, line 60: this `namespace` and `exec` combintation allows us to dynamically run our compiled execution script, constraining what is accessible during runtime and ensuring no global state is ever mutated.

Also, see how we pass `this` and `root` -- leveraging the same pattern we use with bloblang scripts/mappings.

internal/impl/python/runtime/entrypoint.py
```diff
@@ -0,0 +1,89 @@
...
+
+def _load_extra_libs(extra_libs: list[str]):
+    injected = {}
+    if not extra_libs:
+        return injected
+    for lib_name in extra_libs:
+        name = lib_name.strip()
+        if name:
+            try:
+                injected[name] = importlib.import_module(name)
```
gregfurman, line 20: Some python magic that allows us to dynamically extract which libraries we're making permissable to the sandboxed execution.

internal/impl/python/scripts/install.sh
```diff
@@ -0,0 +1,17 @@
...
+
+WASM_RUNTIME_DIR="$(dirname "${BASH_SOURCE[0]}")/.."
+WASM_BINARY_URL="https://github.com/vmware-labs/webassembly-language-runtimes/releases/latest/download/python-3.12.0.wasm"
+WASM_PATH="$WASM_RUNTIME_DIR/runtime/python-3.12.0.wasm"
+
+wget -q "$WASM_BINARY_URL" -O "$WASM_PATH"
+
+if command -v sha256sum >/dev/null; then
+    (cd "$WASM_RUNTIME_DIR/runtime" && sha256sum --check --status python-3.12.0.wasm.sha256sum)
+elif command -v shasum >/dev/null; then
```
gregfurman: fallback to `shasum` since `sha256sum` is not available on darwin

internal/impl/python/scripts/install.sh
```diff
@@ -0,0 +1,17 @@
...
+set -e
+
+WASM_RUNTIME_DIR="$(dirname "${BASH_SOURCE[0]}")/.."
+WASM_BINARY_URL="https://github.com/vmware-labs/webassembly-language-runtimes/releases/latest/download/python-3.12.0.wasm"
+WASM_PATH="$WASM_RUNTIME_DIR/runtime/python-3.12.0.wasm"
+
+wget -q "$WASM_BINARY_URL" -O "$WASM_PATH"
+
+if command -v sha256sum >/dev/null; then
+    (cd "$WASM_RUNTIME_DIR/runtime" && sha256sum --check --status python-3.12.0.wasm.sha256sum)
```
gregfurman: Since we're downloading and embedding a WASM executable into bento, we need to verify the integrity of the executable we're intending to use. So we download and commit the SHA256 checksum, then verify the downloaded file's checksum.

internal/impl/wasm/wasmpool/pool.go
```diff
@@ -0,0 +1,77 @@
...
+)
+
+type constructor[T api.Module] func(context.Context) (T, error)
+
+type WasmModulePool[T api.Module] struct {
+	ctor         constructor[T]
+	instancePool sync.Pool
+}
+
+func NewWasmModulePool[T api.Module](ctx context.Context, ctor constructor[T]) (*WasmModulePool[T], error) {
```
gregfurman, line 20: Unfortunately WASM modules (at least with `wazero`) are not safe for concurrent executions. Instead, we create a module pool (backed by a `sync.Pool`) and allow the GC to dynamically spin up instances depending on load.

internal/impl/wasm/wasmpool/pool.go
```diff
@@ -0,0 +1,77 @@
...
+func (p *WasmModulePool[T]) Get(ctx context.Context) (T, error) {
+	pooled := p.instancePool.Get()
+	if pooled == nil {
+		return p.newInstance(ctx)
+	}
+	return pooled.(T), nil
+}
+
+func (p *WasmModulePool[T]) Put(instance T) {
+	_ = runtime.AddCleanup(&instance, func(inst T) {
```
gregfurman, line 55: If the GC cleans up the instance, which is possible given `sync.Pool`, we need to ensure the resources are actually freed up. Hence, this gross `runtime.AddCleanup` hook that triggers when the module is GC'd

internal/impl/wasm/wasmpool/pool.go
```diff
@@ -0,0 +1,77 @@
...
+		inst.Close(context.Background())
+	}, instance)
+
+	p.instancePool.Put(instance)
+}
+
+func (p *WasmModulePool[T]) Close(ctx context.Context) error {
```
gregfurman, line 62: While it's probably easier to just close the entire runtime to ensure all associated modules (and resources) are cleaned up, this `Close` will retrieve all objects from the pool and close them. 

While thinking of this as a `Reset` or `Cleanup` conceptually more correct, `Close` feels more idiomatic 🤷

internal/impl/python/instance.go
```diff
@@ -0,0 +1,126 @@
...
+		// this is a blocking call, so we have to deploy in a goroutine
+		mod, err = p.runtime.InstantiateModule(ctx, p.compiled, cfg)
+		if err != nil {
+			instantiateErr <- fmt.Errorf("instantiation failed: %w", err)
+			outW.Close()
+		}
+	}()
+
+	handshakeErr := make(chan error, 1)
+	go func() {
```
gregfurman, line 47: We do a quick-and-dirty handshake to ensure the python env is up-and-running. The python module should write a `"READY"` signal to the output socket which we need to check is received before proceeding.

internal/impl/python/instance.go
```diff
@@ -0,0 +1,126 @@
...
+
+func (pi *pythonInstance) runRequest(input []byte) (stdout []byte, stderr []byte, err error) {
+	if err := binary.Write(pi.stdinW, binary.BigEndian, uint32(len(input))); err != nil {
+		return nil, pi.stderrBuf.Bytes(), err
+	}
+	if _, err := pi.stdinW.Write(input); err != nil {
+		return nil, pi.stderrBuf.Bytes(), err
+	}
+
+	header := make([]byte, 5)
```
gregfurman, line 100: First byte signals that the response was a success or exception. Last four bytes tell us how long the response is.

aronchick:
## Testing Instructions

I've tested this locally and it works great! Here's how others can try it:

### Quick Start

```bash
# 1. Checkout the PR
gh pr checkout 621

# 2. Download the Python WASM runtime (~50MB) and test configs
bash internal/impl/python/scripts/install.sh
gh gist clone https://gist.github.com/aronchick/5d9391b03f9266b994c7a36043d743a4 python-examples

# 3. Run an example (first build takes ~30-60s)
go run -tags "x_wasm" ./cmd/bento/main.go -c python-examples/hello.yaml
```

Expected output: `"Hello, World!"`

### Example Configs

Full test configs available in the gist: **<GIST_URL>**

| File | Description |
|------|-------------|
| `hello.yaml` | Simple greeting |
| `test1.yaml` | JSON transformation |
| `test2.yaml` | Array processing (sum, count, double) |
| `test3.yaml` | Message filtering |
| `test4.yaml` | Stdlib imports (math, json, re) |
| `test5.yaml` | E-commerce order transformation |
| `test6.yaml` | Error handling demo |

### Quick Snippets

**Array processing:**
```yaml
- python:
    script: |
      root = {"sum": sum(this), "doubled": [x * 2 for x in this]}
```

**With imports:**
```yaml
- python:
    imports: [math, json, re]
    script: |
      root = {"ceil": math.ceil(this["value"])}
```

**Filtering (set root=None to drop):**
```yaml
- python:
    script: |
      root = this if this.get("status") == "active" else None
```

### Run Unit Tests

```bash
go test -tags "x_wasm" -v ./internal/impl/python/...
```

All 18 tests pass (~18 seconds).

### Key Points

- **Build tag required:** Always use `-tags "x_wasm"`
- **Cold start:** ~1 second per Python instance
- **Familiar syntax:** Uses `this` and `root` like Bloblang
- **Sandboxed:** Runs in WASM, no filesystem/network access

jem-davies:
Question, 

I see you have a TODO: 

> Extend the CI (or Makefile) to run the downloader script and build with the x_wasm tag -- allowing the .wasm to be embedded at compile time and used in the processor.

Is the plan then that for the binaries we publish to the [release page](https://github.com/warpstreamlabs/bento/releases) be embedded with the `.wasm`? 

Just trying to put myself in place of a user at the moment and am seeing that I need to follow instructions on the PR to get the processor to work - but they aren't on the doc page for this new processor?

gregfurman:
> Question, 
> 
> 
> 
> I see you have a TODO: 
> 
> 
> 
> > Extend the CI (or Makefile) to run the downloader script and build with the x_wasm tag -- allowing the .wasm to be embedded at compile time and used in the processor.
> 
> 
> 
> Is the plan then that for the binaries we publish to the [release page](https://github.com/warpstreamlabs/bento/releases) be embedded with the `.wasm`? 
> 
> 
> 
> Just trying to put myself in place of a user at the moment and am seeing that I need to follow instructions on the PR to get the processor to work - but they aren't on the doc page for this new processor? 
> 
> 

Thanks for giving this a look!

Yeah so the plan is to embed this binary inside the executable. We can also zip the bin with brotli (which makes it much smaller) and just unzip at runtime if the processor is used.

Wouldn't expect a user to run any of this. 

Ideally would want to drop the x_wasm tag and allow this in the standard built.

Thoughts?

jem-davies:
> Ideally would want to drop the x_wasm tag and allow this in the standard built.

Yeah, that is what I was thinking 👍

gregfurman:
@jem-davies OK have updated this to use `make generate` with an associated downloader script in `runtime/install.go`. 

This also allows the downloading to be compatible with windows builds since the `.sh` has been replaced with a go script that both retrieves and validates the binary.

We just need to add `make generate` to the goreleaser process to ensure the `.wasm` is present 👍 

Question though: are we OK with the binary being downloaded each time we need to run a CI job? I see linting is (rightfully)  failing now since the `embed` directive cannot find the non-existent file `.wasm` and `entrypoint.py` files

internal/impl/python/processor.go
```diff
@@ -0,0 +1,202 @@
...
+			return newPythonProcessor(conf, mgr)
+		})
+	if err != nil {
+		panic(err)
+	}
+}
+
+func pythonProcessorSpec() *service.ConfigSpec {
+	return service.NewConfigSpec().
+		Stable().
```
jem-davies: Thoughts about downgrading to 'experimental' and then getting this merged?

gregfurman:
@jem-davies OK will merge! 

FYI Pydantic recently released https://github.com/pydantic/monty which is basically a python subset written in Rust. I ran some tests locally and managed to compile it to a WASM bundle and it's only like 4mb uncompressed. Will look into replacing this VMWare binary at some point with it instead.

Tbh I'm not the biggest fan of embedding a zip into the go binary but 🤷

PR #624, opened 2025-12-20
feat(kubernetes): add kubernetes_watch input for real-time resource monitoring

aronchick:
## Summary

Add a new `kubernetes_watch` input that uses the Kubernetes Watch API to stream real-time events for cluster resources. This enables event-driven pipelines that react to Kubernetes state changes.

### Features
- Watch any resource type via dynamic discovery (pods, deployments, CRDs, etc.)
- Per-namespace or cluster-wide monitoring
- Label and field selector filtering
- Automatic reconnection with exponential backoff
- Proper handling of 410 Gone errors (expected watch API behavior)

### Implementation Highlights
- Uses `DeferredDiscoveryRESTMapper` for dynamic GVR resolution (no static resource list)
- Supports in-cluster auth, kubeconfig, or explicit credentials
- Resource version tracking for resumable watches

### Example Configuration

```yaml
input:
  kubernetes_watch:
    resource: pods
    namespaces:
      - default
    label_selector: "app=myapp"
```

## Test Plan
- [x] Unit tests for config parsing
- [x] Unit tests for auth configuration  
- [x] Lint passes (0 issues)
- [x] All tests pass
- [ ] Manual testing with local cluster

## Files Changed (~1,060 lines)
- `internal/impl/kubernetes/` - New kubernetes input package
- `go.mod`, `go.sum` - k8s client-go dependencies


aronchick:
I've built some simple testing scripts as well.

https://github.com/aronchick/bento/tree/kubernetes-integration-tests/testing/kubernetes

internal/impl/kubernetes/auth.go
```diff
@@ -0,0 +1,258 @@
...
+	}
+
+	// Token authentication
+	token, _ := conf.FieldString("token")
+	tokenFile, _ := conf.FieldString("token_file")
+
+	if token != "" {
+		config.BearerToken = token
+	} else if tokenFile != "" {
+		tokenBytes, err := os.ReadFile(tokenFile)
```
gregfurman: Could you rather use `ifs` instead of `os` for file reading? See example file reading here:

https://github.com/warpstreamlabs/bento/blob/ac10c9066920efbcbde8167ff1ab2752cb630ef7/internal/impl/io/cache_file.go#L61

internal/impl/kubernetes/auth.go
```diff
@@ -0,0 +1,258 @@
...
+	}
+
+	return config, nil
+}
+
+// InClusterNamespace returns the namespace this pod is running in,
+// or "default" if not running in a cluster.
+func InClusterNamespace() string {
+	// Try to read the namespace from the service account
+	nsBytes, err := os.ReadFile("/var/run/secrets/kubernetes.io/serviceaccount/namespace")
```
gregfurman: Is it a fair assumption that bento will have access to this? Also, will this location change between darwin and linux distros? 

Perhaps we add a build-tag to exclude this procesor from `windows` builds

internal/impl/kubernetes/config.go
```diff
@@ -0,0 +1,51 @@
...
+
+// CommonFields returns config fields shared across all Kubernetes inputs.
+func CommonFields() []*service.ConfigField {
+	return []*service.ConfigField{
+		service.NewStringListField("namespaces").
+			Description("Namespaces to watch. Empty list means all namespaces.").
+			Default([]any{}).
+			Example([]string{"default"}).
+			Example([]string{"production", "staging"}),
+		service.NewStringField("label_selector").
```
gregfurman: Why not we make this a string map instead? See below example:

https://github.com/warpstreamlabs/bento/blob/ac10c9066920efbcbde8167ff1ab2752cb630ef7/internal/impl/gcp/input_bigquery_select.go#L97

I think it's more ergonomic to pass in a map of kvs in YAML instead of a comma delimited list i.e

```yaml
label_selector:
  app: myapp
  env: prod
```

This way, we can also dynamically set different fields using environment variables:

```yaml
label_selector:
  app: myapp
  env: ${K8S_ENVIRONMENT:dev}
```

internal/impl/kubernetes/config.go
```diff
@@ -0,0 +1,51 @@
...
+			Description("Kubernetes label selector to filter resources (e.g., 'app=myapp,env=prod').").
+			Default("").
+			Example("app=myapp").
+			Example("app in (frontend,backend)"),
+		service.NewStringField("field_selector").
```
gregfurman: same comment as above re: maps ☝️

internal/impl/kubernetes/auth.go
```diff
@@ -0,0 +1,258 @@
...
+		for _, p := range paths {
+			if p == "" {
+				continue
+			}
+			expanded = append(expanded, expandHomePath(p))
+		}
+		if len(expanded) > 0 {
+			loadingRules.Precedence = expanded
+		}
+	}
```
gregfurman, line 211: q: What about passing in a kubeconfig file as bytes and unmarshalling it at runtime?

Think you can use the[ `genericclioptions` package](https://pkg.go.dev/k8s.io/cli-runtime/pkg/genericclioptions) for this (from [S/O](https://stackoverflow.com/a/75733041/11257137)):

```golang
   kubeconfigData := []byte("my_kubeconfig")

    configOverrides := &genericclioptions.ConfigOverrides{ClusterDefaults: genericclioptions.NewClusterDefaults()}
    configLoader := genericclioptions.NewConfigFlags(false).WithOverrideFlags(configOverrides)
    configLoader.KubeConfig = bytes.NewBuffer(kubeconfigData)
```

internal/impl/kubernetes/input_watch.go
```diff
@@ -0,0 +1,476 @@
...
+			"kubernetes_resource_creation_timestamp",
+		})).
+		Fields(AuthFields()...).
+		Fields(CommonFields()...).
+		Field(service.NewStringField("resource").
+			Description("Standard Kubernetes resource type to watch.").
+			Default("").
+			Example("pods").
+			Example("deployments").
+			Example("configmaps")).
```
gregfurman, line 87: nit: Should we not allow for watching multiple resources here? Unsure if the K8s api supports this but perhaps we could multiplex the ingestion into a single stream 🤔

internal/impl/kubernetes/input_watch.go
```diff
@@ -0,0 +1,476 @@
...
+				Description("API version for the custom resource (e.g., 'v1', 'v1beta1').").
+				Default("v1"),
+			service.NewStringField("resource").
+				Description("Plural name of the custom resource (e.g., 'crontabs').").
+				Default(""),
+		).
+			Description("Custom Resource Definition to watch. Use this for CRDs instead of 'resource'.").
+			Optional().
+			Advanced()).
+		Field(service.NewStringListField("event_types").
```
gregfurman, line 102: q: If this isn't supported at the API level (i.e you can determine exactly what types of events you want to filter out) I'm unsure it makes too much sense to support this explicitly in the component. 

Like, you can always use bloblang to filter out those records that are `DELETED` if you only care about those that are of type `ADDED`.

internal/impl/kubernetes/input_watch.go
```diff
@@ -0,0 +1,476 @@
...
+			Description("Watch event types to include.").
+			Default([]any{"ADDED", "MODIFIED", "DELETED"}).
+			Example([]string{"ADDED", "DELETED"}).
+			Example([]string{"MODIFIED"})).
```
gregfurman: Perhaps we should consider an alternative approach that allows us to constrain what types are added here i.e

```golang
		Field(service.NewObjectListField("event_types",
			service.NewStringAnnotatedEnumField("", map[string]string{
				string(watch.Added):    "Add events.",
				string(watch.Deleted):  "Delete events.",
				string(watch.Modified): "Modified events.",
			}).Description("List of operations to watch.").Default(string(watch.Added)),
		)).
```

Otherwise, think this would be a nice use-case for adding a `NewStringEnumField` type 🤔

internal/impl/kubernetes/input_watch.go
```diff
@@ -0,0 +1,476 @@
...
+		return nil, fmt.Errorf("failed to parse request_timeout: %w", err)
+	}
+
+	// Get Kubernetes client (needed for RESTMapper)
+	if k.clientSet, err = GetClientSet(context.Background(), conf); err != nil {
+		return nil, fmt.Errorf("failed to create kubernetes client: %w", err)
+	}
+
+	// Determine GVR from resource or custom_resource
+	resource, _ := conf.FieldString("resource")
```
gregfurman, line 221: What are your thoughts on adding a `custom_resources` field i.e

```golang
		Field(service.NewObjectListField("custom_resource_definitions",
			service.NewStringField("resource").Description("The resource name."),
			service.NewStringField("group").Description("Name of the CRD group."),
			service.NewStringField("version").Description("The CRD version."),
		)).Description("A list of custom resource definition objects").Optional().
```

And we can directly reference this in `resource` i.e

```yaml
custom_resources:
  - resource: foo
    group: bar
    version: buz
resource: foo # ref the CRD we defined above
```

Also, I already mentioned above, but should we be allowing for multple resources to be watched at a time?

internal/impl/kubernetes/input_watch.go
```diff
@@ -0,0 +1,476 @@
...
+			watchCancel()
+			// If the error is a "410 Gone" error, it means the resource version is too old.
+			// Reset the resource version to force a fresh list on the next loop.
+			if serr, ok := err.(*k8serrors.StatusError); ok && serr.ErrStatus.Code == http.StatusGone {
+				k.log.Warnf("Watch for %s in namespace %s returned 410 Gone, resetting resource version", k.gvr.Resource, namespace)
+				k.setResourceVersion(namespace, "")
+			} else {
+				k.log.Errorf("Failed to watch %s in namespace %s: %v", k.gvr.Resource, namespace, err)
+			}
+			backoff := calculateBackoff(retryAttempt)
```
gregfurman: suggestion: Perhaps give a look at our internal `CommonRetryBackOffFields and `CommonRetryBackOffCtorFromParsed` functions in the `retries` package. You can define the boff retrier in config and extract an object very easily:

https://github.com/warpstreamlabs/bento/blob/ac10c9066920efbcbde8167ff1ab2752cb630ef7/internal/retries/retries.go#L19-L24

https://github.com/warpstreamlabs/bento/blob/ac10c9066920efbcbde8167ff1ab2752cb630ef7/internal/retries/retries.go#L53

Usage:

https://github.com/warpstreamlabs/bento/blob/ac10c9066920efbcbde8167ff1ab2752cb630ef7/internal/impl/aws/output_sqs.go#L73-L75

internal/impl/kubernetes/input_watch.go
```diff
@@ -0,0 +1,476 @@
...
+				select {
+				case k.eventChan <- watchEvent{eventType: eventType, object: obj}:
+				case <-k.shutSig.SoftStopChan():
+					return
+				}
+
+			case watch.Bookmark:
+				if obj, ok := event.Object.(*unstructured.Unstructured); ok {
+					k.setResourceVersion(namespace, obj.GetResourceVersion())
+				}
```
gregfurman: q: Does the component support `Bookmark` types?

internal/impl/kubernetes/package.go
```diff
@@ -0,0 +1,54 @@
...
+//   - Pod log streaming (kubernetes_logs)
+//   - Cluster event monitoring (kubernetes_events)
+//   - Resource state changes (kubernetes_watch)
+//   - Prometheus metrics scraping (kubernetes_metrics)
+//
+// All components share a common authentication layer supporting:
+//   - Automatic in-cluster detection
+//   - Kubeconfig file authentication
+//   - Service account tokens
+//   - Explicit credentials
```
gregfurman: We don't normally define package level documentation for unexported/internal packages

```suggestion
```

internal/impl/kubernetes/package.go
```diff
@@ -0,0 +1,54 @@
...
+
+const (
+	// Backoff parameters for watch reconnection
+	minBackoff = 1 * time.Second
+	maxBackoff = 60 * time.Second
+)
+
+// keyCache optimizes metadata key generation by interning strings
+// like "kubernetes_labels_app" to avoid repeated allocations.
+var keyCache sync.Map
```
gregfurman: If I define multiple watchers, is it deseriable for this to be shared?

Also, do we ever clear the `keyCache`? Concerned this is a risk for a memory leak of some kind.

internal/impl/kubernetes/input_watch.go
```diff
@@ -0,0 +1,476 @@
...
+
+func (k *kubernetesWatchInput) listExistingResources(resourceInterface dynamic.ResourceInterface, namespace string) {
+	listOpts := metav1.ListOptions{
+		LabelSelector: k.labelSelector,
+		FieldSelector: k.fieldSelector,
+		Limit:         500, // Process in chunks of 500
+	}
+
+	for {
+		listCtx, cancel := requestContext(k.requestTimeout, k.shutSig.SoftStopChan())
```
gregfurman: You can replace this function usage by deriving a context from the shutdown signaller that is cancelled either if the timeout is reached or the shutdown signal is called:

```suggestion
		ctx, cancel := context.WithTimeout(context.Background(), k.requestTimeout)
		listCtx, _ := k.shutSig.SoftStopCtx(ctx)
```

internal/impl/kubernetes/input_watch.go
```diff
@@ -0,0 +1,476 @@
...
+		list, err := resourceInterface.List(listCtx, listOpts)
+		cancel()
```
gregfurman: nit: We should do a `defer cancel()` after context creation to avoid leaks.

internal/impl/kubernetes/input_watch.go
```diff
@@ -0,0 +1,476 @@
...
+				return
+			}
+		}
+	}
+}
+
+func (k *kubernetesWatchInput) Read(ctx context.Context) (*service.Message, service.AckFunc, error) {
+	select {
+	case event := <-k.eventChan:
+		return k.eventToMessage(event)
```
gregfurman: We should check if the channel is closed:

```suggestion
	case event, ok := <-k.eventChan:
		if !ok {
			return nil, nil, service.ErrEndOfInput
		}
		return k.eventToMessage(event)
```

internal/impl/kubernetes/input_watch.go
```diff
@@ -0,0 +1,476 @@
...
+	case <-ctx.Done():
+		return nil, nil, ctx.Err()
+	case <-k.shutSig.SoftStopChan():
+		return nil, nil, service.ErrEndOfInput
+	}
+}
+
+func (k *kubernetesWatchInput) eventToMessage(event watchEvent) (*service.Message, service.AckFunc, error) {
```
gregfurman, line 468: praise: this is super neat!

internal/impl/kubernetes/input_watch.go
```diff
@@ -0,0 +1,476 @@
...
+	return msg, func(ctx context.Context, err error) error {
+		return nil // Watch events don't require acknowledgment
+	}, nil
+}
+
+func (k *kubernetesWatchInput) Close(ctx context.Context) error {
+	k.shutSig.TriggerSoftStop()
+	k.shutSig.TriggerHasStopped()
+	return nil
+}
```
gregfurman, line 523: We should be closing the `k.eventchan` channel here and setting it to `nil` so no further reads can occur.

Also, you're not using the shutdown signallers `HasStoppedChan` or `HasStoppedCtx` so the extra call to trigger the `HasStopped` signal is unnecessary.

FYI, you'd use the `TriggerHasStopped` if you were draining the channel or waiting for resources to close, and the component can only really be marked as closed once that has occurred.

internal/impl/kubernetes/input_watch_test.go
```diff
@@ -0,0 +1,76 @@
+package kubernetes
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestKubernetesWatchConfigParse(t *testing.T) {
```
gregfurman, line 16: suggestion: I wonder if we could write some integration tests and use k3d to actually simulate the component watching events in k8s 🤔 Else, if there is an in-memory cluster we can use for testing that would also be brilliant.

aronchick: this is exactly what i used locally to test! https://github.com/aronchick/bento/tree/kubernetes-integration-tests/testing/kubernetes

internal/impl/kubernetes/input_watch.go
aronchick: i have lots more stuff :) i just wanted to keep this PR small.

I have a series of follow-up PRs planned that build on this foundation:

1. kubernetes_events - Stream Kubernetes cluster events (the Event resource type, useful for monitoring cluster health/alerts)
2. kubernetes_logs - Stream pod logs (similar to kubectl logs -f)
3. kubernetes_metrics - Pull metrics from metrics-server

For watching multiple resource types in a single pipeline, users can use Bento's built-in broker input to combine multiple kubernetes_watch inputs, this keeps each watcher simple and composable rather than adding multiplexing complexity to the component itself.

Happy to discuss the multi-resource approach more, but my instinct is that broker handles this cleanly and keeps the component focused.

internal/impl/kubernetes/input_watch.go
aronchick: Yeah, you're not wrong - tried to expand a bit here. My understanding is you CANNOT filter upstream.

internal/impl/kubernetes/auth.go
aronchick: Good suggestion! i'm not IN LOVE with two fields, but put them both in because it is convenient

internal/impl/kubernetes/input_watch.go
aronchick: for sure! i wanted to keep this PR small - i have three more PRs waiting (one will explicitly handle CRDs).

internal/impl/kubernetes/input_watch.go
aronchick: Let's drop it (for now)

internal/impl/kubernetes/input_watch.go
aronchick: thanks!!

internal/impl/kubernetes/input_watch.go
aronchick: fixed

internal/impl/kubernetes/package.go
aronchick: Fixed!

internal/impl/kubernetes/auth.go
aronchick: fixed

internal/impl/kubernetes/auth.go
aronchick: done

internal/impl/kubernetes/config.go
aronchick: Done

internal/impl/kubernetes/input_watch.go
aronchick: Done

internal/impl/kubernetes/input_watch.go
aronchick: done

internal/impl/kubernetes/config.go
aronchick: done

internal/impl/kubernetes/input_watch.go
aronchick: updated!

internal/impl/kubernetes/package.go
aronchick: done!

internal/impl/kubernetes/input_watch.go
aronchick: done!

aronchick:
@gregfurman I've addressed your feedback - think i got to everything. Just LMK if you need anything else!

internal/impl/io/input_file.go
```diff
@@ -130,7 +130,7 @@ func (f *fileConsumer) Connect(ctx context.Context) error {
 	return nil
 }
 
-func (f *fileConsumer) getReader(ctx context.Context) (scannerInfo, error) {
+func (f *fileConsumer) getReader(_ context.Context) (scannerInfo, error) {
```
gregfurman: Can we revert these back? I'd rather the diff not get too large here

```suggestion
func (f *fileConsumer) getReader(ctx context.Context) (scannerInfo, error) {
```

internal/impl/io/metrics_json_api.go
```diff
@@ -31,7 +31,7 @@ type jsonAPIMetrics struct {
 	timestamp time.Time
 }
 
-func newJSONAPI(logger *service.Logger) (*jsonAPIMetrics, error) {
+func newJSONAPI(_ *service.Logger) (*jsonAPIMetrics, error) {
```
gregfurman: Same as above regarding diff:

```suggestion
func newJSONAPI(logger *service.Logger) (*jsonAPIMetrics, error) {
```

internal/impl/io/processor_command.go
```diff
@@ -87,7 +87,7 @@ type commandProc struct {
 	argsMapping *bloblang.Executor
 }
 
-func newCommandProcFromParsed(conf *service.ParsedConfig, mgr *service.Resources) (proc *commandProc, err error) {
+func newCommandProcFromParsed(conf *service.ParsedConfig, _ *service.Resources) (proc *commandProc, err error) {
```
gregfurman:
```suggestion
func newCommandProcFromParsed(conf *service.ParsedConfig, mgr *service.Resources) (proc *commandProc, err error) {
```

internal/impl/kubernetes/auth_unix.go
```diff
@@ -0,0 +1,16 @@
+//go:build unix
```
gregfurman, line 1: praise: Neat usage of build tags here 🥳

internal/impl/kubernetes/config.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+	pairs := make([]string, 0, len(labels))
+	for k, v := range labels {
+		pairs = append(pairs, k+"="+v)
+	}
+	return strings.Join(pairs, ",")
+}
+
+// MetadataDescription returns the standard metadata documentation block.
+func MetadataDescription(fields []string) string {
```
gregfurman: nit: these functions wont ever be exported so probably better to keep them private. 

Also, could be a bit tidier to use varadic arguments:

```suggestion
func MetadataDescription(fields ...string) string {
```

internal/impl/kubernetes/input_watch.go
```diff
@@ -0,0 +1,506 @@
...
+` + "```" + `
+
+### Watch Events
+
+Each message includes the watch event type in metadata:
+- **ADDED**: Resource was created
+- **MODIFIED**: Resource was updated
+- **DELETED**: Resource was removed
+` + MetadataDescription([]string{
+			"kubernetes_watch_event_type",
```
gregfurman, line 69: nit: make these metadata fields constant so there wont ever be drift between the docs and the metadata fields being populated in `eventToMessage`

internal/impl/kubernetes/input_watch.go
```diff
@@ -0,0 +1,506 @@
...
+		listOpts := metav1.ListOptions{
+			LabelSelector: k.labelSelector,
+			FieldSelector: k.fieldSelector,
+		}
+
+		// If include_initial_list is true on first run, list existing resources
+		resourceVersion := k.getResourceVersion(namespace)
+		if k.includeInitialList && resourceVersion == "" {
+			k.listExistingResources(resourceInterface, namespace)
+		}
```
gregfurman, line 310: Should this reside inside the loop? Should we not be initialising it beforehand?

internal/impl/kubernetes/input_watch.go
```diff
@@ -0,0 +1,506 @@
...
+
+		// Use stored resource version if available
+		resourceVersion = k.getResourceVersion(namespace)
+		if resourceVersion != "" {
+			listOpts.ResourceVersion = resourceVersion
+		}
+
+		// Start watching
+		watchCtx, watchDone := k.shutSig.SoftStopCtx(context.Background())
```
gregfurman: nit: perhaps rather use `defer watchDone()` since that'll we cancel the ctx and avoid leaks

internal/impl/kubernetes/input_watch.go
```diff
@@ -0,0 +1,506 @@
...
+	creationTS := obj.GetCreationTimestamp()
+	if !creationTS.Time.IsZero() {
+		msg.MetaSetMut("kubernetes_resource_creation_timestamp",
+			creationTS.Format(time.RFC3339))
+	}
+
+	// Add labels as metadata
+	for key, value := range obj.GetLabels() {
+		msg.MetaSetMut(k.keyCache.getMetaKey("kubernetes_labels_", key), value)
+	}
```
gregfurman, line 496: What if a label changes? Do we ever ensure the cache is invalidated?

Perhaps its a better idea to just write the labels out each time and avoid the cache confusion. wdyt?

internal/impl/kubernetes/input_watch.go
```diff
@@ -0,0 +1,506 @@
...
+		namespaces = []string{""} // Empty string = all namespaces
+	}
+
+	for _, ns := range namespaces {
+		k.wg.Add(1)
+		go func(namespace string) {
+			defer k.wg.Done()
+			k.watchNamespace(namespace)
+		}(ns)
+	}
```
gregfurman, line 277: We can use the fancy new waitgroup method 👇 

```suggestion
	for _, ns := range namespaces {
		k.wg.Go(func() {
			k.watchNamespace(ns)
		})
	}
```

internal/impl/kubernetes/input_watch.go
```diff
@@ -0,0 +1,506 @@
...
+		}
+	}
+
+	return k, nil
+}
+
+func (k *kubernetesWatchInput) getResourceVersion(namespace string) string {
+	k.mu.Lock()
+	defer k.mu.Unlock()
+	return k.resourceVers[namespace]
```
gregfurman, line 253: nit: consider making `mu` a `RWMutex` since you're doing a write lock on this map when _technically_ this is a (cheaper) read operation

internal/impl/kubernetes/input_watch.go
```diff
@@ -0,0 +1,506 @@
...
+		return nil // Watch events don't require acknowledgment
+	}, nil
+}
+
+func (k *kubernetesWatchInput) Close(ctx context.Context) error {
+	k.shutSig.TriggerSoftStop()
+	k.wg.Wait()
+	close(k.eventChan)
+	return nil
+}
```
gregfurman, line 523: I know I'm being pedantic here but if we're waiting too long on a component's `Close` that can be problematic.

This is a quick-and-dirty solution but one could leverage the `TriggerHasStopped` to signal the `eventChan` and all watchers have been closed. Note, that the `eventChan` could still contain data so doing a `case _, ok := <-k.eventChan:` would require properly draining the channel before reaching the close signal.

```suggestion
func (k *kubernetesWatchInput) Close(ctx context.Context) error {
	go func() {
		k.shutSig.TriggerSoftStop()
		k.wg.Wait()
		close(k.eventChan)
		k.shutSig.TriggerHasStopped()
	}()

	select {
	case <-ctx.Done():
		return ctx.Err()
	case <-k.shutSig.HasStoppedChan():
		return nil
	}
}
```

internal/impl/kubernetes/auth_unix.go
aronchick: TY! Stole this from my time on Kubernetes :)

internal/impl/io/input_file.go
aronchick: apologies! accidentally over aggressive linting on my side. Reverted.

internal/impl/kubernetes/config.go
aronchick: done

internal/impl/kubernetes/input_watch.go
aronchick: only four hard problems in computing - naming, cache invalidation, off-by-one errors, only once delivery, and naming.

(Removing caching)

internal/impl/kubernetes/input_watch.go
aronchick: done

internal/impl/kubernetes/input_watch.go
aronchick: done.

internal/impl/kubernetes/input_watch.go
aronchick: done

internal/impl/kubernetes/input_watch.go
aronchick: ooo, don't think i saw that before lol. changed.

internal/impl/kubernetes/input_watch.go
aronchick: makes sense to me!

aronchick:
@gregfurman All feedback addressed and tests passing locally - ready for re-review!

aronchick:
@gregfurman tag you're it. (i thinki got everything)

gregfurman:
@aronchick Will give this a final look today! Thanks for addressing all my feedback 🙌

gregfurman (commented):
Looks good! Last, you just need to add the `kubernetes` implementation to `public/components` in order to register this plugin to the global bento environment. 

Here's an example of this for the `mqtt` component:

https://github.com/warpstreamlabs/bento/blob/85332f02a530ed436721bcc92169e4940d7bc34e/public/components/mqtt/package.go#L1-L6

then we can add this to `public/components/all` so that importing in `github.com/warpstreamlabs/bento/public/components/all` will automatically register the component to the global environment i.e

https://github.com/warpstreamlabs/bento/blob/85332f02a530ed436721bcc92169e4940d7bc34e/public/components/all/package.go#L37

Lastly, run `make docs` and commit the generated `.md` files for the kubernetes component that will be used in the bento doc site.

Once done, think it's good to go!

gregfurman:
@jem-davies The K8s API adds an extra 24.4 MB to the bento binary size (at least on Darwin) which ends up constituing like 11% of the size. Should we add this package behind the `x_bento_extra` build tag? Or perhaps a custom `kubernetes` build tag could work as well.

See `gsa` analysis output below of bento binary 👇 

```
┌────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
│ bento                                                                                                              │
├─────────┬─────────────────────────────────────────────────────────────────────────────────────┬────────┬───────────┤
│ PERCENT │ NAME                                                                                │ SIZE   │ TYPE      │
├─────────┼─────────────────────────────────────────────────────────────────────────────────────┼────────┼───────────┤
│ 12.10%  │ __rodata __TEXT                                                                     │ 27 MB  │ section   │
│ 12.03%  │ __rodata __DATA_CONST                                                               │ 27 MB  │ section   │
│ 6.91%   │ k8s.io/api                                                                          │ 15 MB  │ vendor    │
│ 5.85%   │ github.com/aws/aws-sdk-go-v2                                                        │ 13 MB  │ vendor    │
│ 5.30%   │ github.com/elastic/go-elasticsearch/v9                                              │ 12 MB  │ vendor    │
│ 5.14%   │ github.com/apache/arrow-go/v18                                                      │ 11 MB  │ vendor    │
│ 4.25%   │ k8s.io/client-go                                                                    │ 9.4 MB │ vendor    │
├─────────┼─────────────────────────────────────────────────────────────────────────────────────┼────────┼───────────┤
│ 100.00% │ Known                                                                               │ 222 MB │           │
│ 100%    │ Total                                                                               │ 222 MB │           │
└─────────┴─────────────────────────────────────────────────────────────────────────────────────┴────────┴───────────┘
```

jem-davies:
> @jem-davies The K8s API adds an extra 24.4 MB to the bento binary size (at least on Darwin) which ends up constituing like 11% of the size. Should we add this package behind the `x_bento_extra` build tag? Or perhaps a custom `kubernetes` build tag could work as well.

My take is that if a user is worried about binary size or say, CVE's in components they don't use - then we should advise them to create their own distribution where they import select components. 

So therefore I don't think we should use a build tag requiring users to specifically 'opt-in' to a k8s component.

PR #631, opened 2025-12-29
fix(sql): add IAM token refresh support for sql_select processor

marxus:
When using IAM authentication for RDS, the token expires after ~15 minutes. The sql_select processor now detects connection failures (driver.ErrBadConn) and automatically reconnects with a fresh IAM token, then retries the query.

Changes:
- Store driver, dsn, and connSettings on struct for reconnection
- Add reconnect() method that closes old connection and establishes new one
- Detect driver.ErrBadConn in ProcessBatch and trigger reconnection
- Retry failed query after successful reconnection

Related to: #469

### DIFF
--- a/internal/impl/sql/processor_sql_select.go
+++ b/internal/impl/sql/processor_sql_select.go
@@ -3,6 +3,8 @@ package sql
 import (
 	"context"
 	"database/sql"
+	"database/sql/driver"
+	"errors"
 	"fmt"
 	"sync"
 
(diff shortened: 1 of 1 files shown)


marxus:
THIS IS NOT TO BE MERGED YET. the fix was "vibed" and havent been tested yet. it seems like the above fix on #469 only apply to partial sql methods. not sure if this is the right way to add the reconnect mechanism. so treat this contribute as a proposal for a fix.

image

PR #633, opened 2025-12-29
doc fix new root span with link kafka version

jem-davies:
erroneous version number for a new field `new_root_span_with_link` - this PR updates it to v1.14.0

### DIFF
--- a/public/service/config_extract_tracing.go
+++ b/public/service/config_extract_tracing.go
@@ -32,7 +32,6 @@ func NewExtractTracingSpanMappingField() *ConfigField {
 func NewRootSpanWithLinkField() *ConfigField {
 	return NewBoolField(nrswlField).
 		Description("EXPERIMENTAL: Starts a new root span with link to parent.").
-		Version("1.0.0").
 		Optional().
 		Advanced()
 }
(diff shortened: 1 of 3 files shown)


public/service/config_extract_tracing.go
```diff
@@ -32,7 +32,6 @@ func NewExtractTracingSpanMappingField() *ConfigField {
 func NewRootSpanWithLinkField() *ConfigField {
 	return NewBoolField(nrswlField).
 		Description("EXPERIMENTAL: Starts a new root span with link to parent.").
-		Version("1.0.0").
```
gregfurman, line 35: suggestion: Should we not rather replace this with `Version("1.14.0")`?

Also, you can see how this was used in NATS:

https://github.com/warpstreamlabs/bento/blob/ac10c9066920efbcbde8167ff1ab2752cb630ef7/internal/impl/nats/docs.go#L23-L28

jem-davies: Yeah I saw that and realised that it must be overwritten - not sure it will make much practical difference - or will be confusing to have to calls to `.Version()` imo .

PR #635, opened 2025-12-29
feat(core/messages): update message proto spec for multi-batch support

gregfurman:
## Motivation

The current message protobuf defintions do not have support for multiple batches, useful for `ProcessBatch` implementations since multiple batches are returned.

## Changes

- Adds a `Batches` type to the message protobuf definition in `resources/protos/message.proto`
- Adds helper methods for translating to and from a batch of batches that are in protobuf format i.e `UnmarshalBatchesFromProto` and `MarshalBatchesToProto`

### DIFF
--- a/internal/message/protobuf.go
+++ b/internal/message/protobuf.go
@@ -59,13 +59,13 @@ func UnmarshalBatchesFromProto(b []byte) ([]Batch, error) {
 	}
 
 	batches := make([]Batch, len(protoBatch.GetBatches()))
-	for _, protoBatch := range protoBatch.GetBatches() {
+	for i, protoBatch := range protoBatch.GetBatches() {
 		batch, err := toBatch(protoBatch)
 		if err != nil {
 			return nil, err
 		}
 
-		batches = append(batches, batch)
+		batches[i] = batch
 	}
 
 	return batches, nil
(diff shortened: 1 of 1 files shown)


internal/message/protobuf.go
```diff
@@ -52,6 +52,45 @@ func UnmarshalBatchFromProto(b []byte) (Batch, error) {
...
+	}
+
+	batches := make([]Batch, len(protoBatch.GetBatches()))
+	for _, protoBatch := range protoBatch.GetBatches() {
+		batch, err := toBatch(protoBatch)
+		if err != nil {
+			return nil, err
+		}
+
+		batches = append(batches, batch)
```
jem-davies: Is that correct to make a slice with a predetermined size of len(protoBatch.GetBatches()) but then append to it?

internal/message/protobuf.go
```diff
@@ -52,6 +52,45 @@ func UnmarshalBatchFromProto(b []byte) (Batch, error) {
...
 }
 
+func UnmarshalBatchesFromProto(b []byte) ([]Batch, error) {
+	protoBatch := &pb.Batches{}
+	if err := proto.Unmarshal(b, protoBatch); err != nil {
+		return nil, err
+	}
+
+	batches := make([]Batch, len(protoBatch.GetBatches()))
+	for _, protoBatch := range protoBatch.GetBatches() {
```
jem-davies: The protoBatch var inside the loop is the same as the one outside the loop - could it be `pb` or something instead?

jem-davies (changes requested):
Couple of comments

jem-davies (approved):

PR #645, opened 2026-01-08
add input file tail

jem-davies:
Some context: I had decided to implement this as a simpler alternative to #514.

### DIFF
--- a/website/docs/components/inputs/file_tail.md
+++ b/website/docs/components/inputs/file_tail.md
@@ -20,6 +20,8 @@ This component is experimental and therefore subject to change or removal outsid
 :::
 Tails a file
 
+Introduced in version 1.15.0.
+
 
 <Tabs defaultValue="common" values={[
   { label: 'Common', value: 'common', },


internal/impl/io/input_file_tail.go
```diff
@@ -0,0 +1,370 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	var tailOpts []tailOpt
+
+	tailOpts = append(tailOpts, withPollInterval(pollInterval))
+	tailOpts = append(tailOpts, withLineChanBufferSize(lineBufferSize))
+	tailOpts = append(tailOpts, withStartPosition(startPosition))
+	tailOpts = append(tailOpts, withLogger(mgr.Logger()))
```
jem-davies:
```suggestion
	tailOpts := []tailOpt{
		withPollInterval(pollInterval),
		withLineChanBufferSize(lineBufferSize),
		withStartPosition(startPosition),
		withLogger(mgr.Logger()),
	}
```

internal/impl/io/input_file_tail.go
```diff
@@ -0,0 +1,370 @@
...
+		}, nil
+
+	case <-ctx.Done():
+		return nil, nil, ctx.Err()
+	}
+}
+
+func (fti *FileTailInput) Close(ctx context.Context) error {
+	fti.tail.cancel()
+	fti.tail.file.Close()
```
jem-davies: I think we need to wait for the fti.tail.cancel() to stop the watch() func - then call fti.tail.file.Close() ... 

But because I am using the closing of the channels already to signal "service.ErrNotConnected" then I think we will need to use the shutdown signaller

internal/impl/io/input_file_tail.go
```diff
@@ -0,0 +1,395 @@
...
+	if err != nil {
+		return fmt.Errorf("seek file: %w", err)
+	}
+
+	var truncation bool
+	var moved bool
+
+	if !os.SameFile(t.fileInfo, tempInfo) {
+		if t.logger != nil {
+			t.logger.Info(fmt.Sprintf("Handling rotation for %v", t.path))
```
gregfurman: Can we downgrade these rotation logs to the `debug` level?

internal/impl/io/input_file_tail.go
```diff
@@ -0,0 +1,395 @@
...
+
+func (t *tail) reopenIfMovedOrTruncated() error {
+	tempInfo, err := os.Stat(t.path)
+	if err != nil {
+		return fmt.Errorf("stat file: %w", err)
+	}
+
+	pos, err := t.file.Seek(0, io.SeekCurrent)
+	if err != nil {
+		return fmt.Errorf("seek file: %w", err)
```
gregfurman: Perhaps some of these errors should have more info added:

```suggestion
		return fmt.Errorf("failed to set seek on %s: %w", t.file.Name(). , err)
```

internal/impl/io/input_file_tail.go
```diff
@@ -0,0 +1,395 @@
...
+
+		select {
+		case <-ctx.Done():
+			t.doneChan <- struct{}{}
+			return
+		default:
+		}
+
+		line, err := t.reader.ReadBytes('\n')
+		line = bytes.TrimRight(line, "\r\n")
```
gregfurman, line 307: Any particular reason why we're exclusively using `\n` as a delimiter here?

internal/impl/io/input_file_tail.go
```diff
@@ -0,0 +1,395 @@
...
+}
+
+//------------------------------------------------------------------------------
+
+type tail struct {
+	path         string
+	pollInterval time.Duration
+
+	file     *os.File
+	fileInfo os.FileInfo
```
gregfurman, line 198: Thoughts on using the `*service.FS` from the `mgr *service.Resources` instead for file reading?


i.e

```go
mgr.FS().OpenFile()
```

This way we're relying on our `FS` fixtures instead of the `os` package

internal/impl/io/input_file_tail.go
```diff
@@ -0,0 +1,395 @@
...
+
+	case <-ctx.Done():
+		return nil, nil, ctx.Err()
+
+	case <-fti.shutSig.HardStopChan():
+		return nil, nil, nil
+	}
+}
+
+func (fti *FileTailInput) Close(ctx context.Context) error {
```
gregfurman, line 171: Is this idempotent? Also, we should probably select on the `ctx.Done()` to allow returning early if the `ctx` is cancelled

internal/impl/io/input_file_tail.go
jem-davies: Yes could have a config option for that

internal/impl/io/input_file_tail.go
jem-davies: done this 👍

internal/impl/io/input_file_tail.go
jem-davies: Added a check to see if the HardStop has already been called & added a select with a case for the ctx too 👍

internal/impl/io/input_file_tail.go
jem-davies: I have updated the component description to say "Reads lines from a file continuously with handling of log rotation." -- we could add it as a config option I suppose - but then would be interested too see why / what else people would want to use as a delimiter. 

If there is an issue raised it should be ok to add a new config option with a default of "\n" for backwards comp. 

So thinking to leave this as-is

jem-davies:
@gregfurman  - thanks for review  🙏  - I have either altered the code / responded to each comment 😄

internal/impl/io/input_file_tail.go
```diff
@@ -0,0 +1,414 @@
...
+	ftiFieldPath              = "path"
+	ftiFieldPollInterval      = "poll_interval"
+	ftiFieldMaxLineBufferSize = "line_buffer"
+	ftiFieldStartPosition     = "start_position"
+)
+
+func fileTailInputSpec() *service.ConfigSpec {
+	return service.NewConfigSpec().
+		Categories("Local").
+		Summary("Tails a file").
```
jem-davies, line 28:
```suggestion
		Summary("Tails a file").
		Version("1.15.0").
```

PR #655, opened 2026-01-16
feat: add JSON schema generator tool

### LINKED ISSUE #647: Provide JSON Schema to enable auto completion in IDEs
Is there a possibility to get a [JSON Schema](https://json-schema.org/) of all available bento components like it is provided for CUE?
This would enable auto completion and validation for bento yaml config files for a broad range of IDEs via the [yaml-language-server](https://github.com/redhat-developer/yaml-language-server).

Here is a mock-up with a minimal JSON Schema:

https://github.com/user-attachments/assets/ab6729ce-afc9-43a3-b31e-36d60731b97d

mueckinger:
Add a new command-line tool `json_schema_gen` that generates JSON schemas from Bento configuration specifications. This tool converts `FieldSpec` definitions into proper JSON Schema format, handling various field types, nested objects, arrays, and special component configurations like processors, caches, and rate limits.

The implementation includes:
- Core JSON Schema structs (`JSONSchema`, `Property`)
- Conversion logic from Bento field types to JSON Schema types
- Special handling for component arrays (processors, caches, etc.)
- Default value inclusion in property descriptions
- Example values preservation
- Proper type mapping for all supported Bento field types

This enables auto completion and validation capabilities for Bento configurations in almost every IDE.

This resolves #647.

I decided not to include it in the main `bento` CLI, but create a separate tool as schema creation is usually not required locally. But if you like, we can include this in the CLI later. It's more important to serve the schema via the bento repo.

If you accept this PR please run `go run cmd/tools/json_schema_gen/main.go > bento.json` and put the resulting JSON Schema in a place of your choice in the bento repo. It is VERY IMPORTANT that this location and file name MUST NOT change!
Once the schema file is in place I will register the Schema for `bento.yaml` as it is used in the [Dockerfile](https://github.com/warpstreamlabs/bento/blob/main/resources/docker/Dockerfile) at [schemastore.org](https://www.schemastore.org/), so the schema will automatically be imported by the IDEs, once a user creates a `bento.yaml` file.


aronchick:
oooo very very cool

cmd/tools/json_schema_gen/main.go
```diff
@@ -0,0 +1,275 @@
...
+	jsonSchema.Properties = configProps
+
+	// Output the JSON Schema
+	output, err := json.MarshalIndent(jsonSchema, "", "  ")
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Error marshaling JSON: %v\n", err)
+		os.Exit(1)
+	}
+
+	fmt.Println(string(output))
```
jem-davies, line 265: Wondering if it would be better to write to a file?

jem-davies:
I have added the output of the binary to ./resources/schemastore and have opened the PR: https://github.com/SchemaStore/schemastore/pull/5293 

🙏 Thanks for this contribution ! 

I will open up an issue to see about adding this to the makefile & Release process. 

Issue #658

cmd/tools/json_schema_gen/main.go
jem-davies: We can take a look at this as part of: #658

PR #664, opened 2026-01-20
Add aws_s3_stream output

triddell:
## Overview
This pull request introduces `aws_s3_stream`, a new output plugin designed to stream generic text-based files (JSON, NDJSON, CSV, logs) directly to Amazon S3 using multipart uploads. The implementation addresses two critical issues: excessive memory consumption and incorrect partition routing in the existing batch approach.

**Primary benefit**: Massive memory reduction (95%+) enables processing large datasets that fail with batch approach, while maintaining constant memory usage.

## Key Problems Addressed

**Memory Efficiency**: The current `aws_s3` output with batching buffers entire batches in memory before writing, consuming approximately 3.31 GB for 100K events. This solution reduces memory usage to ~180 MB—a 95% improvement. For 500K events, the batch approach fails completely after 15+ minutes, while streaming succeeds in 38.56 seconds with only 170 MB of memory.

**Partition Routing**: When using archive processors in batching, metadata from only the first message is preserved, which can cause partition misrouting. This output's partition_by design prevents this pattern by maintaining per-message partition evaluation.

**Small File Problem**: Batch processing creates many small files instead of optimal large files per partition. For example, 500K events produces 13+ incomplete files with batch vs 1 optimal 788 MB file with streaming, negatively impacting query performance and S3 costs.

## Solution Architecture

The implementation uses:
- Direct streaming to S3 multipart uploads without batch buffering
- Per-message partition routing via `partition_by` expressions
- Per-partition writer pooling with automatic lifecycle management
- Configurable flush triggers: size (5MB), count (10K), or time (10s)

Memory footprint: 170–260 MB (constant regardless of dataset size).

## Files Added

**Core implementation** (5 files):
- `output_aws_s3_stream.go` (438 lines): Plugin configuration and partition routing logic
- `s3_streaming_writer.go` (397 lines): Generic S3 streaming writer with multipart upload management
- `output_aws_s3_stream_test.go` (373 lines): Unit tests for output configuration and partition logic
- `output_aws_s3_stream_integration_test.go` (371 lines): LocalStack integration tests
- `s3_streaming_writer_test.go` (405 lines): Unit tests for streaming writer with mock S3 client

## Files Modified

None. This PR only adds new files without modifying any existing code.

## Configuration Example

```yaml
output:
  aws_s3_stream:
    bucket: my-logs-bucket
    path: 'logs/date=${! meta("date") }/service=${! meta("service") }/${! uuid_v4() }.json'
    partition_by:
      - '${! meta("date") }'
      - '${! meta("service") }'
    max_buffer_bytes: 5242880  # 5MB (S3 multipart optimal)
    max_buffer_count: 10000
    max_buffer_period: 10s
    content_type: 'application/x-ndjson'
```

### With Compression

```yaml
pipeline:
  processors:
    - bloblang: |
        root = content().string() + "\n"
    - compress:
        algorithm: gzip

output:
  aws_s3_stream:
    bucket: my-logs-bucket
    path: 'logs/date=${! meta("date") }/${! uuid_v4() }.json.gz'
    partition_by:
      - '${! meta("date") }'
    content_encoding: 'gzip'
```

## Testing

The PR includes 30 unit tests and 4 LocalStack-based integration tests validating:

**Unit tests (18 tests for output plugin)**:
- Configuration parsing (basic, partition_by, buffer settings, content type/encoding)
- Error cases (missing bucket, missing path, invalid period)
- Path interpolation with dynamic metadata
- Partition grouping logic
- Backwards compatibility (behavior without partition_by)

**Unit tests (12 tests for streaming writer)**:
- Writer creation with defaults and custom config
- Initialization lifecycle (single init, double-init error, uninitialized write error)
- Buffer flushing on size threshold (>5MB), count threshold, and stats tracking
- Close behavior (completes upload, write-after-close errors)
- Multipart uploads (multiple parts, proper part numbering)
- Content encoding metadata propagation

**Integration tests (4 tests with LocalStack)**:
- Basic write operations with 150 messages
- Partition routing with 4 partitions (2 dates × 2 regions), verifying separate files
- Content type and encoding metadata validation
- Edge case: single message handling

**Manual testing** with OCSF security event data (complex nested JSON):
- 50K, 100K, 500K event validation
- Partition correctness: 100% accuracy with interleaved test data (alternating accounts)
- Memory profiling: Constant usage across all dataset sizes
- Compression trade-offs: GZIP adds 31% processing time, saves 54% storage

## Performance Results

| Dataset | Batch Memory | Batch Time | Stream Memory | Stream Time | Memory Reduction |
|---------|--------------|------------|---------------|-------------|------------------|
| 50K     | 1.73 GB      | 2.88s      | 0.17 GB       | 4.80s       | **90%**          |
| 100K    | 3.31 GB      | 6.13s      | 0.18 GB       | 11.19s      | **95%**          |
| 500K    | **FAILED**   | >900s      | 0.17 GB       | 38.56s      | **93%+**         |

At small scales, batch is faster. At production scales, **batch fails completely** while streaming succeeds with constant memory.

## Why No Built-in Compression?

Unlike `aws_s3`, this output does **not** provide a `compress` parameter. S3 multipart uploads require non-final parts to be ≥5MB. After compressing a 5MB buffer, the compressed size is unpredictable (often <5MB), causing S3 to reject the upload.

**Solution**: Use pipeline-level compression via the `compress` processor (see example above). This works reliably because concatenated gzip streams are valid, and each compressed message can be decompressed independently.

### Before (Broken)
```yaml
output:
  aws_s3:
    path: 'partition=${! meta("key") }/${! uuid_v4() }.json'
    batching:
      count: 10000
      processors:
        - archive:
            format: lines
```

### After (Fixed)
```yaml
pipeline:
  processors:
    - bloblang: |
        root = content().string() + "\n"

output:
  aws_s3_stream:
    path: 'partition=${! meta("key") }/${! uuid_v4() }.json'
    partition_by: ['${! meta("key") }']
    max_buffer_bytes: 5242880
```

**Key changes:**
1. Move `archive: lines` logic to pipeline (add `\n` to each message)
2. Add `partition_by` parameter for correct routing
3. Remove `batching` section from output
4. Use `aws_s3_stream` instead of `aws_s3`

## Backwards Compatibility

No breaking changes. This PR introduces a new plugin (`aws_s3_stream`) alongside the existing `aws_s3` output. Users can migrate at their own pace. All code is new with zero modifications to existing files.

## Related Work

This PR follows the same pattern as PR #661 (`aws_s3_parquet_stream`), which addresses identical partition routing and memory issues for Parquet format. Both implementations use:
- Per-message partition evaluation
- Per-partition writer pooling
- Streaming multipart uploads
- Constant memory usage

## Documentation

The plugin includes comprehensive inline documentation:
- Detailed config spec with examples
- Description of partition routing behavior
- Buffer tuning guidelines
- Compression recommendations
- Migration guide from `aws_s3`


triddell:
## CI Fixes

Fixed all linting errors and CI check failures:
- Fixed documentation generation by removing YAML example in description that was causing validation errors
- Fixed Bloblang syntax in examples (`timestamp().format()` → `now().ts_format()`)
- Replaced `fmt.Errorf` with `errors.New` for static error strings
- Updated deprecated AWS SDK endpoint resolver to use `BaseEndpoint`
- Fixed testify verbose assertions and removed unnecessary type conversions
- Fixed errcheck error in test
- Applied gofmt formatting

All tests and lint checks now pass. I will review the generated documentation output more closely for any additional improvements needed.

jem-davies:
### Regarding the partition routing problem

In the PR description you have written: 

> The batch approach evaluates path expressions once per batch rather than per message.

But does it? 

In the existing `aws_s3` output I can see: 

[here](https://github.com/warpstreamlabs/bento/blob/fb2b6cab29a954d36d1b97e59a3240a755a3d098/internal/impl/aws/output_s3.go#L390)

```go
		key, err := msg.TryInterpolatedString(i, a.conf.Path)
		if err != nil {
			return fmt.Errorf("key interpolation: %w", err)
		}
```

For each message in the batch we do evaluate the key? 

In your config you shared: 

```yaml
output:
  aws_s3:
    path: 'partition=${! meta("key") }/${! uuid_v4() }.json'
    batching:
      count: 10000
      processors:
        - archive:
            format: lines
```

You are using the `archive` processor to reduce the 10k batch into 1 message - and you should be seeing something like:  

```
ERRO Failed to send message to aws_s3: key interpolation: message contains multiple valid documents  @service=bento label="" path=root.output
```

`edit: perhaps not ...`

assuming the individual messages are json. 

--- 

Sounds like you could use the [group_by_value](https://warpstreamlabs.github.io/bento/docs/components/processors/group_by_value/) processor - like: 

```yaml
input:
  batched:
    child:
      file:
        paths:
          - ./test_data/*.json
    policy:
      count: 10

pipeline:
  processors:
    - group_by_value:
        value: ${! json("key") }

    - mapping: |
        if batch_index() == 0 { meta key = this.key }

    - archive:
        format: lines

output:
  aws_s3:
    bucket: bento-aws-s3-stream-test
    path: ${! meta("key") }/foobar.json
```

./test_data/ is populated with some json files that have a json key "key" with either "A" / "B" as the value - and then in the s3 bucket you get two different s3 keys: /A/foobar.json & /B/foobar.json - split such that the s3 keys "A" & "B" only have the json files with the matching "key" value. 

---

Could you elaborate on what the issue is exactly with the `aws_s3` output not correctly routing? - because I am currently thinking that it would be caused by using the `archive` processor in the `output.batching.processors` section.

jem-davies:
Yeah it is mentioned on the [`archive` processor docs](https://warpstreamlabs.github.io/bento/docs/components/processors/archive) : 

> The resulting archived message adopts the metadata of the first message part of the batch.

jem-davies:
Guess the [aws_s3 batching docs ](https://warpstreamlabs.github.io/bento/docs/components/outputs/aws_s3#batching) fail to mention this...

internal/impl/aws/output_aws_s3_stream.go
```diff
@@ -0,0 +1,422 @@
...
+	return &s3StreamOutput{
+		conf:     conf,
+		log:      mgr.Logger(),
+		s3Client: s3Client,
+		writers:  make(map[string]*S3StreamingWriter),
+	}, nil
+}
+
+func (s *s3StreamOutput) Connect(ctx context.Context) error {
+	s.log.Infof("Streaming S3 output configured for bucket: %s", s.conf.Bucket)
```
jem-davies:
```suggestion
```

internal/impl/aws/output_aws_s3_stream.go
```diff
@@ -0,0 +1,422 @@
...
+### Credentials
+
+By default Bento will use a shared credentials file when connecting to AWS services.
+You can find out more [in this document](/docs/guides/cloud/aws).
+`).
+		Fields(
+			service.NewStringField(ssoFieldBucket).
+				Description("The S3 bucket to upload files to."),
+			service.NewInterpolatedStringField(ssoFieldPath).
+				Description("The path for each file. Supports Bloblang interpolation for dynamic partitioning.").
```
jem-davies:
```suggestion
				Description("The path for each file.").
```

internal/impl/aws/output_aws_s3_stream.go
```diff
@@ -0,0 +1,422 @@
...
+	ssoFieldMaxBufferPeriod    = "max_buffer_period"
+	ssoFieldBatching           = "batching"
+	ssoFieldContentType        = "content_type"
+	ssoFieldContentEncoding    = "content_encoding"
+)
+
+func s3StreamOutputSpec() *service.ConfigSpec {
+	return service.NewConfigSpec().
+		Beta().
+		Version("1.0.0").
```
jem-davies:
```suggestion
		Version("1.16.0").
```

internal/impl/aws/output_aws_s3_stream.go
```diff
@@ -0,0 +1,422 @@
...
+				Advanced(),
+			service.NewIntField(ssoFieldMaxBufferBytes).
+				Description("Maximum buffer size in bytes before flushing to S3. Default is 10MB.").
+				Default(10*1024*1024).
+				Advanced(),
+			service.NewIntField(ssoFieldMaxBufferCount).
+				Description("Maximum number of messages to buffer before flushing to S3. Default is 10000.").
+				Default(10000).
+				Advanced(),
+			service.NewStringField(ssoFieldMaxBufferPeriod).
```
jem-davies:
```suggestion
			service.NewDurationField(ssoFieldMaxBufferPeriod).
```

internal/impl/aws/output_aws_s3_stream.go
```diff
@@ -0,0 +1,422 @@
...
+		return
+	}
+
+	var periodStr string
+	if periodStr, err = pConf.FieldString(ssoFieldMaxBufferPeriod); err != nil {
+		return
+	}
+	if conf.MaxBufferPeriod, err = time.ParseDuration(periodStr); err != nil {
+		return conf, fmt.Errorf("failed to parse max_buffer_period: %w", err)
+	}
```
jem-davies:
```suggestion
	if conf.MaxBufferPeriod, err = pConf.FieldDuration(ssoFieldMaxBufferPeriod); err != nil {
		return
	}
```

internal/impl/aws/output_aws_s3_stream.go
```diff
@@ -0,0 +1,422 @@
...
+	log      *service.Logger
+	s3Client *s3.Client
+
+	// Writer pool for managing multiple partition paths
+	writersMut sync.Mutex
+	writers    map[string]*S3StreamingWriter
+}
+
+func newS3StreamOutput(conf s3StreamConfig, mgr *service.Resources) (*s3StreamOutput, error) {
+	s3Client := s3.NewFromConfig(conf.aconf, func(o *s3.Options) {
```
jem-davies: Consider moving the creation of the s3Client to the Connect func.

internal/impl/aws/output_aws_s3_stream.go
```diff
@@ -0,0 +1,422 @@
...
+			return fmt.Errorf("failed to write to partition %s: %w", pg.key, err)
+		}
+	}
+
+	return nil
+}
+
+func (s *s3StreamOutput) writeToPartition(ctx context.Context, partitionKey string, path string, batch service.MessageBatch) error {
+	// Get or create writer (with lock)
+	s.writersMut.Lock()
```
jem-davies: Perhaps it would make sense to use a RWMutex here - hold the read lock when we need that - and if/when we need to update the map for a new partition we can hold a write lock? 

Would have to check again that it doesn't exist after acquiring the write lock.

internal/impl/aws/output_aws_s3_stream.go
```diff
@@ -0,0 +1,422 @@
...
+	writer, exists := s.writers[partitionKey]
+
+	if !exists {
+		// Evaluate content type and encoding for this partition
+		var contentType string
+		var contentEncoding string
+		var err error
+
+		if len(batch) > 0 {
+			contentType, err = batch.TryInterpolatedString(0, s.conf.ContentType)
```
jem-davies: So we are just using the first message in a batch to get the contentType & ContentEncoding - which makes sense - but isn't mentioned on the docs.

internal/impl/aws/output_aws_s3_stream.go
```diff
@@ -0,0 +1,422 @@
...
+	}
+
+	return nil
+}
+
+func (s *s3StreamOutput) Close(ctx context.Context) error {
+	s.writersMut.Lock()
+	defer s.writersMut.Unlock()
+
+	s.log.Infof("Closing %d active writers", len(s.writers))
```
jem-davies:
```suggestion
```

Or downgrade to a debug log

internal/impl/aws/output_aws_s3_stream.go
```diff
@@ -0,0 +1,422 @@
...
+	ssoFieldContentType        = "content_type"
+	ssoFieldContentEncoding    = "content_encoding"
+)
+
+func s3StreamOutputSpec() *service.ConfigSpec {
+	return service.NewConfigSpec().
+		Beta().
+		Version("1.0.0").
+		Categories("Services", "AWS").
+		Summary(`Streams data directly to S3 using multipart uploads with minimal memory overhead.`).
```
jem-davies:
```suggestion
		Summary(`Streams data to S3 using multipart uploads.`).
```

internal/impl/aws/output_aws_s3_stream.go
```diff
@@ -0,0 +1,422 @@
...
+		Description(`
+This output writes files to S3 by streaming content incrementally using S3 multipart uploads.
```
jem-davies:
```suggestion
This output writes to S3 by streaming content incrementally using S3 multipart uploads.
```

internal/impl/aws/output_aws_s3_stream.go
```diff
@@ -0,0 +1,422 @@
...
+Unlike the standard `+"`aws_s3`"+` output (which buffers the entire file in memory),
+this output streams data directly to S3, reducing memory usage by 80-90% for large files.
```
jem-davies:
```suggestion
```

Whilst I don't think that this part of the description is wrong - I think that it is focusing on the particular use-case of `aws_s3` with archiving multiple messages somewhat - and removing it will help readers to understand that the difference is using the multipart uploads with `aws_s3_stream`.

internal/impl/aws/output_aws_s3_stream.go
```diff
@@ -0,0 +1,422 @@
...
+- **S3 Multipart Upload**: Leverages S3 multipart uploads for reliable large file transfers
+- **Flexible Buffering**: Configure by bytes, message count, or time period
+
+## When to Use
+
+Use this output instead of `+"`aws_s3`"+` when:
+- Writing large files (>1GB) that would consume too much memory
+- Streaming continuous data that needs to be partitioned into files
+- You need dynamic partition routing (e.g., one file per account/date combination)
+- Memory-constrained environments (containers, Lambda, ECS)
```
jem-davies: I would probably reduce this - to focus on the difference between `aws_s3` & `aws_s3_stream` - particularly around the multipart uploads. 

Yeah it would seem that pointing out that this one would be better to use when writing large objects/files/messages. 

It seems that it repeats itself regarding partition routing / dynamic paths - which the aws_s3 should do as well - you just have to be careful with using the archive processor.

internal/impl/aws/output_s3_stream.go
```diff
@@ -0,0 +1,422 @@
...
+`).
+		Fields(
+			service.NewStringField(ssoFieldBucket).
+				Description("The S3 bucket to upload files to."),
+			service.NewInterpolatedStringField(ssoFieldPath).
+				Description("The path for each file. Supports Bloblang interpolation for dynamic partitioning.").
+				Example(`logs/${! timestamp_unix() }-${! uuid_v4() }.log`).
+				Example(`data/date=${! meta("date") }/account=${! meta("account") }/${! uuid_v4() }.json`),
+			service.NewInterpolatedStringListField(ssoFieldPartitionBy).
+				Description("Optional list of interpolated string expressions that determine writer partitioning. Messages with the same partition values are written to the same file. The full path is only evaluated once when a new partition is encountered. This allows using functions like uuid_v4() in the path for unique filenames per partition. If omitted, the full path is evaluated per message for backwards compatibility.").
```
jem-davies, line 69:
```suggestion
				Description("Optional list of interpolated string expressions that determine writer partitioning. Messages with the same partition values are written to the same file. The full path is only evaluated once when a new partition is encountered. This allows using functions like uuid_v4() in the path for unique filenames per partition. If omitted, the full path is evaluated per message.").
```

I found this to be difficult to understand - but after reading the code I did understand - perhaps would be useful to elaborate how the output works with regards to the partitions / writers, in the output description - then it might make more sense. 🤔

internal/impl/aws/output_aws_s3_stream.go
```diff
@@ -0,0 +1,422 @@
...
+			service.NewBoolField(ssoFieldForcePathStyleURLs).
+				Description("Forces path style URLs for S3 requests.").
+				Default(false).
+				Advanced(),
+			service.NewIntField(ssoFieldMaxBufferBytes).
+				Description("Maximum buffer size in bytes before flushing to S3. Default is 10MB.").
+				Default(10*1024*1024).
+				Advanced(),
+			service.NewIntField(ssoFieldMaxBufferCount).
+				Description("Maximum number of messages to buffer before flushing to S3. Default is 10000.").
```
jem-davies:
```suggestion
				Description("Maximum number of messages to buffer before flushing to S3.").
```

internal/impl/aws/output_aws_s3_stream.go
```diff
@@ -0,0 +1,422 @@
...
+				Default(10000).
+				Advanced(),
+			service.NewStringField(ssoFieldMaxBufferPeriod).
+				Description("Maximum duration to buffer messages before flushing to S3. Default is 10s.").
```
jem-davies:
```suggestion
				Description("Maximum duration to buffer messages before flushing to S3.").
```

internal/impl/aws/output_aws_s3_stream.go
```diff
@@ -0,0 +1,422 @@
...
+				Default("10s").
+				Advanced(),
+			service.NewInterpolatedStringField(ssoFieldContentType).
+				Description("The content type to set for uploaded files. Supports interpolation.").
```
jem-davies:
```suggestion
				Description("The content type to set for uploaded files.").
```

NewInterpolatedField will add something to the effect of "supports interpolation" to the description so would need to take out or will have a duplicated sentence.

internal/impl/aws/output_aws_s3_stream.go
```diff
@@ -0,0 +1,422 @@
...
+				Default("application/octet-stream").
+				Advanced(),
+			service.NewInterpolatedStringField(ssoFieldContentEncoding).
+				Description("The content encoding to set for uploaded files (e.g., gzip). Supports interpolation.").
```
jem-davies:
```suggestion
				Description("The content encoding to set for uploaded files (e.g., gzip).").
```

jem-davies (commented):
Will take another look very soon - to give a test - and look more at the s3_streaming_writer.go file.

triddell:
You're right - the partition routing issue is specific to using `archive` in `output.batching.processors`, which adopts only the first message's metadata. I see you noted that the `aws_s3` batching docs don't mention this behavior.

I'll update the PR description to focus on the memory efficiency benefits rather than framing this as a general partition routing problem with `aws_s3`.

internal/impl/aws/output_aws_s3_stream.go
jem-davies: This should come out because when the Bento engine will create a INFO log when the Connect func completes. (We gonna have 2 logs for same thing)

internal/impl/aws/s3_streaming_writer.go
```diff
@@ -0,0 +1,398 @@
...
+	shouldFlush := false
+	if w.uploadSize >= w.maxBufferBytes {
+		shouldFlush = true
+	} else if len(w.messageBuffer) >= w.maxBufferCount {
+		shouldFlush = true
+	}
+
+	if shouldFlush {
+		return w.flush(ctx)
+	}
```
jem-davies:
```suggestion
    if w.uploadSize >= w.maxBufferBytes || len(w.messageBuffer) >= w.maxBufferCount {
        return w.flush(ctx)
    }
```

internal/impl/aws/s3_streaming_writer.go
```diff
@@ -0,0 +1,398 @@
...
+		if time.Since(w.lastFlush) < w.maxBufferPeriod {
+			return nil
+		}
+	}
+
+	w.partNumber++
+
+	// Upload part with retry
+	var uploadErr error
+	for attempt := 0; attempt < 3; attempt++ {
```
jem-davies:
```suggestion
	for attempt := range 3 {
```

internal/impl/aws/s3_streaming_writer.go
```diff
@@ -0,0 +1,398 @@
...
+			}
+			w.flushTimer = time.AfterFunc(w.maxBufferPeriod, func() {
+				w.flushIfNeeded(context.Background())
+			})
+
+			return nil
+		}
+
+		uploadErr = err
+		time.Sleep(time.Second * time.Duration(1<<attempt)) // Exponential backoff
```
jem-davies: You are going to be unnecessarily sleeping on the last iteration. 

We have [BackOffField](https://pkg.go.dev/github.com/warpstreamlabs/bento/public/service#NewBackOffField) that could be used here, allowing users to configure this - maybe that would be good?

jem-davies: @triddell - added a PR for this: https://github.com/triddell/bento/pull/2/

internal/impl/aws/output_s3_stream_writer.go
```diff
@@ -0,0 +1,391 @@
...
+	resp, err := w.s3Client.CreateMultipartUpload(ctx, input)
+	if err != nil {
+		return fmt.Errorf("failed to create multipart upload: %w", err)
+	}
+
+	w.uploadID = resp.UploadId
+
+	// Start flush timer
+	w.flushTimer = time.AfterFunc(w.maxBufferPeriod, func() {
+		w.flushIfNeeded(context.Background())
```
jem-davies, line 157: There is a lot of calls to `context.Background()` in this file, when a ctx is available - how come we are making a new context, in many places?

jem-davies (commented):
I have added a PR to alter the backoff logic & make it configurable https://github.com/triddell/bento/pull/2/ 

I have given it a test with a 200MB file and is looking good 🥳

triddell:
Fixed! Changed error path `abortMultipartUpload` calls (lines 258, 308) to use the passed `ctx`. 

Kept `context.Background()` in timer callbacks (lines 155, 239, 283) since those goroutines outlive request contexts - added comments explaining why.

PR #669, opened 2026-01-23
Add support for AWS Kinesis Enhanced Fan-Out as an input

jbeemster:
This pull request introduces a new consumption mode for Kinesis with Enhanced FanOut support. What this lets us do is provide a dedicated consumption pipe for Bento instead of depending on shared polling.

As much as possible I have tried to re-use the existing Kinesis input architectures and have only added EFO in terms of consuming the events but checkpointing + batching have been left identical. Quite a lot of the logical patterns have also been copied from the polling implementation and then altered to support EFO.

### DIFF
--- a/internal/impl/aws/input_kinesis_efo.go
+++ b/internal/impl/aws/input_kinesis_efo.go
@@ -186,6 +186,7 @@ func (k *kinesisReader) runEFOConsumer(wg *sync.WaitGroup, info streamInfo, shar
 	// Channels for timed batches and message flush
 	var nextTimedBatchChan <-chan time.Time
 	var nextFlushChan chan<- asyncMessage
+	var nextRecordsChan <-chan []types.Record
 	commitCtx, commitCtxClose := context.WithTimeout(k.ctx, k.commitPeriod)
 
 	go func() {
@@ -292,8 +293,14 @@ func (k *kinesisReader) runEFOConsumer(wg *sync.WaitGroup, info streamInfo, shar
 
 			if pendingMsg.msg != nil {
 				nextFlushChan = k.msgChan
+				nextRecordsChan = nil
 			} else {
 				nextFlushChan = nil
+				if len(pending) == 0 {
+					nextRecordsChan = recordsChan
+				} else {
+					nextRecordsChan = nil
+				}
 			}
 
 			if nextTimedBatchChan == nil {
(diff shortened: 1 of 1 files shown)


internal/impl/aws/input_kinesis.go
```diff
@@ -319,6 +378,18 @@ func newKinesisReaderFromConfig(conf kiConfig, batcher service.BatchPolicy, sess
...
+
+	// Check if Enhanced Fan Out is enabled
+	if k.conf.EnhancedFanOut != nil && k.conf.EnhancedFanOut.Enabled {
+		k.efoEnabled = true
+		k.log.Infof("Enhanced Fan Out enabled")
+
+		// Validate EFO configuration
+		if k.conf.EnhancedFanOut.ConsumerName != "" && k.conf.EnhancedFanOut.ConsumerARN != "" {
+			return nil, errors.New("cannot specify both consumer_name and consumer_arn in enhanced_fan_out config")
+		}
```
jem-davies, line 390: Could you add a [LintRule](https://pkg.go.dev/github.com/warpstreamlabs/bento/public/service#ConfigSpec.LintRule) for this? It will catch misconfiguration earlier if so.

internal/impl/aws/input_kinesis_efo.go
```diff
@@ -0,0 +1,473 @@
...
+}
+
+// ensureConsumerRegistered registers the consumer if needed and returns the consumer ARN
+func (m *kinesisEFOManager) ensureConsumerRegistered(ctx context.Context) (string, error) {
+	if m.consumerARN != "" {
+		m.log.Infof("Using provided consumer ARN: %s", m.consumerARN)
+		return m.consumerARN, nil
+	}
+
+	m.log.Infof("Registering Enhanced Fan Out consumer: %s for stream: %s", m.consumerName, m.streamARN)
```
jem-davies: INFO?

jem-davies (commented):
Had a quick scan through - will look more in-depth very soon ... Also could we downgrade the log calls to debug maybe - it seems that is what the rest of the code does?

jbeemster:
@jem-davies we have re-opened this PR now from our official company fork (apologies) - https://github.com/warpstreamlabs/bento/pull/680

Its the same code with some minor tweaks that we have found during our testing. Its looking stable now however!

PR #674, opened 2026-01-26
feature: add duckdb support

sapk:
This PR add duckdb to sql available driver when build with `x_bento_extra` tag. 

I look at #365 for what has already be done. 
It is now fairly more simple since it is now managed by [duckdb](https://github.com/duckdb/duckdb-go) directly with pre-compiled binaries integrated and the license that goes along.

This superseed #365.

Let me know if you prefer that I rebase over @James-Gilbert- PR to give him credit. 
I already had done a poc of it before but I definitely re-use what he has done like the lint check or integration test.

### DIFF
--- a/internal/impl/sql/conn_fields.go
+++ b/internal/impl/sql/conn_fields.go
@@ -46,7 +46,7 @@ The ` + "`snowflake`" + ` driver supports multiple DSN formats. Please consult [
 
 The ` + "[`gocosmos`](https://pkg.go.dev/github.com/microsoft/gocosmos)" + ` driver is still experimental, but it has support for [hierarchical partition keys](https://learn.microsoft.com/en-us/azure/cosmos-db/hierarchical-partition-keys) as well as [cross-partition queries](https://learn.microsoft.com/en-us/azure/cosmos-db/nosql/how-to-query-container#cross-partition-query). Please refer to the [SQL notes](https://github.com/microsoft/gocosmos/blob/main/SQL.md) for details.
 
-The ` + "[`duckdb`](https://github.com/duckdb/duckdb-go)" + ` driver require cgo to link the DuckDB static library. It is only available in builds with the ` + "`x_bento_extra`" + ` build tag or the CGO-enabled Docker image (-cgo tag postfix).`).
+The ` + "[`duckdb`](https://github.com/duckdb/duckdb-go)" + ` driver requires cgo to link the DuckDB static library. It is only available in builds with the ` + "`x_bento_extra`" + ` build tag or the CGO-enabled Docker image (-cgo tag postfix).`).
 	Example("clickhouse://username:password@host1:9000,host2:9000/database?dial_timeout=200ms&max_execution_time=60").
 	Example("foouser:foopassword@tcp(localhost:3306)/foodb").
 	Example("postgres://foouser:foopass@localhost:5432/foodb?sslmode=disable").
--- a/website/docs/components/caches/sql.md
+++ b/website/docs/components/caches/sql.md
@@ -152,7 +152,7 @@ The `snowflake` driver supports multiple DSN formats. Please consult [the docs](
 
 The [`gocosmos`](https://pkg.go.dev/github.com/microsoft/gocosmos) driver is still experimental, but it has support for [hierarchical partition keys](https://learn.microsoft.com/en-us/azure/cosmos-db/hierarchical-partition-keys) as well as [cross-partition queries](https://learn.microsoft.com/en-us/azure/cosmos-db/nosql/how-to-query-container#cross-partition-query). Please refer to the [SQL notes](https://github.com/microsoft/gocosmos/blob/main/SQL.md) for details.
 
-The [`duckdb`](https://github.com/duckdb/duckdb-go) driver require cgo to link the DuckDB static library. It is only available in builds with the `x_bento_extra` build tag or the CGO-enabled Docker image (-cgo tag postfix).
+The [`duckdb`](https://github.com/duckdb/duckdb-go) driver requires cgo to link the DuckDB static library. It is only available in builds with the `x_bento_extra` build tag or the CGO-enabled Docker image (-cgo tag postfix).
 
 
 Type: `string`  
--- a/website/docs/components/inputs/sql_raw.md
+++ b/website/docs/components/inputs/sql_raw.md
@@ -157,7 +157,7 @@ The `snowflake` driver supports multiple DSN formats. Please consult [the docs](
 
 The [`gocosmos`](https://pkg.go.dev/github.com/microsoft/gocosmos) driver is still experimental, but it has support for [hierarchical partition keys](https://learn.microsoft.com/en-us/azure/cosmos-db/hierarchical-partition-keys) as well as [cross-partition queries](https://learn.microsoft.com/en-us/azure/cosmos-db/nosql/how-to-query-container#cross-partition-query). Please refer to the [SQL notes](https://github.com/microsoft/gocosmos/blob/main/SQL.md) for details.
 
-The [`duckdb`](https://github.com/duckdb/duckdb-go) driver require cgo to link the DuckDB static library. It is only available in builds with the `x_bento_extra` build tag or the CGO-enabled Docker image (-cgo tag postfix).
+The [`duckdb`](https://github.com/duckdb/duckdb-go) driver requires cgo to link the DuckDB static library. It is only available in builds with the `x_bento_extra` build tag or the CGO-enabled Docker image (-cgo tag postfix).
 
 
 Type: `string`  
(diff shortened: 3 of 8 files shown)


sapk:
The docker cgo image overhead is 53MB compared to main.
```
ghcr.io/warpstreamlabs/bento   1.14.0-15-g57b8faf72-cgo   93d17c8eda1e   419MB
ghcr.io/warpstreamlabs/bento   1.14.0-19-gfa4c18ce0-cgo   b177dfeab4c4    472MB
```

sapk:
I just added an explanation to why it is not available by default in the docs. I try to stay short and hesitate to add where to find all dsn parameters.

website/docs/components/processors/sql_select.md
```diff
@@ -157,13 +157,16 @@ The following is a list of supported drivers, their placeholder style, and their
...
 
 The [`gocosmos`](https://pkg.go.dev/github.com/microsoft/gocosmos) driver is still experimental, but it has support for [hierarchical partition keys](https://learn.microsoft.com/en-us/azure/cosmos-db/hierarchical-partition-keys) as well as [cross-partition queries](https://learn.microsoft.com/en-us/azure/cosmos-db/nosql/how-to-query-container#cross-partition-query). Please refer to the [SQL notes](https://github.com/microsoft/gocosmos/blob/main/SQL.md) for details.
 
+The [`duckdb`](https://github.com/duckdb/duckdb-go) driver require cgo to link the DuckDB static library. It is only available in builds with the `x_bento_extra` build tag or the CGO-enabled Docker image (-cgo tag postfix).
```
gregfurman:
```suggestion
The [`duckdb`](https://github.com/duckdb/duckdb-go) driver requires cgo to link the DuckDB static library. It is only available in builds with the `x_bento_extra` build tag or the CGO-enabled Docker image (-cgo tag postfix).
```

gregfurman:
Hey 👋  Thanks for following up on this integration! Were you able to verify that this works with the CGO Dockerfile?

> It is now fairly more simple since it is now managed by [duckdb](https://github.com/duckdb/duckdb-go) directly with pre-compiled binaries integrated and the license that goes along.

Also, yeah I'd appreciate if you could do rebase over of the original PR to give @James-Gilbert- credit that'd be much appreciated.

Once done, I think we can ship this :shipit:

sapk:
@gregfurman I tested with setting up sql_insert output to a ducklake (need some init_statement for setup) which is a "lakehouse" based on parquet with a standard sql (sqlite, duckdb or porstgres ) on top to achieve crud and is supported natively by duckdb. 
One limitation that I found in my poc, it fail to commit in // but it work well with `max_in_flight: 1`. But it is probably because I mostly use the default setup and it probably more efficient with some configuration adjustment (ex: data inlining).

I think it would be useful to add later duckdb appender to be more efficient and have a dedicated implementation for duckdb but let's first support the basic sql.

gregfurman:
@sapk Looks like we need to run `make docs` again. I don't have permissions to do this so just ping me once the CI is green 👀

sapk:
Sorry, I hadn't much time in the previous days. I should be able to work on it soon. I will ping you when this ready for review. (rebase and updated).

sapk:
Should be good now. I retested with:
```bash
make docker-cgo
docker run -ti --rm -v "$(pwd)/config/test_duckdb.yaml:/bento.yaml" -v "$(pwd)/tmp:/tmp" ghcr.io/warpstreamlabs/bento:1.15.0-24-g3a1e7cc95-cgo
```

With this test configuration:
```
input:
  generate:
    count: 10000
    interval: ""
    mapping: |
      root.id = fake("uuid_digit")
      root.timestamp = fake("unix_time")
      root.value = random_int(max: 1000)
      root.email = fake("email")
      root.name = fake("name")
      root.ip_address = fake("ipv4")
      root.date = root.timestamp.ts_format("2006-01-02", "UTC")
      root.year = root.timestamp.ts_format("2006", "UTC")

output:
  sql_insert:
    driver: duckdb
    dsn: "/tmp/test_data.duckdb?memory_limit=4GB&TimeZone=UTC&parquet_metadata_cache=true"
    table: test_data
    columns:
      - id
      - timestamp
      - value
      - email
      - name
      - ip_address
      - date
      - year
    args_mapping: |
      root = [
        this.id,
        this.timestamp,
        this.value,
        this.email,
        this.name,
        this.ip_address,
        this.date,
        this.year
      ]
    init_statement: |
      CREATE TABLE IF NOT EXISTS test_data (
        id VARCHAR,
        timestamp BIGINT,
        value INTEGER,
        email VARCHAR,
        name VARCHAR,
        ip_address VARCHAR,
        date VARCHAR,
        year VARCHAR
      )

```

sapk:
@gregfurman sorry I forgot to ping you in previous message. This pr should be good for review now.

gregfurman:
@sapk OK brilliant! I'll give this another look a little later today 🙂

internal/impl/sql/conn_fields.go
iamramtin, line 1: Have you had a chance to test in-memory mode? DuckDB supports `:memory:` as a DSN. Some users may expect to use this for ephemeral states but it's not documented here

internal/impl/sql/conn_fields_no_cgo.go
```diff
@@ -0,0 +1,7 @@
+//go:build !x_bento_extra
```
iamramtin, line 1: probably a nit, but the build tag here uses `!x_bento_extra` and not something like `!cgo`. Is there a scenario where a user sets `x_bento_extra` and `CGO_ENABLED=0`?

maybe worth adding just a comment at least, something like:

```
// x_bento_extra requires CGO_ENABLED=1 at build time.
```

internal/impl/sql/integration_cgo_test.go
```diff
@@ -0,0 +1,54 @@
...
+		if db != nil {
+			if err := db.Close(); err != nil {
+				t.Errorf("error closing duckdb database: %v", err)
+			}
+		}
+	})
+
+	require.Eventually(t, func() bool {
+		return db.Ping() == nil
+	}, time.Second, 100*time.Millisecond, "duckdb did not respond to ping in time")
```
iamramtin, line 39: DuckDB is embedded and should not need to wait for the ping? given that there's no network call it should just succeed on the first call, right?

if I'm not mistaken, I think you could just do `require.NoError(t, db.Ping())` instead

public/components/sql/duckdb.go
iamramtin, line 1: the file x_bento_extra.go is still empty. Do we expect DuckDB expected to appear there, or is the intent that it's implicitly included through this file alone via `package all`. I feel that leaving it empty may confuse maintainers reading x_bento_extra.go to understand what's being gated by the build flag and what's not?

internal/impl/sql/conn_fields.go
```diff
@@ -38,16 +38,20 @@ The following is a list of supported drivers, their placeholder style, and their
 ` + "| `spanner` | `projects/[project]/instances/[instance]/databases/dbname` |" + `
 ` + "| `trino` | [`http[s]://user[:pass]@host[:port][?parameters]`](https://github.com/trinodb/trino-go-client#dsn-data-source-name) |" + `
 ` + "| `gocosmos` | [`AccountEndpoint=<cosmosdb-endpoint>;AccountKey=<cosmosdb-account-key>[;TimeoutMs=<timeout-in-ms>][;Version=<cosmosdb-api-version>][;DefaultDb/Db=<db-name>][;AutoId=<true/false>][;InsecureSkipVerify=<true/false>]`](https://pkg.go.dev/github.com/microsoft/gocosmos#readme-example-usage) |" + `
+` + "| `duckdb` | `/path/to/filename.duckdb[?config_option=value&...]` |" + `
```
iamramtin, line 41: note, the example below uses a relative path but absolute here, not sure if this is intended, but this asymmetry appears in all the docs too

internal/impl/sql/conn_fields_no_cgo.go
gregfurman: FYI, this is already mentioned.

https://github.com/warpstreamlabs/bento/blob/c34cc6c7eef7c64bbc558a6fef9aad05dc6f345a/public/components/all/x_bento_extra.go#L5-L6

public/components/sql/duckdb.go
gregfurman: SInce duckdb is a part of the `sql` component, which should be in all, we don't expect it to be there.

If, for example, an entire package was _only_ ever used in CGO distributions then we would.

internal/impl/sql/conn_fields_no_cgo.go
iamramtin: whoops, missed this, thanks

gregfurman (approved):
Thanks for the superb PR @sapk! I think we should 100% look into the duckdb appender in a follow-up PR.

(also ty @iamramtin for the extra set of 👀 given I'm no duckdb expert)

internal/impl/sql/conn_fields.go
iamramtin: I actually ran simple test via the following bento config, looks like this works, we should just document it in a follow up iteration

```
input:
  generate:
    mapping: 'root.val = random_int() % 50'
    count: 3

pipeline:
  processors:
    - sql_raw:
        driver: duckdb
        dsn: ":memory:"
        query: "SELECT ? * 2 + 1 AS result"
        args_mapping: "root = [this.val]"
        init_statement: ""
        conn_max_open: 1

```

PR #678, opened 2026-01-27
add payload option for processor http and output http_client

### LINKED ISSUE #673: http_client body attribute with interpolation
Currently i have current message:

```
{"url":"ip", "payload": {json_data}}
```

It would be nice to have a body option in the http_client:

```
http_client:
  url: http://${!this.url}/
  verb: POST
  body: ${!this.payload} # not a multipart
```

lublak:
fixes #673

### DIFF
--- a/internal/impl/io/output_http_client.go
+++ b/internal/impl/io/output_http_client.go
@@ -27,7 +27,7 @@ It's also possible to set the body of the HTTP request using the optional field 
 
 It's possible to propagate the response from each HTTP request back to the input source by setting ` + "`propagate_response` to `true`" + `. Only inputs that support [synchronous responses](/docs/guides/sync_responses) are able to make use of these propagated responses.` + service.OutputPerformanceDocs(true, true)).
 		Field(httpclient.ConfigField("POST", true,
-			service.NewInterpolatedStringField("payload").Description("An alternative payload to deliver for each request.").Optional().Advanced(),
+			service.NewInterpolatedStringField("payload").Description("An alternative payload to deliver for each request.").Optional().Advanced().Version("1.16.0"),
 			service.NewBoolField("batch_as_multipart").
 				Description("Send message batches as a single request using [RFC1341](https://www.w3.org/Protocols/rfc1341/7_2_Multipart.html). If disabled messages in batches will be sent as individual requests.").
 				Advanced().Default(false),
--- a/internal/impl/io/processor_http.go
+++ b/internal/impl/io/processor_http.go
@@ -61,7 +61,7 @@ pipeline:
 `,
 		).
 		Field(httpclient.ConfigField("POST", false,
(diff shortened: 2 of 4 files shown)


internal/impl/io/output_http_client.go
jem-davies:
```suggestion
The URL, header and payload values of this output can be dynamically set using function interpolations described [here](/docs/configuration/interpolation#bloblang-queries).

By default, the body of the HTTP request is the raw contents of the message payload. If the message has multiple parts (is a batch) the request will be sent according to [RFC1341](https://www.w3.org/Protocols/rfc1341/7_2_Multipart.html). This behaviour can be disabled by setting the field ` + "[`batch_as_multipart`](#batch_as_multipart) to `false`" + `.

It's also possible to set the body of the HTTP request using the optional field ` + "[`payload`](#payload)" + `, which (if set) will take precedent.
```

internal/impl/io/processor_http.go
jem-davies, line 28:
```suggestion
The URL, header and payload values of this processor can be dynamically set using function interpolations described [here](/docs/configuration/interpolation#bloblang-queries).

By default, the body of the HTTP request is the raw contents of the message payload. It's also possible to set the body of the HTTP request using the optional field `+"[`payload`](#payload)"+`, which (if set) will take precedent.

In order to map or encode the payload to a specific request body, and map the response back into the original payload instead of replacing it entirely, you can use the `+"[`branch` processor](/docs/components/processors/branch)"+`.

```

internal/impl/io/processor_http.go
```diff
@@ -56,6 +56,7 @@ pipeline:
 `,
 		).
 		Field(httpclient.ConfigField("POST", false,
+			service.NewInterpolatedStringField("payload").Description("An alternative payload to deliver for each request.").Optional(),
```
jem-davies:
```suggestion
			service.NewInterpolatedStringField("payload").Description("An alternative payload to deliver for each request.").Optional().Advanced(),
```

internal/impl/io/output_http_client.go
```diff
@@ -23,6 +23,7 @@ The body of the HTTP request is the raw contents of the message payload. If the
 
 It's possible to propagate the response from each HTTP request back to the input source by setting ` + "`propagate_response` to `true`" + `. Only inputs that support [synchronous responses](/docs/guides/sync_responses) are able to make use of these propagated responses.` + service.OutputPerformanceDocs(true, true)).
 		Field(httpclient.ConfigField("POST", true,
+			service.NewInterpolatedStringField("payload").Description("An alternative payload to deliver for each request.").Optional(),
```
jem-davies:
```suggestion
			service.NewInterpolatedStringField("payload").Description("An alternative payload to deliver for each request.").Optional().Advanced(),
```

lublak:
@jem-davies all done

PR #680, opened 2026-01-30
Add support for AWS Kinesis Enhanced Fan-Out as an input

matus-tomlein:
Reopening PR #669 from @jbeemster

This pull request introduces a new consumption mode for Kinesis with Enhanced FanOut support. What this lets us do is provide a dedicated consumption pipe for Bento instead of depending on shared polling.

As much as possible we have tried to re-use the existing Kinesis input architectures and have only added EFO in terms of consuming the events but checkpointing + batching have been left identical. Quite a lot of the logical patterns have also been copied from the polling implementation and then altered to support EFO.

### DIFF
--- a/internal/impl/aws/input_kinesis_efo.go
+++ b/internal/impl/aws/input_kinesis_efo.go
@@ -19,13 +19,22 @@ import (
 // This is a retryable error that should trigger a backoff before resubscribing.
 var errBackpressureTimeout = errors.New("backpressure timeout waiting for space in pending pool")
 
+// kinesisEFOAPI is the subset of kinesis.Client methods used by kinesisEFOManager.
+type kinesisEFOAPI interface {
+	RegisterStreamConsumer(ctx context.Context, params *kinesis.RegisterStreamConsumerInput, optFns ...func(*kinesis.Options)) (*kinesis.RegisterStreamConsumerOutput, error)
+	DescribeStreamConsumer(ctx context.Context, params *kinesis.DescribeStreamConsumerInput, optFns ...func(*kinesis.Options)) (*kinesis.DescribeStreamConsumerOutput, error)
+}
+
 // kinesisEFOManager handles Enhanced Fan Out consumer registration and lifecycle
 type kinesisEFOManager struct {
 	streamARN    string
 	consumerName string
 	consumerARN  string
-	svc          *kinesis.Client
+	svc          kinesisEFOAPI
 	log          *service.Logger
+	// pollInterval controls how long waitForActiveConsumer waits between status
+	// checks. Defaults to 2 seconds; overridable in tests for faster iteration.
+	pollInterval time.Duration
 }
 
 // newKinesisEFOManager creates a new EFO manager
@@ -126,36 +135,44 @@ func (m *kinesisEFOManager) waitForActiveConsumer(ctx context.Context) error {
 	waiterCtx, cancel := context.WithTimeout(ctx, 2*time.Minute)
 	defer cancel()
 
-	ticker := time.NewTicker(2 * time.Second)
+	interval := m.pollInterval
+	if interval == 0 {
+		interval = 2 * time.Second
+	}
+	ticker := time.NewTicker(interval)
 	defer ticker.Stop()
 
 	for {
+		// Check consumer status immediately before waiting for the next tick
+		describeInput := &kinesis.DescribeStreamConsumerInput{
+			ConsumerARN: aws.String(m.consumerARN),
+		}
+
+		output, err := m.svc.DescribeStreamConsumer(waiterCtx, describeInput)
+		if err != nil {
+			return fmt.Errorf("failed to describe consumer: %w", err)
+		}
+
+		if output.ConsumerDescription != nil {
+			status := output.ConsumerDescription.ConsumerStatus
+			m.log.Debugf("Consumer status: %s", status)
+
+			if status == types.ConsumerStatusActive {
+				m.log.Debugf("Consumer is now ACTIVE")
+				return nil
+			}
+
+			if status == types.ConsumerStatusDeleting {
+				return errors.New("consumer is being deleted")
+			}
+		}
+
 		select {
+		case <-ctx.Done():
+			return fmt.Errorf("context cancelled waiting for consumer to become ACTIVE: %w", ctx.Err())
 		case <-waiterCtx.Done():
 			return fmt.Errorf("timeout waiting for consumer to become ACTIVE: %w", waiterCtx.Err())
 		case <-ticker.C:
-			describeInput := &kinesis.DescribeStreamConsumerInput{
-				ConsumerARN: aws.String(m.consumerARN),
-			}
-
-			output, err := m.svc.DescribeStreamConsumer(waiterCtx, describeInput)
-			if err != nil {
-				return fmt.Errorf("failed to describe consumer: %w", err)
-			}
-
-			if output.ConsumerDescription != nil {
-				status := output.ConsumerDescription.ConsumerStatus
-				m.log.Debugf("Consumer status: %s", status)
-
-				if status == types.ConsumerStatusActive {
-					m.log.Debugf("Consumer is now ACTIVE")
-					return nil
-				}
-
-				if status == types.ConsumerStatusDeleting {
-					return errors.New("consumer is being deleted")
-				}
-			}
 		}
 	}
 }
--- a/internal/impl/aws/input_kinesis_efo_test.go
+++ b/internal/impl/aws/input_kinesis_efo_test.go
@@ -1,0 +1,210 @@
+package aws
+
+import (
+	"context"
+	"errors"
+	"testing"
+	"time"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/service/kinesis"
+	"github.com/aws/aws-sdk-go-v2/service/kinesis/types"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/warpstreamlabs/bento/public/service"
+)
+
+// mockEFOClient implements kinesisEFOAPI for testing.
+type mockEFOClient struct {
+	describeResponses []mockDescribeResponse
+	describeCallCount int
+	// repeatLast causes the last response to be repeated indefinitely once all
+	// responses are exhausted. Useful for testing context cancellation.
+	repeatLast bool
+}
+
+type mockDescribeResponse struct {
+	status types.ConsumerStatus
+	err    error
+}
+
+func (m *mockEFOClient) DescribeStreamConsumer(_ context.Context, _ *kinesis.DescribeStreamConsumerInput, _ ...func(*kinesis.Options)) (*kinesis.DescribeStreamConsumerOutput, error) {
+	idx := m.describeCallCount
+	m.describeCallCount++
+	if idx >= len(m.describeResponses) {
+		if m.repeatLast && len(m.describeResponses) > 0 {
+			// Repeat the last response indefinitely (useful for context-cancel tests)
+			idx = len(m.describeResponses) - 1
+		} else {
+			// Default: return ACTIVE if we run out of responses
+			return &kinesis.DescribeStreamConsumerOutput{
+				ConsumerDescription: &types.ConsumerDescription{
+					ConsumerStatus: types.ConsumerStatusActive,
+				},
+			}, nil
+		}
+	}
+	resp := m.describeResponses[idx]
+	if resp.err != nil {
+		return nil, resp.err
+	}
+	return &kinesis.DescribeStreamConsumerOutput{
+		ConsumerDescription: &types.ConsumerDescription{
+			ConsumerStatus: resp.status,
+		},
+	}, nil
+}
+
+func (m *mockEFOClient) RegisterStreamConsumer(_ context.Context, _ *kinesis.RegisterStreamConsumerInput, _ ...func(*kinesis.Options)) (*kinesis.RegisterStreamConsumerOutput, error) {
+	return &kinesis.RegisterStreamConsumerOutput{
+		Consumer: &types.Consumer{
+			ConsumerARN: aws.String("arn:aws:kinesis:us-east-1:123:stream/test/consumer/bento-test"),
+		},
+	}, nil
+}
+
+func newTestEFOManager(client kinesisEFOAPI) *kinesisEFOManager {
+	return &kinesisEFOManager{
+		streamARN:    "arn:aws:kinesis:us-east-1:123:stream/test",
+		consumerARN:  "arn:aws:kinesis:us-east-1:123:stream/test/consumer/bento-test",
+		svc:          client,
+		log:          service.MockResources().Logger(),
+		pollInterval: 5 * time.Millisecond, // fast polling for unit tests
+	}
+}
+
+// TestWaitForActiveConsumer_AlreadyActive checks that if the consumer is already
+// ACTIVE on the first describe call, the function returns immediately without
+// waiting for a tick.
+func TestWaitForActiveConsumer_AlreadyActive(t *testing.T) {
+	client := &mockEFOClient{
+		describeResponses: []mockDescribeResponse{
+			{status: types.ConsumerStatusActive},
+		},
+	}
+	mgr := newTestEFOManager(client)
+
+	start := time.Now()
+	err := mgr.waitForActiveConsumer(context.Background())
+	elapsed := time.Since(start)
+
+	require.NoError(t, err)
+	assert.Equal(t, 1, client.describeCallCount, "should describe exactly once")
+	// Should return well under the poll interval (5ms in tests, 2s in production)
+	assert.Less(t, elapsed, 50*time.Millisecond, "should return immediately, not wait for tick")
+}
+
+// TestWaitForActiveConsumer_TransitionsToActive checks that the function polls
+// until the consumer becomes ACTIVE after a few CREATING responses.
+func TestWaitForActiveConsumer_TransitionsToActive(t *testing.T) {
+	client := &mockEFOClient{
+		describeResponses: []mockDescribeResponse{
+			{status: types.ConsumerStatusCreating},
+			{status: types.ConsumerStatusCreating},
+			{status: types.ConsumerStatusActive},
+		},
+	}
+	mgr := newTestEFOManager(client)
+
+	err := mgr.waitForActiveConsumer(context.Background())
+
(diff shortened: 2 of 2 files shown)


internal/impl/aws/input_kinesis_efo.go
```diff
@@ -0,0 +1,486 @@
...
+					}
+					return
+				} else {
+					// Schedule resubscription with continuation sequence
+					if continuationSeq != "" {
+						resubscribeChan <- continuationSeq
+					} else {
+						// Use latest checkpointed sequence
+						resubscribeChan <- recordBatcher.GetSequence()
+					}
```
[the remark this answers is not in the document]
matus-tomlein: Addressed in commit 90b6dd7 - added context cancellation checks before sending on any channels in the subscription goroutine.

internal/impl/aws/input_kinesis.go
```diff
@@ -141,6 +177,27 @@ Use the `+"`batching`"+` fields to configure an optional [batching policy](/docs
...
+				Description("Enable Enhanced Fan Out mode for push-based streaming with dedicated throughput.").
+				Default(false),
+			service.NewStringField(kiEFOFieldConsumerName).
+				Description("Consumer name for EFO registration. Auto-generated if empty: bento-clientID.").
+				Default("").
+				Optional(),
+			service.NewStringField(kiEFOFieldConsumerARN).
+				Description("Existing consumer ARN to use. If provided, skips registration.").
+				Default("").
+				Optional().
```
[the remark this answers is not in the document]
matus-tomlein: Addressed in commit 074a2cd - removed redundant Optional() from consumer_arn field.

internal/impl/aws/input_kinesis_efo.go
```diff
@@ -0,0 +1,486 @@
...
+					return
+				}
+
+				// Retryable error - backoff and retry
+				k.log.Warnf("EFO subscription error for shard %v, will retry: %v", shardID, err)
+				backoffDuration := boff.NextBackOff()
+				time.AfterFunc(backoffDuration, func() {
+					// Trigger resubscription after backoff
+					select {
+					case subscriptionTrigger <- recordBatcher.GetSequence():
```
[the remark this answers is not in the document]
matus-tomlein: Addressed in commit 9ee5f4a - captured sequence before the closure and added context cancellation check.

internal/impl/aws/input_kinesis.go
```diff
@@ -319,6 +378,18 @@ func newKinesisReaderFromConfig(conf kiConfig, batcher service.BatchPolicy, sess
...
+
+	// Check if Enhanced Fan Out is enabled
+	if k.conf.EnhancedFanOut != nil && k.conf.EnhancedFanOut.Enabled {
+		k.efoEnabled = true
+		k.log.Infof("Enhanced Fan Out enabled")
+
+		// Validate EFO configuration
+		if k.conf.EnhancedFanOut.ConsumerName != "" && k.conf.EnhancedFanOut.ConsumerARN != "" {
+			return nil, errors.New("cannot specify both consumer_name and consumer_arn in enhanced_fan_out config")
+		}
```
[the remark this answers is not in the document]
matus-tomlein: Addressed in commit 7184569 - removed the duplicate validation since it's already covered by the LintRule and newKinesisEFOManager.

internal/impl/aws/input_kinesis.go
```diff
@@ -141,6 +177,27 @@ Use the `+"`batching`"+` fields to configure an optional [batching policy](/docs
...
 			Description("Whether to consume from the oldest message when a sequence does not yet exist for the stream.").
 			Default(true),
+		service.NewObjectField(kiFieldEnhancedFanOut,
+			service.NewBoolField(kiEFOFieldEnabled).
+				Description("Enable Enhanced Fan Out mode for push-based streaming with dedicated throughput.").
+				Default(false),
+			service.NewStringField(kiEFOFieldConsumerName).
+				Description("Consumer name for EFO registration. Auto-generated if empty: bento-clientID.").
+				Default("").
+				Optional(),
```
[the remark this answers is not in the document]
matus-tomlein: Addressed in commit 074a2cd - removed redundant Optional() from consumer_name field.

internal/impl/aws/input_kinesis_efo.go
```diff
@@ -0,0 +1,486 @@
...
+			}
+		} else {
+			startingPosition = &types.StartingPosition{
+				Type: types.ShardIteratorTypeLatest,
+			}
+		}
+	} else {
+		// Continue from last sequence
+		startingPosition = &types.StartingPosition{
+			Type:           types.ShardIteratorTypeAfterSequenceNumber,
```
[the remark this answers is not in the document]
matus-tomlein: After investigation, the current implementation using `AFTER_SEQUENCE_NUMBER` appears to be correct.

The official AWS KCL (Kinesis Client Library) uses `AFTER_SEQUENCE_NUMBER` when resubscribing with a `ContinuationSequenceNumber`. From [IteratorBuilder.java](https://github.com/awslabs/amazon-kinesis-client/blob/master/amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/IteratorBuilder.java):

```java
public static StartingPosition.Builder reconnectRequest(...) {
  return apply(..., ShardIteratorType.AFTER_SEQUENCE_NUMBER);
}
```

According to the [AWS documentation](https://docs.aws.amazon.com/kinesis/latest/APIReference/API_SubscribeToShardEvent.html), `ContinuationSequenceNumber` "captures your shard progress even when no data is written to the shard" - meaning it represents the last position processed, not the next record to read. Using `AFTER_SEQUENCE_NUMBER` ensures we start reading after that position, avoiding re-processing.

Using `AT_SEQUENCE_NUMBER` would cause the record at `ContinuationSequenceNumber` to be re-delivered, potentially causing duplicates.

internal/impl/aws/input_kinesis_efo.go
```diff
@@ -0,0 +1,486 @@
...
+		shardFinishedChan := make(chan struct{}, 1)
+
+		var subscriptionWg sync.WaitGroup
+		subscriptionWg.Add(1)
+		go func() {
+			defer subscriptionWg.Done()
+			for sequence := range subscriptionTrigger {
+				continuationSeq, shardFinished, err := k.efoSubscribeAndStream(k.ctx, info, shardID, sequence, recordsChan)
+				if err != nil {
+					errorsChan <- err
```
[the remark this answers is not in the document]
matus-tomlein: After analysis, I believe the current implementation is correct.

The subscription goroutine follows this pattern:
1. Call `efoSubscribeAndStream()`
2. If error, send to `errorsChan` (blocking)
3. Loop back to `for sequence := range subscriptionTrigger` and **wait** for the next sequence

The goroutine cannot send multiple errors in rapid succession because after sending one error, it blocks waiting for a new sequence from `subscriptionTrigger`. A new sequence is only sent after the main loop processes the error and schedules a resubscription via `time.AfterFunc`.

The blocking send on `errorsChan` is intentional - it provides backpressure so the subscription goroutine waits until the error is processed before proceeding. This ensures errors are not lost and are handled in order.

With the buffer of 1, the flow is:
1. Error occurs → sent to buffer (non-blocking if buffer empty)
2. Goroutine waits for next sequence
3. Main loop processes error, schedules resubscription
4. New sequence sent → goroutine proceeds
5. Next error (if any) can use the now-empty buffer

The buffer of 1 is sufficient here and prevents unnecessary blocking on the happy path while maintaining correct error handling semantics.

internal/impl/aws/input_kinesis_efo.go
```diff
@@ -0,0 +1,486 @@
...
+	}
+
+	return &kinesisEFOManager{
+		streamARN:    streamARN,
+		consumerName: consumerName,
+		consumerARN:  conf.ConsumerARN,
+		svc:          svc,
+		log:          log,
+	}, nil
+}
```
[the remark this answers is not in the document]
jbeemster: Consumer de registration needs to be done very carefully. Consider the case where you have multiple pods in an auto-scaling architecture - if you scale-down do you want to issue a de-registration call and break your other pod that is still actively consuming?

For this reason (and similar to the DynamoDB table) - automated creation is fine and is a good helper. But destruction should be controlled by the user or an external process (internally we manage the consumer via Terraform before deploying the pods).

internal/impl/aws/input_kinesis_efo.go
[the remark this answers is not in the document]
matus-tomlein: Addressed in commit 536f46e - now tracking the last received record's sequence number as a fallback when `ContinuationSequenceNumber` is not provided.

The fix tracks `lastReceivedSeq` (the sequence number of the last record in each batch) and uses it as a fallback when `ContinuationSequenceNumber` is nil. This ensures we resume from the last received record rather than the last acknowledged record, avoiding re-delivery of in-flight records.

jem-davies:
@matus-tomlein - could you mark the PR as draft if changes are still happening? Also does this PR (#680) include #644? 

Could you add 644 to this PR?

matus-tomlein:
hey @jem-davies, sorry, we just came across an issue with some limited throughput due to the back-pressure mechanism when running this under load, so doing some small changes. Have marked it as draft and will let you know once we figure it out.

Yes, #644 is included in this PR.

jem-davies:
> hey @jem-davies, sorry, we just came across an issue with some limited throughput due to the back-pressure mechanism when running this under load, so doing some small changes. Have marked it as draft and will let you know once we figure it out.
> 
> Yes, #644 is included in this PR.

Ok great - still going to look at this - whilst you do that - want to get this in for the next release - thanks for the effort again 🙏

internal/impl/aws/input_kinesis_efo.go
```diff
@@ -0,0 +1,486 @@
...
+	if err := eventStream.Err(); err != nil {
+		// Check if it's just end of stream
+		if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
+			return continuationSeq, shardFinished, nil
+		}
+		return continuationSeq, shardFinished, fmt.Errorf("error receiving event: %w", err)
+	}
+
+	return continuationSeq, shardFinished, nil
+}
```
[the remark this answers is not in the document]
jem-davies: @matus-tomlein - could this be sorted with something like: https://github.com/snowplow-incubator/bento/pull/2 ?

jem-davies:
Would it be possible to do something like [this](https://github.com/snowplow-incubator/bento/pull/2) for an int test?

matus-tomlein:
hi @jem-davies, we have done some more load testing of the EFO implementation to work out some issues. in particular we have done the changes mentioned in this PR around introducing a new global pending pool and a self-sufficient error handling: https://github.com/snowplow-incubator/bento/pull/1

This PR should now be ready for review! We don't have any other issues that we are working on now, based on our testing it seems to be working well (we are currently running it on a couple of deployments).

jem-davies:
@matus-tomlein This PR should fix the CI: https://github.com/snowplow-incubator/bento/pull/3 - keen to get this in!

internal/impl/aws/input_kinesis_efo.go
```diff
@@ -0,0 +1,637 @@
...
+	waiterCtx, cancel := context.WithTimeout(ctx, 2*time.Minute)
+	defer cancel()
+
+	ticker := time.NewTicker(2 * time.Second)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-waiterCtx.Done():
+			return fmt.Errorf("timeout waiting for consumer to become ACTIVE: %w", waiterCtx.Err())
```
jem-davies: If the parent `ctx` is cancelled won't this error message be misleading?

internal/impl/aws/input_kinesis_efo.go
```diff
@@ -0,0 +1,637 @@
...
+		case <-ticker.C:
+			describeInput := &kinesis.DescribeStreamConsumerInput{
```
jem-davies: Could you do this before waiting for the tick? 

i.e.: 

for -> do the stuff -> select cases -> continue the loop? 

rather than waiting for the first tick?

internal/impl/aws/input_kinesis_checkpointer.go
```diff
@@ -180,54 +180,86 @@ type awsKinesisClientClaim struct {
...
-			if claim.LeaseTimeout, scanErr = time.Parse(time.RFC3339Nano, s.Value); scanErr != nil {
-				return nil, fmt.Errorf("failed to parse claim lease: %w", scanErr)
+		for _, item := range page.Items {
+			// Extract ShardID - required for all checkpoint entries
+			var shardID string
+			if s, ok := item["ShardID"].(*types.AttributeValueMemberS); ok {
+				shardID = s.Value
+			}
+			if shardID == "" {
+				continue
```
jem-davies, line 233: Before we have: 

```go
		if claim.ShardID == "" {
			return nil, errors.New("failed to extract shard id from claim")
```

But now we continue if shardID == "" ? 

Is that a change in behaviour?

internal/impl/aws/input_kinesis_efo.go
```diff
@@ -0,0 +1,637 @@
...
+					k.globalPendingPool.Release(len(records))
+				default:
+					return
+				}
+			}
+		}
+
+		var subscriptionWg sync.WaitGroup
+		subscriptionWg.Add(1)
+		go func() {
```
jem-davies:
```suggestion
subscriptionWg.Go(func() {
```

nit: this will be flagged on #701

internal/impl/aws/input_kinesis_efo.go
```diff
@@ -0,0 +1,637 @@
...
+					}
+					select {
+					case <-k.ctx.Done():
+						return
+					case resubscribeChan <- currentSeq:
+					}
+					break // Exit retry loop, wait for next trigger from main loop
+				}
+			}
+		}()
```
jem-davies:
```suggestion
		})
```

from the nit above re. `go fix`

internal/impl/aws/input_kinesis_pending_pool_test.go
```diff
@@ -0,0 +1,248 @@
...
+	// Start many goroutines acquiring and releasing
+	for i := 0; i < acquireCount; i++ {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			require.True(t, pool.Acquire(context.Background(), acquireSize))
+			time.Sleep(10 * time.Millisecond)
+			pool.Release(acquireSize)
+		}()
+	}
```
jem-davies, line 207:
```suggestion
	// Start many goroutines acquiring and releasing
	for range acquireCount {
		wg.Go(func() {
			require.True(t, pool.Acquire(context.Background(), acquireSize))
			time.Sleep(10 * time.Millisecond)
			pool.Release(acquireSize)
		})
	}
```

nit: `go fix`

internal/impl/aws/input_kinesis.go
```diff
@@ -141,9 +186,35 @@ Use the `+"`batching`"+` fields to configure an optional [batching policy](/docs
 		service.NewBoolField(kiFieldStartFromOldest).
 			Description("Whether to consume from the oldest message when a sequence does not yet exist for the stream.").
 			Default(true),
+		service.NewObjectField(kiFieldEnhancedFanOut,
```
jem-davies, line 189: It would be good to use the `.Version("1.16.0")` func here too.

internal/impl/aws/input_kinesis_efo.go
matus-tomlein: Fixed — the error message now distinguishes between the two cases: parent `ctx` cancellation produces `"context cancelled waiting for consumer to become ACTIVE"` and the internal 2-minute waiter timeout produces `"timeout waiting for consumer to become ACTIVE"`.

internal/impl/aws/input_kinesis_efo.go
matus-tomlein: Done — restructured the loop so the status check runs immediately at the top of each iteration, then waits in `select` for the next tick. On first entry this means we describe right away with no delay, which also avoids an unnecessary 2s wait when the consumer is already ACTIVE on restart.

internal/impl/aws/input_kinesis_checkpointer.go
matus-tomlein: Yes, intentional — added a comment to explain the reasoning. `GetCheckpointsAndClaims` scans all checkpoints in bulk, so skipping a single malformed row (no `ShardID`) is preferable to failing the entire query and blocking all shard progress. The single-item `getCheckpoint` path is correctly strict and still returns an error in that case.

internal/impl/aws/input_kinesis_efo.go
matus-tomlein: Applied — changed to `subscriptionWg.Go(func() {`.

internal/impl/aws/input_kinesis_efo.go
matus-tomlein: Applied.

internal/impl/aws/input_kinesis_pending_pool_test.go
matus-tomlein: Applied — updated to `for range acquireCount` with `wg.Go(func() {`.

internal/impl/aws/input_kinesis.go
matus-tomlein: Added `.Version("1.16.0")` to the `enhanced_fan_out` field.

internal/impl/aws/input_kinesis_efo.go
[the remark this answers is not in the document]
matus-tomlein: Not addressed in this PR — happy to follow up in a separate PR once this is merged, similar to the approach in snowplow-incubator/bento#2.

matus-tomlein:
Thanks for the review @jem-davies!

We have noticed another OOM issue when the EFO input is applied to a Kinesis stream with 60+ shards. We will be looking into that shortly. Sorry for the delay on this, will try to get that resolved asap!

jem-davies:
> Thanks for the review @jem-davies!
> 
> Sorry for the delay on this, will try to get that resolved asap!

No worries at all - this PR is a great benefit - thanks for contributing it back to Bento 😎

jem-davies:
CI fixes: https://github.com/snowplow-incubator/bento/pull/4 @matus-tomlein looking to get 👈  merged then I think we are done!

jem-davies:
Added lint / doc fixes to: https://github.com/warpstreamlabs/bento/pull/751 and merged! TYSM 🥳 


Raised: https://github.com/warpstreamlabs/bento/issues/753 -

Issue #649, opened 2026-01-09
Add watcher capability for Input / file component


mueckinger:
Current situation: bento consumes files in path and stops.
Enable a long running polling / file watcher process which creates new messages when new files are written to disk, similar to [sftp `watcher`](https://warpstreamlabs.github.io/bento/docs/components/inputs/sftp#watcher)
I guess this would require a lock-mechanism (e.g. lock-file) and  `delete_on_finish` or `move_on_finish` as proposed in #648 to be stateful about processed files.

jem-davies:
Yes I was thinking about this recently - I was thinking that you could have 'watcher' input component that triggers a child input when an event from [fsnotify](https://github.com/fsnotify/fsnotify) occurs.

henrikschristensen:
plus one for this. really need this too.
would be happy too help with the implementation if needed.

henrikschristensen:
Maybe this could be achieved with a fseventinput that delivers metadata about the filesystem event. In combination with existing components would it not be able to put together a combined input that can use the metadata to read a file, if that is what is desired?
Maybe "paths" in the file input would need to support function interpolation to make it work.

jem-davies:
> Maybe this could be achieved with a fseventinput that delivers metadata about the filesystem event. In combination with existing components would it not be able to put together a combined input that can use the metadata to read a file, if that is what is desired? Maybe "paths" in the file input would need to support function interpolation to make it work.

Yes this is what I was thinking - unsure there is a precedent for a child input receiving a message from a parent input - and then using it for interpolation though.

henrikschristensen:
> > Maybe this could be achieved with a fseventinput that delivers metadata about the filesystem event. In combination with existing components would it not be able to put together a combined input that can use the metadata to read a file, if that is what is desired? Maybe "paths" in the file input would need to support function interpolation to make it work.
> 
> Yes this is what I was thinking - unsure there is a precedent for a child input receiving a message from a parent input - and then using it for interpolation though.

Maybe it would be more natural if the fseventinput was followed by a file processor. I have missed that functionality before. It could have different operations. Read, write, delete, move etc.
That way the fseventinput would be more simple, and the file processor could have good use in other situations.

henrikschristensen:
could be implemented with.

 ```yaml
input:
  fsevent:
	paths:
	  - /somepath
  processors:
    - bloblang: |
		root = if metadata("fsevent_operation") == "WRITE" {
			content()
		} else {
			deleted()
		}
[...]

Issue #232, opened 2025-02-08
File monitoring input


mbneimann:
The current file input does not support monitoring or tailing for changes, so I suggest to either add this feature or create a new input.

This feature have been suggested before, but never implemented. Probably because of the work it is to create something that works on multiple platforms and can keep track of changes when files are moved, renamed, truncated, etc.

A shortcut for a feature like this would be to look at FileBeat from Elastic. FileBeat is written in Go, is open source under Apache 2.0, and actively maintained.

FileBeat works similar to Bento with inputs --> transformations --> outputs and has delivery guaranties between inputs and outputs, so the basic concepts are not far from each other.

https://www.elastic.co/guide/en/beats/filebeat/current/how-filebeat-works.html
https://github.com/elastic/beats/tree/main/filebeat
https://github.com/elastic/beats

I have used FileBeat on both Windows and Linux and found it very reliable at tracking file changes. I don't like how the rest of FileBeat works and that the outputs are obviously very tailored for ending up in Elasticsearch.

I am not (yet) a Go developer, so can't really point to exactly how to leverage the inspiration from the FileBeat source code. And I am no expert in what is allowed with the software under the Apache 2.0 license.

My first idea however is that it should be possible to make a Bento input that imports and wraps the relevant parts of the filebeat input that reads from files and keeps the state.

jem-davies:
https://github.com/warpstreamlabs/bento/issues/181 related issue

jem-davies:
https://github.com/warpstreamlabs/bento/pull/645

jem-davies:
We have the https://warpstreamlabs.github.io/bento/docs/components/inputs/file_tail & https://github.com/warpstreamlabs/bento/pull/688 - that should close this ticket

Issue #181, opened 2024-12-02
Handle missing files in `csv.paths`


sepastian:
Hello, thanks for keeping Bento alive!

I have an issue trying to create a stream with optional files, i.e. a CSV may exists, but there's no guarantee.

If the file exists, everything is running as expected. If a file is missing, Bento fails to start with the following error.

```
{"@service":"bento","level":"error","msg":"Failed to create stream (input-data): \
  failed to init input input-data path root.input: requires at least one input file path"}
```

Here's my config, i'm leaving out CSV options not relevant here:

```
input:
  label: "input-data"
  csv:
    paths:
      - /data/*_data.csv
```

The name of the file contains a date field, e.g. `/data/20241202_data.csv`.

So my question is: is it possible to create a stream, if the corresponding file `/data/*_data.csv` does not exist?

More generally, for my understanding: does bento listen for new files? Or is Bento starting, processing a batch of existing files, then shutting down again?

Thanks for any clues on this!

jem-davies:
"More generally, for my understanding: does bento listen for new files? Or is Bento starting, processing a batch of existing files, then shutting down again?"

Bento currently doesn't have the ability to listen for new files, we could add a new input component that could make use of [fs_notify](https://github.com/fsnotify/fsnotify). 

And yeah, Bento will spin up - process the files in the given path - then shutdown, if there isn't a single file in the given path the stream won't be created... 

Suggestions for ways forward with current functionality maybe: 

- [dynamic input](https://warpstreamlabs.github.io/bento/docs/components/inputs/dynamic)
- [streams mode](https://warpstreamlabs.github.io/bento/docs/guides/streams_mode/about) 

both of these things essentially enable the creation of streams via a REST API, such that you could potentially create the stream, when you know the file to exist... that may or not be ok depending on your specific circumstances. 

Happy to look into a file_watcher component soon - is this a complete blocker - what would the timelines be for this?

jem-davies:
https://github.com/warpstreamlabs/bento/pull/688

PR #688, opened 2026-02-09
Implement filesystem watcher input

henrikschristensen:


henrikschristensen:
resolves #649

internal/impl/io/input_fsevent.go
```diff
@@ -0,0 +1,186 @@
...
+
+	f.watcher = watcher
+	f.eventChan = eventChan
+	return nil
+}
+
+func (f *fsEventWatcher) Read(ctx context.Context) (*service.Message, service.AckFunc, error) {
+	f.cMut.Lock()
+	eventChan := f.eventChan
+	f.cMut.Unlock()
```
gregfurman: If `f.eventChan` is set to `nil`, then won't we potentially run into a panic when reading from a `nil` channel here:

```golang
case msg, open := <-eventChan:
```

Perhaps guard against this with a:

```suggestion
	f.cMut.Lock()
	eventChan := f.eventChan
	if f.eventChan == nil {
	   return nil, nil, service.ErrNotConnected
	}
	f.cMut.Unlock()
```

internal/impl/io/input_fsevent.go
```diff
@@ -0,0 +1,186 @@
...
+		return err
+	}
+
+	for _, path := range f.paths {
+		err = watcher.Add(path)
+		if err != nil {
+			f.log.Errorf("Failed to add path %v: %s", path, err)
+			return err
+		}
+	}
```
gregfurman, line 169: If all paths no longer exist, we end up with a watcher that does nothing i.e

Given the config:

```yaml
# watch-config.yaml
input:
  fsevent:
    paths: [ephemeral-dir]
```

If we run this against an existing dir, then delete that dir, the input continues on despite not having anything to watch:

```bash
mkdir ephemeral-dir
bento -c watch-config.yaml &

touch ephemeral-dir/temp
# Out: {"fsevent_mod_time":"2026-02-09T12:48:58+02:00","fsevent_mod_time_unix":1770634138,"fsevent_operation":"CREATE","fsevent_path":"ephemeral-dir/temp"}

rm -rf ephemeral-dir 

# Out: {"fsevent_mod_time":"2026-02-09T12:49:07+02:00","fsevent_mod_time_unix":1770634147,"fsevent_operation":"REMOVE","fsevent_path":"ephemeral-dir"}

mkdir ephemeral-dir
# NO OUTPUT

touch ephemeral-dir/temp
# NO OUTPUT
```

internal/impl/io/input_fsevent.go
```diff
@@ -0,0 +1,186 @@
...
+	return nil
+}
+
+func (f *fsEventWatcher) Read(ctx context.Context) (*service.Message, service.AckFunc, error) {
+	f.cMut.Lock()
+	eventChan := f.eventChan
+	f.cMut.Unlock()
+
+	select {
+	case msg, open := <-eventChan:
```
gregfurman, line 234: `f.eventChan` is never closed. Do you perhaps mean to close it after that goroutine finishes?

https://github.com/henrikschristensen/bento/blob/f3bf54d14b48e5f045a870aec05e72baf41ba1d7/internal/impl/io/input_fsevent.go#L123-L124

i.e 

```golang
	go func() {
		defer close(eventChan)
		for {
// ...
```

internal/impl/io/input_fsevent.go
```diff
@@ -0,0 +1,186 @@
...
+	case <-ctx.Done():
+		return nil, nil, ctx.Err()
+	}
+}
+
+func (f *fsEventWatcher) Close(ctx context.Context) error {
+	f.cMut.Lock()
+	defer f.cMut.Unlock()
+
+	err := f.watcher.Close()
```
gregfurman: Can we make `Close()` idempotent? If I call it after the component has already closed (and the `f.watcher == nil`) the call will panic.

internal/impl/io/input_fsevent.go
```diff
@@ -0,0 +1,186 @@
...
+		message := service.NewMessage(nil)
+		message.MetaSetMut("fsevent_operation", msg.event.Op.String())
+		message.MetaSetMut("fsevent_path", msg.event.Name)
+		message.MetaSetMut("fsevent_mod_time_unix", msg.timestampUnix)
+		timestamp := time.Unix(msg.timestampUnix, 0).Format(time.RFC3339)
+		message.MetaSetMut("fsevent_mod_time", timestamp)
+
+		return message, func(ctx context.Context, res error) error {
+			return nil
+		}, nil
```
gregfurman, line 264: Behaviourally, should we be closing the input if all filepaths we want to watch no longer exist? Or should we be retrying until the paths are present again?

This approach is a bit ham-fisted but we could go:

```golang
	select {
	case msg, open := <-eventChan:
		if !open {
			f.cMut.Lock()
			f.eventChan = nil
			f.watcher = nil
			f.cMut.Unlock()
            // NOTE: no files being watched means input is exhausted
		}
    // ...
		if len(f.watcher.WatchList()) == 0 {
            // FIXME: this should only close once
			f.log.Warn("Nothing being watched.")
			defer f.Close(context.TODO())
		}
		return message, func(ctx context.Context, res error) error {
			return nil
		}, nil
   // ...
```

internal/impl/io/input_fsevent.go
```diff
@@ -0,0 +1,186 @@
...
+	}
+
+	go func() {
+		for {
+			select {
+			case event, ok := <-watcher.Events:
+				if !ok {
+					return
+				}
+				msg := watcherEventMsg{
```
gregfurman, line 204: If a subdir is created, its contents wont be watched i.e

```sh
mkdir sandbox
bento -c watch-config.yaml &

touch sandbox/file1
mkdir sandbox/subdir
touch sandbox/subdir/file2
```

gives us:

```jsonl
{"fsevent_mod_time":"2026-02-09T13:46:44+02:00","fsevent_mod_time_unix":1770637604,"fsevent_operation":"CREATE","fsevent_path":"sandbox/file1"}
{"fsevent_mod_time":"2026-02-09T13:46:48+02:00","fsevent_mod_time_unix":1770637608,"fsevent_operation":"CREATE","fsevent_path":"sandbox/subdir"}
```

internal/impl/io/input_fsevent.go
```diff
@@ -0,0 +1,186 @@
...
+		if err != nil {
+			f.log.Errorf("Failed to add path %v: %s", path, err)
+			return err
+		}
+	}
+
+	go func() {
+		for {
+			select {
+			case event, ok := <-watcher.Events:
```
gregfurman, line 175: I'm wondering if we should be dynamically modifying what is being watched within this loop. Like if a new subdir is created, we should be globbing and adding it's contents to the watch list (via `f.watcher.Add()`).

I was working on a similar component a while back and that did some logic depending on the operation type. I think a similar approach could be taken to add/remove what is being watched.

See https://github.com/warpstreamlabs/bento/blob/1a7355c1f2e4426784997c9b824a74a13781b8f3/internal/impl/io/input_file_watcher.go#L154-L184

Tbh, it could just be easier to refresh every `T` seconds (or ms) and re-create the watcher based on the new state of the FS.

Lmk your thoughts!

gregfurman (changes requested):

internal/impl/io/input_fsevent.go
henrikschristensen: if it is a file being watched, that is perhaps being read and removed. And then recreated with a different content, then I guess we should keep watching. Otherwise we will end up with a similar problem as with the current file input that closes immediately after all files have been processed.

internal/impl/io/input_fsevent.go
henrikschristensen: Should watching subdirs be something that could be turned on with a boolean config setting?
I think I would personally prefer it to be turned off by default.

internal/impl/io/input_fsevent.go
henrikschristensen: Again I think I would personally like something like this to be something not turned on by default, as it could lead to some surprising behaviour. But as a user if I know that is what I want then certainly a great addition.

internal/impl/io/input_fsevent.go
henrikschristensen: added the guard

internal/impl/io/input_fsevent.go
henrikschristensen: Added empty watchlist detection.

internal/impl/io/input_fsevent.go
henrikschristensen: good catch. I added the deferred close in the goroutine

internal/impl/io/input_fsevent.go
henrikschristensen: Added a check to see if the watcher has already been closed

internal/impl/io/input_fsevent.go
henrikschristensen: Changed my mind. :) I think the best options is to close the input with a warning as suggested.

internal/impl/io/input_fsevent.go
henrikschristensen: Took a shot and added a bool field watch_new_subdirs. Default false. I will look for create events and determine if it is a directory being created. if yes it is added to the watcher.

internal/impl/io/input_fsevent.go
henrikschristensen: I think this should be covered with the above approach

internal/impl/io/input_fsevent.go
henrikschristensen: I also added a switch "is_recursive" to enable initial recursive watch

internal/impl/io/input_fsevent.go
```diff
@@ -0,0 +1,248 @@
...
+
+		message := service.NewMessage(nil)
+		message.MetaSetMut("fsevent_operation", msg.event.Op.String())
+		message.MetaSetMut("fsevent_path", msg.event.Name)
+		message.MetaSetMut("fsevent_mod_time_unix", msg.timestampUnix)
+		timestamp := time.Unix(msg.timestampUnix, 0).Format(time.RFC3339)
+		message.MetaSetMut("fsevent_mod_time", timestamp)
+
+		// check to see if the watchlist is empty.
+		f.cMut.Lock()
```
gregfurman: Can we use a `sync.RWMutex` instead? Then we can replace these (more expensive) write locks with `RLock` since these calls are not a protecting explicit writes in memory that need locking.

internal/impl/io/input_fsevent.go
```diff
@@ -0,0 +1,248 @@
...
+		if f.watcher != nil && len(f.watcher.WatchList()) == 0 {
+			f.log.Warn("Nothing being watched. Closing the input.")
+			f.cMut.Unlock()
+			defer f.Close(context.TODO())
```
gregfurman: I assume this is to ensure the lock on `Close` doesn't deadlock?

IMO a cleaner approach would be to trigger this close in a goroutine and allow the `f.cMut.Unlock` to be deferred. i.e

```go
f.cMut.Lock()
defer f.cMut.Unlock()

if f.watcher != nil && len(f.watcher.WatchList()) == 0 {
   go f.Close(context.TODO())
}
// ...
```

(Already mentioned, but these should probably be `RLock` and `RUnlock` calls)

internal/impl/io/input_fsevent.go
```diff
@@ -0,0 +1,248 @@
...
+}
+
+func (f *fsEventWatcher) Close(ctx context.Context) error {
+	f.cMut.Lock()
+	defer f.cMut.Unlock()
+
+	var err error = nil
+	if f.watcher != nil {
+		err = f.watcher.Close()
+		f.watcher = nil
```
gregfurman, line 277: This `nil` assignment will happen anyway in the Read loop, right? Think we should be doing this in one or the other.

internal/impl/io/input_fsevent.go
```diff
@@ -0,0 +1,248 @@
...
+	case <-ctx.Done():
+		return nil, nil, ctx.Err()
+	}
+}
+
+func (f *fsEventWatcher) Close(ctx context.Context) error {
+	f.cMut.Lock()
+	defer f.cMut.Unlock()
+
+	var err error = nil
```
gregfurman: Surprised the linter didn't complain about this 🤔 Since an `error` is an interface, declaring it like this automatically sets it to `nil`.


```suggestion
	var err error
```

internal/impl/io/input_fsevent.go
```diff
@@ -0,0 +1,248 @@
...
+	if f.watcher != nil {
+		err = f.watcher.Close()
```
gregfurman, line 276: q: Is calling `Close` twice on a watcher allowed?

internal/impl/io/input_fsevent.go
```diff
@@ -0,0 +1,248 @@
...
+					if err == nil {
+						if st.IsDir() && !slices.Contains(f.watcher.WatchList(), event.Name) {
+							err = f.watcher.Add(event.Name)
+							if err != nil {
+								f.log.Warnf("Failed to add path %v: %s", event.Name, err)
+							}
+						}
+					} else {
+						f.log.Warnf("Cannot check for new subpath: %s", err)
+					}
```
gregfurman, line 197: Can we unnest this? Unsure if I got the logic right but something like the below is way easier to read (also maybe add a comment for that weedy `st.IsDir() && ...` check)


```golang
				if f.watchNewSubdirs && event.Has(fsnotify.Create) {
					st, err := os.Stat(event.Name)
					if err != nil {
						f.log.Warnf("Cannot check for new subpath: %s", err)
						continue
					}
					
					if !st.IsDir() || slices.Contains(f.watcher.WatchList(), event.Name) {
					    continue
					}
					
					if err := f.watcher.Add(event.Name); err != nil {
						f.log.Warnf("Failed to add path %v: %s", event.Name, err)
					}
					
```

internal/impl/io/input_fsevent.go
```diff
@@ -0,0 +1,248 @@
...
+
+func init() {
+	err := service.RegisterInput("fsevent", fsEventInputSpec(),
+		func(pConf *service.ParsedConfig, res *service.Resources) (service.Input, error) {
+			r, err := fsEventWatcherFromParsed(pConf, res)
+			if err != nil {
+				return nil, err
+			}
+			return r, nil
+		})
```
gregfurman, line 70: nit: can we not just replace all of this with:


```suggestion
	err := service.RegisterInput("fsevent", fsEventInputSpec(),
		func(pConf *service.ParsedConfig, res *service.Resources) (service.Input, error) {
			return  fsEventWatcherFromParsed(pConf, res)
		})
```

internal/impl/io/input_fsevent.go
```diff
@@ -0,0 +1,248 @@
...
+		return err
+	}
+
+	for _, path := range f.paths {
+		err = watcher.Add(path)
+		if err != nil {
+			f.log.Errorf("Failed to add path %v: %s", path, err)
+			return err
+		}
+		if f.resursive {
```
gregfurman: 🧹 spelling

```suggestion
		if f.recursive {
```

gregfurman (changes requested):

internal/impl/io/input_fsevent.go
henrikschristensen: yes. the backend implementations in fsnotify has guards for that

internal/impl/io/input_fsevent.go
henrikschristensen: i did the refactor. actually found out that adding the same path more than once is a noop with nil error. so the could actually be even more simple

internal/impl/io/input_fsevent.go
henrikschristensen: yes. changed it

internal/impl/io/input_fsevent.go
henrikschristensen: corrected

internal/impl/io/input_fsevent.go
henrikschristensen: I have habits from 6-7 different languages. :)  

removed it.

internal/impl/io/input_fsevent.go
henrikschristensen: removed it

internal/impl/io/input_fsevent.go
henrikschristensen: changed it to a read/write lock. and use rlock and runlock in the read method

internal/impl/io/input_fsevent.go
henrikschristensen: changed into suggested

internal/impl/io/input_fsevent.go
```diff
@@ -0,0 +1,251 @@
...
+		message.MetaSetMut("fsevent_mod_time", timestamp)
+
+		// Try to get file info, but don't fail if the file doesn't exist (e.g., for DELETE events)
+		st, err := os.Stat(msg.event.Name)
+		if err == nil {
+			message.MetaSetMut("fsevent_is_dir", st.IsDir())
+		} else {
+			// For deleted files, we can't stat them, so set is_dir to false
+			message.MetaSetMut("fsevent_is_dir", false)
+		}
```
gregfurman:
```suggestion
		var isDir bool
		if st, err := os.Stat(msg.event.Name); err == nil {
		    isDir == st.IsDir()
		}
		message.MetaSetMut("fsevent_is_dir", isDir)
```

internal/impl/io/input_fsevent.go
```diff
@@ -0,0 +1,251 @@
...
+
+	f.watcher = watcher
+	f.eventChan = eventChan
+	return nil
+}
+
+func (f *fsEventWatcher) Read(ctx context.Context) (*service.Message, service.AckFunc, error) {
+	f.cMut.RLock()
+	defer f.cMut.RUnlock()
+	eventChan := f.eventChan
```
gregfurman, line 226: You should lock here once to retrieve the pointer to the eventChan

```suggestion
	f.cMut.RLock()
	eventChan := f.eventChan
	f.cMut.RUnlock()
```

internal/impl/io/input_fsevent.go
```diff
@@ -0,0 +1,251 @@
...
+
+	if eventChan == nil {
+		return nil, nil, service.ErrNotConnected
+	}
+
+	select {
+	case msg, open := <-eventChan:
+		if !open {
+			f.eventChan = nil
+			f.watcher = nil
```
gregfurman, line 239: This needs a write lock:


```suggestion
            f.cMut.Lock()
            defer f.cMut.Unlock()
			f.eventChan = nil
			f.watcher = nil
```

internal/impl/io/input_fsevent.go
```diff
@@ -0,0 +1,251 @@
...
+			message.MetaSetMut("fsevent_is_dir", st.IsDir())
+		} else {
+			// For deleted files, we can't stat them, so set is_dir to false
+			message.MetaSetMut("fsevent_is_dir", false)
+		}
+
+		// check to see if the watchlist is empty.
+		if f.watcher != nil && len(f.watcher.WatchList()) == 0 {
+			f.log.Warn("Nothing being watched. Closing the input.")
+			go f.Close(context.TODO())
```
gregfurman, line 259: 👍

internal/impl/io/input_fsevent_test.go
```diff
@@ -0,0 +1,503 @@
...
+			continue
+		}
+	}
+
+	assert.True(t, subdirCreated, "Should have received CREATE event for subdirectory")
+	assert.False(t, fileCreated, "Should NOT have received event for file in subdirectory when watch_new_subdirs is disabled")
+}
+
+func TestFSEventWatchNewSubdirsDeleteRecreate(t *testing.T) {
+	dir := t.TempDir()
```
gregfurman, line 326: Is there a chance we can use the `testing/fstest` instead of actually creating files/dirs on the OS?

gregfurman (changes requested):

internal/impl/io/input_fsevent.go
henrikschristensen: refactored to suggested

internal/impl/io/input_fsevent.go
henrikschristensen: refactored to suggested

internal/impl/io/input_fsevent.go
henrikschristensen: refactored to suggested

internal/impl/io/input_fsevent_test.go
henrikschristensen: I dont think that is possible unless we write a mock for fsnotify too that can produce filesystem events when changes are made to the MapFS in testing/fstest. 
I have not been able to find any examples of doing that. 
I am aware of the dilemma that the tests rely on the actual OS that it is executed on, but abstracting that part out into some kind of mock would be a lot of work i fear.
I assume fsnotify itself is tested cross platform.

henrikschristensen:
@gregfurman 

It is a known problem that fsnotify produces a lot of redundant events when f.ex. a file is saved to a dir. Then it will typical produce a CREATE event followed by several WRITE events depending on file size and maybe a CHMOD event. This makes it hard to detect when a file done being written to the disk and as a result u risk reading parts of the file or empty content.
The suggested solution from the fsnotify project is a simple deduplication/debouncing like this example https://github.com/fsnotify/fsnotify/blob/main/cmd/fsnotify/dedup.go. 

I have tried to do something similar with existing bento components but I think it makes the resulting yaml very complicated and error prone.
So I think the best approach for this input is to implement something like the suggested dedup code.
An alternative could be to refactor the code into using this project https://github.com/sgtdi/fswatcher witch has buildin support for dedup and batching of events. But it is a single developer project and maybe not a good option for Bento.

Any thoughts on this?

jem-davies:
https://github.com/warpstreamlabs/bento/issues/181

jem-davies:
Fixes #232

henrikschristensen:
I have been running this input in production for a while now. Works great but had to add the following features to make it usable in a live system.

1. Write Deduplication
    • Added write_dedup_timeout field to suppress rapid WRITE events
    • When enabled, only fires a single WRITTEN event after the timeout expires
    • Uses timer-based deduplication to reduce event noise
2. File Extension Filtering
    • Added extensions field to filter events by file extensions
    • Supports patterns like [".txt", ".log"]
    • Only emits events for files matching the allowed extensions
3. Existing Files Polling
    • Added existing_files_poll_interval to scan directories periodically
    • Added existing_files_min_age to skip recently modified files
    • Emits CREATE events for pre-existing files that meet age criteria
    • Tracks scanned files to avoid duplicate events

gregfurman:
@henrikschristensen So sorry I missed your last update. I was changing roles around that time and completely missed your comment.

> It is a known problem that fsnotify produces a lot of redundant events [...]

I was going to suggest a `write_dedupe_window` or something similar. Seems you already went with this approach based on your latest update 😄

>  • Added write_dedup_timeout field to suppress rapid WRITE events

I'm curious what the resulting config would've looked like when using a bento cache + dedupe (see [example](https://warpstreamlabs.github.io/bento/docs/components/processors/dedupe/#examples)) and whether this would've solved the issue 👀 

Otherwise, I'll give the code a final pass now that you're implemented this new functionality. Really appreciate your patience on this with my gauntlet of a review cycle 🙏

henrikschristensen:
@gregfurman  Actually this does solve the issue and it is not too bad I think.
``` yaml   
sequence:
        inputs:
          - file:
              paths:
                - /path/to/first/*.xml
                - /path/to/first/*.txt
                - /path/to/second/*.xml
                - /path/to/second/*.txt
              scanner:
                to_the_end: {}
          - fsevent:
              paths: ["/path/to/first", "/path/to/second"]
              is_recursive: false
              watch_new_subdirs: false
              extensions:
                - xml
                - txt
            processors:
              - dedupe:
                  cache: trades_file_dedup
                  key: '${! metadata("fsevent_path") }'
                  drop_on_err: false
              - sleep:
                  duration: 1s

cache_resources:
  - label: trades_file_dedup
    memory:
      default_ttl: 60s

```

so I will remove the dedup logic from the code again as it does add quite a bit of complication.

internal/impl/io/input_fsevent.go
```diff
@@ -0,0 +1,280 @@
...
+
+const (
+	fsEventInputFieldPaths           = "paths"
+	fsEventInputFieldIsRecursive     = "is_recursive"
+	fsEventInputFieldWatchNewSubdirs = "watch_new_subdirs"
+	fsEventInputFieldExtensions      = "extensions"
+)
+
+func fsEventInputSpec() *service.ConfigSpec {
+	return service.NewConfigSpec().
```
gregfurman, line 24: suggestion(non-blocking): Perhaps we provide the example config you gave for deduplication using bento config

gregfurman (approved):
LGTM 🤩 Thanks for addressing all of my comments :shipit:

internal/impl/io/input_fsevent.go
gregfurman: i.e https://github.com/warpstreamlabs/bento/pull/688#issuecomment-4280034707

PR #689, opened 2026-02-09
Implement file processor

### LINKED ISSUE #649: Add watcher capability for Input / file component

henrikschristensen:
resolves #649


henrikschristensen:
gonna try and reopen without that workflow approval thing.

PR #690, opened 2026-02-09
Implement file processor

### LINKED ISSUE #649: Add watcher capability for Input / file component

henrikschristensen:
resolves #649


internal/impl/io/processor_file.go
```diff
@@ -0,0 +1,331 @@
...
+	}
+	path = filepath.Clean(path)
+
+	file, err := p.nm.FS().Open(path)
+	if err != nil {
+		return nil, fmt.Errorf("failed to open file '%s': %w", path, err)
+	}
+	defer file.Close()
+
+	fileInfo, err := file.Stat()
```
jem-davies:
```suggestion
	fileInfo, err := p.nm.FS().Stat(path)
```

internal/impl/io/processor_file.go
```diff
@@ -0,0 +1,331 @@
...
+	}
+
+	if err := p.nm.FS().Remove(srcPath); err != nil {
+		return nil, fmt.Errorf("failed to delete source file '%s': %w", srcPath, err)
+	}
+
+	return service.MessageBatch{msg}, nil
+}
+
+func (p *fileProcessor) processStat(ctx context.Context, msg *service.Message) (service.MessageBatch, error) {
```
jem-davies:
```suggestion
func (p *fileProcessor) processStat(_ context.Context, msg *service.Message) (service.MessageBatch, error) {
```

internal/impl/io/processor_file.go
```diff
@@ -0,0 +1,331 @@
...
+	path = filepath.Clean(path)
+
+	if err := p.nm.FS().Remove(path); err != nil {
+		return nil, fmt.Errorf("failed to delete file '%s': %w", path, err)
+	}
+
+	return service.MessageBatch{msg}, nil
+}
+
+func (p *fileProcessor) processMove(ctx context.Context, msg *service.Message) (service.MessageBatch, error) {
```
jem-davies, line 316: Would be good to see if we could use os.Rename in cases it would make sense to?

internal/impl/io/processor_file.go
```diff
@@ -0,0 +1,331 @@
...
+
+- **read**: Read file content into the message
+- **write**: Write message content to a file
+- **delete**: Delete a file
+- **move**: Move/rename a file
+- **stat**: Get file information without reading content
+
+### Metadata
+
+When reading or getting file info (stat), this processor adds the following metadata fields:
```
jem-davies, line 56: Here we are saying that when "reading" we add the following metadata fields but at least some of the fields are missing in the func `processRead` such as "file_name" - perhaps add a function for the metadata so we ensure consistency.

internal/impl/io/processor_file.go
```diff
@@ -0,0 +1,331 @@
...
+	newMsg.MetaSetMut("file_path", path)
+	newMsg.MetaSetMut("file_size", fileInfo.Size())
+	newMsg.MetaSetMut("file_mod_time_unix", fileInfo.ModTime().Unix())
+	newMsg.MetaSetMut("file_mod_time", fileInfo.ModTime().Format(time.RFC3339))
+
+	return service.MessageBatch{newMsg}, nil
+}
+
+func (p *fileProcessor) processWrite(ctx context.Context, msg *service.Message) (service.MessageBatch, error) {
+	path, err := p.conf.Path.TryString(msg)
```
jem-davies, line 249: Perhaps would change this to get the message bytes and write that to the file rather than an option interpolation field - then mention the branch processor similar to the http processor in the docs like: 

[this](https://warpstreamlabs.github.io/bento/docs/components/processors/http#:~:text=In%20order%20to%20map%20or%20encode%20the%20payload%20to%20a%20specific%20request%20body%2C%20and%20map%20the%20response%20back%20into%20the%20original%20payload%20instead%20of%20replacing%20it%20entirely%2C%20you%20can%20use%20the%20branch%20processor.)

jem-davies (commented):
I see that you have altered it to draft after I started to take a look... 

But as well as the comments - what about thinking about the shutdown procedure - i.e. reading from a big file and Bento is shutdown - will that happen gracefully - 

What about the atomicity of the operations too?

internal/impl/io/processor_file.go
henrikschristensen: @jem-davies  Do u know of any other way of detecting if os.Rename will work, other than just trial and error? Otherwise it should simply just try os.Rename, and if that fails revert to the current impl.

internal/impl/io/processor_file.go
henrikschristensen: @jem-davies  yes makes perfect sense to write the message content.

Would branching make sense here? should the write not just preserve the message as there would be no real response from a file write?

henrikschristensen:
@jem-davies should the read method use a scanner like the file input? I seems that scanner impl handles the prob with large files. Reading in chunks etc.

jem-davies:
> @jem-davies should the read method use a scanner like the file input? I seems that scanner impl handles the prob with large files. Reading in chunks etc.

Yeah that's what I was thinking 👍

internal/impl/io/processor_file.go
jem-davies: >  should the write not just preserve the message as there would be no real response from a file write?

Ahh - yeah of course - no need to mention the branch processor stuff then

internal/impl/io/processor_file.go
jem-davies: Perhaps it would be good to add a new `operation`  option "rename" and document that it might be more useful to use the rename in place of move because it avoids copying the bytes - and have the user decide which is appropriate?

internal/impl/io/processor_file.go
henrikschristensen: Implemented your change

internal/impl/io/processor_file.go
henrikschristensen: Implemented addMetadata

internal/impl/io/processor_file.go
henrikschristensen: I removed the content field and changed it to write the current message content, as suggested.

internal/impl/io/processor_file.go
henrikschristensen: Solved

henrikschristensen:
For gracefull shutdown during large file reads i changed the read method to use a scanner. 
For atomicity of write and move i implemented steps like using temporary intermediate files to make sure we dont alter data if we fail during the operation.

internal/impl/io/processor_file.go
henrikschristensen: I introduced the new operation as suggested, and updated the docs

internal/impl/io/processor_file.go
```diff
@@ -0,0 +1,445 @@
...
+- file_mod_time_unix: File modification time as Unix timestamp
+- file_mod_time: File modification time in RFC3339 format
+- file_name: The name of the file
+- file_is_dir: Whether the file is a directory (true/false)
+- file_mode: File permissions and mode
+`+"```"+``).
+		Fields(
+			service.NewStringEnumField(fileProcessorFieldOperation, "read", "write", "delete", "move", "rename", "stat").
+				Description("The file operation to perform.").
+				Examples("read", "write", "delete", "move", "rename", "stat"),
```
jem-davies:
```suggestion
				Description("The file operation to perform."),
```

nit: remove these examples - the options are already displayed on the docs page.

internal/impl/io/processor_file.go
```diff
@@ -0,0 +1,445 @@
...
+			service.NewInterpolatedStringField(fileProcessorFieldPath).
+				Description("The source file path. Supports interpolation for dynamic paths.").
```
jem-davies:
```suggestion
				Description("The source file path.").
```

It is already mentioned automatically for NewInterpolatedStringFields.

internal/impl/io/processor_file.go
```diff
@@ -0,0 +1,445 @@
...
+				Examples(
+					"/tmp/data.txt",
+					"/tmp/${! json(\"document.id\") }.txt",
+				).LintRule(`if this == "" { [ "'path' must be set to a non-empty string" ] }`),
+			service.NewInterpolatedStringField(fileProcessorFieldDest).
+				Description("The destination path for 'move' and 'rename' operations. Supports interpolation.").
```
jem-davies:
```suggestion
				Description("The destination path for 'move' and 'rename' operations.").
```

internal/impl/io/processor_file.go
```diff
@@ -0,0 +1,445 @@
...
+	"github.com/warpstreamlabs/bento/public/service"
+)
+
+const (
+	fileProcessorFieldOperation = "operation"
+	fileProcessorFieldPath      = "path"
+	fileProcessorFieldDest      = "dest"
+	fileProcessorFieldScanner   = "scanner"
+)
+
```
jem-davies, line 33: nit: Add const for the operations options - and then reference them later

internal/impl/io/processor_file.go
```diff
@@ -0,0 +1,445 @@
...
+	"time"
+
+	"github.com/warpstreamlabs/bento/internal/component"
+	"github.com/warpstreamlabs/bento/public/service"
+)
+
+const (
+	fileProcessorFieldOperation = "operation"
+	fileProcessorFieldPath      = "path"
+	fileProcessorFieldDest      = "dest"
```
jem-davies:
```suggestion
	fileProcessorFieldPath      = "source_path"
	fileProcessorFieldDest      = "destination_path"
```

nit

internal/impl/io/processor_file.go
```diff
@@ -0,0 +1,445 @@
...
+func fileProcessorSpec() *service.ConfigSpec {
+	return service.NewConfigSpec().
+		Categories("Local").
+		Summary(`Performs operations (read, write, delete, move, rename, stat) on files.`).
+		Description(`
+This processor allows you to perform various file operations based on message content. The operation is specified by the 'operation' field, and paths can be dynamically generated using interpolation functions.
+
+### Operations
+
+- **read**: Read file content into the message
```
jem-davies: I think there should be clarification of which is being used `source_path` / `destination_path` - for instance the `write` operation because that will write to the 'path' - perhaps people would expect it to go to `dest`? 

I think it would make sense to say "Read file at `source_path` into the message" / "Write message content to `destination_path`.

internal/impl/io/processor_file.go
```diff
@@ -0,0 +1,445 @@
...
+- **write**: Write message content to a file
+- **delete**: Delete a file
+- **move**: Move a file by copying bytes (works across filesystems)
+- **rename**: Rename/move a file using os.Rename (faster, same filesystem only)
+- **stat**: Get file information without reading content
+
+### Important Differences: move vs rename
+
+- **rename**: Uses os.Rename() which is atomic and fast, but only works within the same filesystem. Fails if source and destination are on different filesystems or partitions.
+- **move**: Uses byte-level copying which works across filesystems but is slower and uses more resources. More reliable for cross-filesystem operations.
```
jem-davies:
```suggestion
- **move**: Move a file by copying bytes
- **rename**: Rename a file using [os.Rename](https://pkg.go.dev/os#Rename)
- **stat**: Get file information without reading content

### move vs rename

You may want to use the `rename` operation instead of using `move` and therefore avoid copying bytes. However you should be aware that there are OS-specific restrictions with this approach and it will not work across different filesystems. 
```

The current description seems to contradict [os.Rename's func description](https://pkg.go.dev/os#Rename) particularly regarding the atomicity.

internal/impl/io/processor_file.go
```diff
@@ -0,0 +1,445 @@
...
+	file, err := p.nm.FS().Open(path)
+	if err != nil {
+		return nil, fmt.Errorf("failed to open file '%s': %w", path, err)
+	}
+
+	fileInfo, err := file.Stat()
+	if err != nil {
+		file.Close()
+		return nil, fmt.Errorf("failed to get file info for '%s': %w", path, err)
+	}
```
jem-davies, line 195:
```suggestion
	file, err := p.nm.FS().Open(path)
	if err != nil {
		return nil, fmt.Errorf("failed to open file '%s': %w", path, err)
	}
	defer file.Close()

	fileInfo, err := file.Stat()
	if err != nil {
		return nil, fmt.Errorf("failed to get file info for '%s': %w", path, err)
	}
```

I think that we should just defer the closing of the file and not call it via the ackFunc.

internal/impl/io/processor_file.go
```diff
@@ -0,0 +1,445 @@
...
+
+	details := service.NewScannerSourceDetails()
+	details.SetName(path)
+
+	scanner, err := p.scanner.Create(file, func(ctx context.Context, err error) error {
+		file.Close()
```
jem-davies:
```suggestion
```

internal/impl/io/processor_file.go
```diff
@@ -0,0 +1,445 @@
...
+		return nil
+	}, details)
+	if err != nil {
+		file.Close()
```
jem-davies:
```suggestion
```

internal/impl/io/processor_file.go
```diff
@@ -0,0 +1,445 @@
...
+		return nil, fmt.Errorf("failed to create scanner for file '%s': %w", path, err)
+	}
+	defer scanner.Close(ctx)
+
+	var allMessages service.MessageBatch
+
+	// Process all batches from scanner until EOF
+	for {
+		parts, ackFn, err := scanner.NextBatch(ctx)
```
jem-davies:
```suggestion
		parts, _, err := scanner.NextBatch(ctx)
```

internal/impl/io/processor_file.go
```diff
@@ -0,0 +1,445 @@
...
+			}
+			newMsg.SetBytes(partBytes)
+			addFileMetadata(newMsg, path, fileInfo)
+
+			allMessages = append(allMessages, newMsg)
+		}
+
+		if err := ackFn(ctx, nil); err != nil {
+			p.log.Warnf("Failed to acknowledge scanner batch: %v", err)
+		}
```
jem-davies:
```suggestion
```

Replace with the defer file.Close()

internal/impl/io/processor_file.go
```diff
@@ -0,0 +1,445 @@
...
+		newMsg := msg.Copy()
+		addFileMetadata(newMsg, path, fileInfo)
+		return service.MessageBatch{newMsg}, nil
+	}
+
+	return allMessages, nil
+}
+
+func (p *fileProcessor) processWrite(_ context.Context, msg *service.Message) (service.MessageBatch, error) {
+	path, err := p.conf.Path.TryString(msg)
```
jem-davies, line 249: Just noting again regarding `path` / `dest` or `source_path` / `destination_path` - here we are writing to `path` and not `dest` - which is fine. But worth documenting this in the operations section of the docs imo.

internal/impl/io/processor_file.go
```diff
@@ -0,0 +1,445 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	if err := p.nm.FS().MkdirAll(filepath.Dir(path), fs.FileMode(0o777)); err != nil {
+		return nil, fmt.Errorf("failed to create directory for '%s': %w", path, err)
+	}
+
+	// Use atomic write pattern: write to temp file, then rename
+	tempFile := path + ".tmp"
```
jem-davies: Would add some sort of random prefix to ".tmp" - so to make sure we don't get a collision.

internal/impl/io/processor_file.go
```diff
@@ -0,0 +1,445 @@
...
+	if err := os.Rename(srcPath, destPath); err != nil {
+		return nil, fmt.Errorf("failed to rename file from '%s' to '%s': %w", srcPath, destPath, err)
+	}
+
+	return service.MessageBatch{msg}, nil
+}
+
+// atomicCopyAndDelete performs an atomic copy from src to dest and then deletes src
+// This ensures that either the operation completes fully or leaves the source intact
+func (p *fileProcessor) atomicCopyAndDelete(_ context.Context, srcPath, destPath string, msg *service.Message) (service.MessageBatch, error) {
```
jem-davies: We have context in the func signature but we aren't going to use it, and it's an unexported func so unclear why.

internal/impl/io/processor_file.go
```diff
@@ -0,0 +1,445 @@
...
+	}
+
+	tempFile := destPath + ".tmp"
+	srcFile, err := p.nm.FS().Open(srcPath)
+	if err != nil {
+		return nil, fmt.Errorf("failed to open source file '%s': %w", srcPath, err)
+	}
+	defer srcFile.Close()
+
+	content, err := io.ReadAll(srcFile)
```
jem-davies: This is potentially going to lead to OOM issues - thinking that this could be re-worked using a streaming approach to writing to the tempFile - and then we could also observe context cancellation.

internal/impl/io/processor_file.go
henrikschristensen: will all scanners move on to NextBatch if we fail to ack the one we are processing?

internal/impl/io/processor_file.go
henrikschristensen: deleted

internal/impl/io/processor_file.go
henrikschristensen: deleted

internal/impl/io/processor_file.go
henrikschristensen: deleted

internal/impl/io/processor_file.go
henrikschristensen: added

internal/impl/io/processor_file.go
henrikschristensen: so I settled for path/destination_path because i think source_path sounds wrong for operations delete and stat. Correct me if I am wrong.

internal/impl/io/processor_file.go
henrikschristensen: added doku for 
path. used for reads, writes, deletes and stat
destination_path. used for move and rename

internal/impl/io/processor_file.go
henrikschristensen: change into suggested

internal/impl/io/processor_file.go
henrikschristensen: changed into suggested

internal/impl/io/processor_file.go
henrikschristensen: changed into suggested

internal/impl/io/processor_file.go
henrikschristensen: changed into suggested

internal/impl/io/processor_file.go
henrikschristensen: changed into suggested

internal/impl/io/processor_file.go
henrikschristensen: changed into suggested

internal/impl/io/processor_file.go
henrikschristensen: went for a suffix instead. 8 random bytes added as filename.tmp_xxxxxxxxxxxxxxxx

internal/impl/io/processor_file.go
henrikschristensen: this was just to have the same function signature for all "process" functions so not important. Remove where not used.

internal/impl/io/processor_file.go
henrikschristensen: added to docs

internal/impl/io/processor_file.go
henrikschristensen: changed it to a streaming approach using io.copy

jem-davies (approved):
Thanks a lot! 🙏

PR #698, opened 2026-02-17
add(bloblang): is_unicode bytes method

gregfurman:
## Motivation

Certain components (i.e the `aws_sqs` output) only support payloads that are valid unicode.

This method will allow us to check this and handle it within a processor.

## Changes

- Adds a bytes method `is_unicode` that can be used to check if a byte array is valid unicode via bloblang.

### DIFF
--- a/internal/impl/pure/bloblang_string.go
+++ b/internal/impl/pure/bloblang_string.go
@@ -1,8 +1,11 @@
 package pure
 
 import (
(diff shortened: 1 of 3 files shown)


jem-davies:
Perhaps we could just handle the SQS stuff with: https://github.com/warpstreamlabs/bento/pull/717?

PR #700, opened 2026-02-18
just clone transport once

lublak:


lublak:
This pull request is also makes it easier more transports like spnego and digest Auth

lublak:
If you look closely, the first thing that is checked is whether the TLS Conf is set. Here in this block, it is then checked whether the current transport is also an http transport, and if so, it is cloned.
Otherwise, a new transport is created.


Afterwards, it is checked whether a proxy URL is configured.
There is already a deviation here:
Here in this block, it is then checked whether the current transport is also an HTTP transport, otherwise an error occurs here.
But! If a TLS Conf has been defined beforehand, this is then automatically an HTTP transport. This means that this block would only run into an error if TLS Conf is not set.

I have also normalized this here in the pull request to the status that a new transport is created for the respective case and does not result in an error.

If desired, we can also adjust the last else block in the pull request as follows:

```go
} else if tlsClientConfig == nil {
	return nil, fmt.Errorf("unable to apply proxy_url to transport, unexpected type %T", h.client.Transport)
} else {
	return nil, fmt.Errorf("unable to apply tls to transport, unexpected type %T", h.client.Transport)
}
```

or

```go
} else if tlsClientConfig == nil {
	return nil, fmt.Errorf("unable to apply proxy_url to transport, unexpected type %T", h.client.Transport)
} else {
	return nil, fmt.Errorf("unable to apply proxy_url or tls to transport, unexpected type %T", h.client.Transport)
}
```

internal/httpclient/client.go
```diff
@@ -75,34 +76,45 @@ func NewClientFromOldConfig(conf OldConfig, mgr *service.Resources, opts ...Requ
...
+				if proxy != nil {
+					cloned.Proxy = proxy
+				}
+
+				h.client.Transport = cloned
+			} else {
+				h.client.Transport = &http.Transport{
+					TLSClientConfig: tlsClientConfig,
+					Proxy:           proxy,
+				}
```
jem-davies: Here if the `h.client.Transport` isn't a *http.Transport (which afaik shouldn't be possible) rather than throw an error, we replace it. I suppose it's subtle but just wondering if that was intentional?

jem-davies: Oh I see you have mentioned this in the PR description - in that case I feel we should change it to what you suggested there 👍

jem-davies: Other than that it seems OK

lublak: @jem-davies thanks for the Review. I will commit a change to return an error instead silently replace the Transport.

PR #709, opened 2026-02-22
add(processor): Create go template processor

### LINKED ISSUE #693: Add template processor
```yaml
pipeline:
  processors:
    - template: |
        https://pkg.go.dev/text/template
```

root is now a string based on the template language from go

just like the other issues i created: i will do it if i have some time

lublak:
fixes #693


gregfurman:
This is super neat! Looking forward to giving this a review 👀

internal/impl/pure/processor_template.go
```diff
@@ -0,0 +1,125 @@
...
+			return nil, err
+		}
+
+		functions = make(template.FuncMap, len(functionsMap)+1)
+		for name, fn := range functionsMap {
+			blob, err := bloblang.Parse(fn)
+			if err != nil {
+				return nil, err
+			}
+			if name == "meta" {
```
gregfurman: Should we not be doing the same for all functions/operators to ensure we don't override defaults? https://pkg.go.dev/text/template#hdr-Functions

internal/impl/pure/processor_template.go
```diff
@@ -0,0 +1,125 @@
...
+
+const (
+	tpFieldText      = "text"
+	tpFieldFunctions = "functions"
+)
+
+func TemplateProcessorSpec() *service.ConfigSpec {
+	return service.NewConfigSpec().
+		Categories("Mapping").
+		Beta().
```
gregfurman: IMO let's comment this out for now and allow this component to be marked as `experimental` -- will require a `make docs` run 🙂 
```suggestion
		// Beta().
```

gregfurman (changes requested):
Apologies for the delay in review here. Once the conflicts are resolved, and my small comments addresses, this LGTM!

internal/impl/pure/processor_template.go
lublak: hmmm good point. I'm not sure if it makes sense to block it completely, since it's also possible to redefine the functions in Go. The reason I specifically check these functions is simply because I defined them myself and would override them if a user were to configure them. This check is there so the user doesn't wonder why their overridden function isn't taking effect. That's the real reason for it.

internal/impl/pure/processor_template.go
lublak: ok! 👍

lublak:
@gregfurman no problem :) all fine!
I addressed the comments, but also add another new function (`from_file` and `sub_templates`).

gregfurman (approved):
LGTM!

gregfurman:
Could we perhaps rename the PR title to `add(processor): Create go template processor` or something similar? Once done, am happy to merge :shipit:

lublak:
done!

PR #710, opened 2026-02-23
add cache exists (with all implementation in each cache component)

### LINKED ISSUE #704: Add cache append + exsits operator
Also on my todo list:

Just appends to the string:

```yaml
pipeline:
  cache:
    resource: cache_resource
    operator: append
    key: ""
    value: ""
```

```yaml
pipeline:
  cache:
    resource: cache_resource
    operator: exists
    key: ""


root = true/false or root = {"cache_resource": true/false} or root = {"exists": true/false}
```

lublak:
fixes #704


internal/component/cache/interface.go
```diff
@@ -17,6 +17,10 @@ type V1 interface {
 	// error if the key does not exist or if the command fails.
 	Get(ctx context.Context, key string) ([]byte, error)
 
+	// Exists attempts to locate a cached value by its key, returns
+	// false if the key does not exist, returns an if the command fails.
+	Exists(ctx context.Context, key string) (bool, error)
```
jem-davies: Adding a new method to the interface definition will mean that any cache implementations that a user of Bento could be using via [a plugin](https://pkg.go.dev/github.com/warpstreamlabs/bento/v4/public/service#example-package-CachePlugin) will break. 

That would be an issue because we follow [semantic versioning ](https://semver.org/) therefore we would only be able to merge this after bumping the major version number. 

I thought that the issue was pertaining to adding an "exists" operator to the cache processor?

jem-davies (commented):
Modifying the interface for a cache will break existing cache implementations 🤔

lublak:
@jem-davies what i can do is: create an extra interface just for Exists.
If the cache implements Exists use the Exists function, if not: use Get function and check for `ErrNotExist`.
I implemented it this way specifically because many caches can only check for the existence of a value without actually reading the value itself.

jem-davies:
> @jem-davies what i can do is: create an extra interface just for Exists. If the cache implements Exists use the Exists function, if not: use Get function and check for `ErrNotExist`. I implemented it this way specifically because many caches can only check for the existence of a value without actually reading the value itself.

This sounds better - yeah. I guess I would just have added a new operator `exists` to the cache processor but I think that your approach might be better - as you say many cache's do have native support for `exists`.

lublak:
@jem-davies are you fine with this implementation? Or should i change some stuff?

jem-davies:
> @jem-davies are you fine with this implementation? Or should i change some stuff?

I had a quick scan through before - will properly look again before the next release 👍

jem-davies:
> @jem-davies are you fine with this implementation? Or should i change some stuff?

Taking a good look right now.

public/service/cache.go
```diff
@@ -21,6 +21,9 @@ type Cache interface {
 	// Get a cache item.
 	Get(ctx context.Context, key string) ([]byte, error)
 
+	// Check if a cache item exists.
+	Exists(ctx context.Context, key string) (bool, error)
+
```
jem-davies:
```suggestion
```

We need to get rid of this don't we - save us altering the public API ?

internal/component/cache/cache_metrics.go
```diff
@@ -80,6 +80,22 @@ func (a *metricsCache) Get(ctx context.Context, key string) ([]byte, error) {
 	return b, err
 }
 
+func (a *metricsCache) Exists(ctx context.Context, key string) (bool, error) {
+	started := time.Now()
+	b, err := CacheKeyExists(a.c, ctx, key)
+	a.mGetLatency.Timing(int64(time.Since(started)))
+	if err != nil {
+		if errors.Is(err, component.ErrKeyNotFound) {
+			a.mGetNotFound.Incr(1)
```
jem-davies: How come we haven't added some new fields to `metricsCache` for the new exists operation? - Won't this be confusing?

internal/filepath/ifs/os.go
```diff
@@ -31,9 +32,27 @@ func ReadFile(f fs.FS, name string) ([]byte, error) {
...
+	defer i.Close()
 	return io.ReadAll(i)
 }
 
+// FileExists checks if a file exists.
+func FileExists(f fs.FS, name string) (bool, error) {
+	if ef, ok := f.(FS); ok {
+		return ef.Exists(name)
+	} else {
+		file, err := f.Open(name)
```
jem-davies, line 44: We could just call os.Stat() - to save actually opening the file?

```go
	_, err := os.Stat(filepath)
	if err != nil {
		if errors.Is(err, fs.ErrNotExist) {
			fileExists = false
		} else {
			return err
		}
	}
```

internal/filepath/ifs/os.go
```diff
@@ -31,9 +32,27 @@ func ReadFile(f fs.FS, name string) ([]byte, error) {
...
+		if err != nil {
+			if errors.Is(err, os.ErrNotExist) {
+				return false, nil
+			}
+			return false, err
+		}
+		file.Close()
```
jem-davies: We would want to defer this, but I think we can avoid opening the file anyway

internal/filepath/ifs/os.go
```diff
@@ -82,6 +101,17 @@ func (o *osPT) OpenFile(name string, flag int, perm fs.FileMode) (fs.File, error
 	return os.OpenFile(name, flag, perm)
 }
 
+func (o *osPT) Exists(name string) (bool, error) {
+	_, err := os.Stat(name)
+	if err != nil {
+		if os.IsNotExist(err) {
```
jem-davies: From the doc-string on os.IsNotExist: ` New code should use errors.Is(err, fs.ErrNotExist). `

internal/impl/pure/cache_multilevel.go
```diff
@@ -126,6 +126,27 @@ func (l *multilevelCache) Get(ctx context.Context, key string) ([]byte, error) {
 	return nil, service.ErrKeyNotFound
 }
 
+func (l *multilevelCache) Exists(ctx context.Context, key string) (bool, error) {
+	for i, name := range l.caches {
+		var data []byte
+		var err error
+		if cerr := l.mgr.AccessCache(ctx, name, func(c service.Cache) {
+			data, err = c.Get(ctx, key)
```
jem-davies, line 134: Could we use the new .Exists here? - Would have to check that .Exists is implemented?

internal/impl/pure/output_cache.go
```diff
@@ -56,6 +58,9 @@ In order to create a unique `+"`key`"+` value per item you should use function i
 					`${!metadata("kafka_key")}`,
 				).
 				Default(`${!count("items")}-${!timestamp_unix_nano()}`),
+			service.NewBoolField(coFieldAppend).
```
jem-davies: Are there some tests for this new functionality?

internal/impl/redis/cache.go
```diff
@@ -143,6 +143,42 @@ func (r *redisCache) Get(ctx context.Context, key string) ([]byte, error) {
...
+	}()
+
+	if r.prefix != "" {
+		key = r.prefix + key
+	}
+
+	for {
+		res, err := r.client.Exists(ctx, key).Result()
+		if err == nil {
+			if res == 1 {
```
jem-davies: Where are these numbers defined - could we reference a package variable instead of a literal?

internal/impl/sql/cache_sql.go
```diff
@@ -206,6 +210,17 @@ func (s *sqlCache) Get(ctx context.Context, key string) (value []byte, err error
 	return
 }
 
+func (s *sqlCache) Exists(ctx context.Context, key string) (exists bool, err error) {
+	err = s.existsBuilder.
+		Where(squirrel.Eq{s.keyColumn: key}).
+		RunWith(s.db).QueryRowContext(ctx).
+		Scan()
+	if err != nil && errors.Is(err, sql.ErrNoRows) {
```
jem-davies: What if the err is something other than `nil` & `sql.ErrNoRows` ?

jem-davies (changes requested):
I will need to check out the changes to internal/impl/pure/processor_cache.go & /output_cache.go more again.

public/service/cache.go
lublak: @jem-davies 

hmmm should i rewrite the tests? Its uses the service.Cache which now doesn't have Exists anymore its now implemented in airGapCache (like SetMulti)

grafik

its used here:

grafik

internal/component/cache/cache_metrics.go
lublak: i simply forgot to work on cache_metrics sry

internal/impl/pure/cache_multilevel.go
lublak: @jem-davies 
no, sadly not. I implemented it with Exists. But i found out: Its not possible.
We need to call l.setUpToLevelPassive(ctx, i, key, data) to sync all caches with the existing value. And we can only sync them if we fetch the value. I mean, yes we can remove the sync. I'm not sure if that's how it's meant to be?
Here are my changes from "Exists" to "Get":
https://github.com/warpstreamlabs/bento/pull/710/changes/3f596601a9b41e4496606f9d1744e71237cab1e2

internal/impl/redis/cache.go
lublak: @jem-davies 
hmm i don't know if there is a package variable, i can add a comment for this?
Its the numeric representation of the boolean:
```
true: 1
false: 0
```
I think i change it also to `return res == 1, nil`

internal/impl/sql/cache_sql.go
lublak: thanks for pointing out this mistake!

internal/filepath/ifs/os.go
lublak: fs.FS doesnt have a stat function, only a Open function. (It doesn't have any other functions either.)
os.FS exists function uses the stat function.
`func FileExists(f fs.FS, name string) (bool, error)`
checks if its an os.FS, than it calls "Exists"
if it's not it fall backs to fs.FS -> Open
grafik

internal/impl/pure/output_cache.go
```diff
@@ -56,6 +59,9 @@ In order to create a unique `+"`key`"+` value per item you should use function i
...
+				Description("When enabled, new messages are appended to existing cache values rather than overwriting them. If the key does not exist the message will be stored as a new value.").
+				Default(false),
```
jem-davies:
```suggestion
				Default(false).
				Version("1.18.0"),
```

internal/impl/pure/cache_ttlru_test.go
```diff
@@ -127,3 +140,77 @@ func BenchmarkTTLRUParallel(b *testing.B) {
...
+	}
+
+	exp := "1"
+	if act, err := c.Get(ctx, key); err != nil {
+		t.Errorf("unexpected error while c.Get(ctx, %q): %v", key, err)
+	} else if string(act) != exp {
+		t.Errorf("Wrong result c.Get(ctx, %q): %v != %v", key, string(act), exp)
+	}
+
+	//exists, err = c.Exists(ctx, key)
```
jem-davies: Commented out code here

lublak:
@jem-davies "failed test" comes from another plugin has nothing to do with the pull request.

internal/impl/pure/output_cache.go
```diff
@@ -56,6 +59,10 @@ In order to create a unique `+"`key`"+` value per item you should use function i
 					`${!metadata("kafka_key")}`,
 				).
 				Default(`${!count("items")}-${!timestamp_unix_nano()}`),
+			service.NewBoolField(coFieldAppend).
+				Description("When enabled, new messages are appended to existing cache values rather than overwriting them. If the key does not exist the message will be stored as a new value.").
+				Default(false).
+				Version("1.18.0"),
```
jem-davies:
```suggestion
				Version("1.19.0"),
```

internal/impl/pure/output_cache.go
```diff
@@ -162,9 +175,35 @@ func (c *CacheWriter) writeMulti(ctx context.Context, msg message.Batch) (err er
 		if terr != nil {
 			return fmt.Errorf("key interpolation error: %w", terr)
 		}
-		items[keyStr] = cache.TTLItem{
-			Value: p.AsBytes(),
-			TTL:   ttl,
+		if c.appendCache {
+			var current []byte
+			if cacheItem, ok := items[keyStr]; ok {
+				current = append(cacheItem.Value, p.AsBytes()...)
```
jem-davies: I think we should replace any calls to `append` with some `make` & `copy`. This is more defensive against the append behaviour, in that it could inadvertently modify the value before we call Set.

jem-davies (dismissed):
I think we should replace calls to `append` with `make` & `copy` in the append operators.

jem-davies:
@gregfurman - I have removed the 'append' operation from the processor & output.

internal/impl/pure/processor_cache_test.go
```diff
@@ -61,6 +61,52 @@ cache:
 	assert.Equal(t, "foo 2", actV.Value)
 }
 
+func TestCacheAppend(t *testing.T) {
+	mgr := mock.NewManager()
+	mgr.Caches["foocache"] = map[string]mock.CacheItem{}
+
+	conf, err := testutil.ProcessorFromYAML(`
+cache:
+  operator: append
```
gregfurman: @jem-davies what about here? Won't this fail?

public/service/cache.go
```diff
@@ -55,12 +55,21 @@ type batchedCache interface {
...
+	Exists(ctx context.Context, key string) (bool, error)
+}
+
 //------------------------------------------------------------------------------
 
 // Implements types.Cache.
 type airGapCache struct {
 	c  Cache
 	cm batchedCache
+	ce existsCache
```
gregfurman, line 72: Is this ever set? If so, can't seem to find it.

Can see `batchedCache` is set here though:

https://github.com/warpstreamlabs/bento/blob/eba3ed4192ea65bbe7bfc6b1d9fb124bea1483cb/public/service/cache.go#L66-L70

public/service/cache.go
```diff
@@ -136,6 +159,10 @@ func (r *reverseAirGapCache) Get(ctx context.Context, key string) ([]byte, error
 	return b, err
 }
 
+func (r *reverseAirGapCache) Exists(ctx context.Context, key string) (bool, error) {
+	return r.c.Exists(ctx, key)
+}
```
jem-davies, line 165:
```suggestion
```

lublak:
@jem-davies i was on a trip. Right now, I don't know yet what changes you've made.
Why the feature "append" was removed?

jem-davies:
> Why the feature "append" was removed?

It isn't safe for concurrent access because the two operations of get & set form two separate contexts (or those operations together are not atomic)

jem-davies:
Tested implementing a cache plugin with Exists() & without Exists()

PR #717, opened 2026-02-26
add remove_invalid_codepoints  to  aws_sqs output

jem-davies:
You can't send any unicode codepoints that don't belong in the set: 

#x9 | #xA | #xD | #x20 to #xD7FF | #xE000 to #xFFFD | #x10000 to #x10FFFF

So maybe we should add a `remove_invalid_codepoints` option to the aws_sqs output to remove them before sending. 

--- 

### TODO

- [ ] refactor test to not use unsafe package

### DIFF
--- a/internal/impl/aws/output_sqs.go
+++ b/internal/impl/aws/output_sqs.go
@@ -36,7 +36,9 @@ const (
 	sqsMaxRecordsCount = 10
 )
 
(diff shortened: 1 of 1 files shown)


jem-davies:
https://github.com/warpstreamlabs/bento/issues/752

PR #722, opened 2026-03-06
add(gcp): BigTable output component

gregfurman:
## Motivation

This PR adds a GCP BigTable sink. Currently, it only supports `Set` operations.

## Changes

- Adds new BigTable output component `gcp_bigtable` that uses string interpolation for dynamically getting the table ID, row key, column, and column family values from a given message.

### DIFF
--- a/internal/impl/gcp/input_pubsub.go
+++ b/internal/impl/gcp/input_pubsub.go
@@ -6,7 +6,7 @@ import (
 	"strings"
 	"sync"
 
-	"cloud.google.com/go/pubsub"
+	"cloud.google.com/go/pubsub" //nolint:staticcheck
 	"google.golang.org/api/option"
 
 	"github.com/warpstreamlabs/bento/public/service"
--- a/internal/impl/gcp/input_pubsub_test.go
+++ b/internal/impl/gcp/input_pubsub_test.go
@@ -5,7 +5,7 @@ import (
 	"testing"
 	"time"
 
-	"cloud.google.com/go/pubsub"
+	"cloud.google.com/go/pubsub" //nolint:staticcheck
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
--- a/internal/impl/gcp/integration_pubsub_test.go
+++ b/internal/impl/gcp/integration_pubsub_test.go
@@ -7,7 +7,7 @@ import (
 	"testing"
 	"time"
 
-	"cloud.google.com/go/pubsub"
+	"cloud.google.com/go/pubsub" //nolint:staticcheck
 	"github.com/ory/dockertest/v3"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
--- a/internal/impl/gcp/output_pubsub.go
+++ b/internal/impl/gcp/output_pubsub.go
@@ -5,7 +5,7 @@ import (
 	"fmt"
 	"sync"
 
-	"cloud.google.com/go/pubsub"
+	"cloud.google.com/go/pubsub" //nolint:staticcheck
 	"github.com/sourcegraph/conc/pool"
 	"google.golang.org/api/option"
 
--- a/internal/impl/gcp/output_pubsub_test.go
+++ b/internal/impl/gcp/output_pubsub_test.go
@@ -5,7 +5,7 @@ import (
 	"errors"
 	"testing"
 
-	"cloud.google.com/go/pubsub"
+	"cloud.google.com/go/pubsub" //nolint:staticcheck
 	"github.com/stretchr/testify/mock"
 	"github.com/stretchr/testify/require"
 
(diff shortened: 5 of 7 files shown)


internal/impl/gcp/output_bigtable.go
```diff
@@ -0,0 +1,247 @@
...
+
+type gcpBigTableOutputConfig struct {
+	ProjectID  string
+	InstanceID string
+
+	TableID *service.InterpolatedString
+
+	RowKey       *service.InterpolatedString
+	ColumnName   *service.InterpolatedString
+	ColumnFamily *service.InterpolatedString
```
gregfurman, line 101: @jem-davies Thoughts on these being interpolated strings? Otherwise, we can make them bloblang mappings and do a `.(string)` type check before using them.

internal/impl/gcp/output_bigtable.go
```diff
@@ -0,0 +1,247 @@
...
+	if g.client == nil {
+		client, err := bigtable.NewClient(ctx, g.conf.ProjectID, g.conf.InstanceID)
+		if err != nil {
+			return fmt.Errorf("could not create bigtable client: %w", err)
+		}
+
+		g.client = client
+	}
+
+	return g.client.PingAndWarm(ctx)
```
jem-davies, line 163: So if g.client.PingAndWarm fails - then we will retry but without creating a new client - is that right? Would it be better if we discarded that client before invocation of Connect again?

internal/impl/gcp/output_bigtable.go
```diff
@@ -0,0 +1,247 @@
...
+
+The `+"`table`"+`, `+"`row_key`"+`, `+"`column`"+`, and `+"`family`"+` fields support [interpolation functions](/docs/configuration/interpolation#bloblang-queries), allowing values to be resolved dynamically per message. This enables routing messages to different tables, rows, column families, or column qualifiers based on message content or metadata.
+
+### Credentials
+
+By default Bento will use a shared credentials file when connecting to GCP services. You can find out more [in this document](/docs/guides/cloud/gcp).
+
+### Batching
+
+Messages in a batch are written using BigTable's [ApplyBulk](https://docs.cloud.google.com/go/docs/reference/cloud.google.com/go/bigtable/latest#cloud_google_com_go_bigtable_Table_ApplyBulk) API call.`).
```
jem-davies:
```suggestion
This output benefits from sending messages as a batch for improved performance. Batches can be formed at both the input and output level. You can find out more [in this doc](/docs/configuration/batching).`).
```

I don't think the current description is wrong: just that a user might infer that "unbatched messages" don't use the ApplyBulk api call. 

I think that the suggestion is better for users to realise that a batching policy will help - and doesn't expose unnecessary implementation details to them.

internal/impl/gcp/output_bigtable.go
```diff
@@ -0,0 +1,247 @@
...
+
+func gcpBigTableOutputSpec() *service.ConfigSpec {
+	return service.NewConfigSpec().
+		Beta().
+		Categories("Services", "GCP").
+		Summary("Writes messages to a GCP BigTable instance.").
+		Description(`
+Each message is written as a [SetCell](https://cloud.google.com/bigtable/docs/reference/data/rpc/google.bigtable.v2#google.bigtable.v2.Mutation.SetCell) mutation into the specified column family and column qualifier.
+
+The `+"`table`"+`, `+"`row_key`"+`, `+"`column`"+`, and `+"`family`"+` fields support [interpolation functions](/docs/configuration/interpolation#bloblang-queries), allowing values to be resolved dynamically per message. This enables routing messages to different tables, rows, column families, or column qualifiers based on message content or metadata.
```
jem-davies:
```suggestion
The `+"`table`"+`, `+"`row_key`"+`, `+"`column`"+`, and `+"`family`"+` fields support [interpolation functions](/docs/configuration/interpolation#bloblang-queries), allowing values to be resolved dynamically. This enables routing messages to different tables, rows, column families, or column qualifiers based on message content or metadata.

:::caution Interpolation of Message Batches
The first message in the batch will resolve the bloblang query for the field `+"`table`"+` and that value will be used for all messages in the batch.
:::
```

I think that removing "per message" from the first sentence is necessary because table isn't per message - it's per batch. 

Also then adding a more visible caution regarding that behaviour with the 'table' field.

internal/impl/gcp/output_bigtable.go
gregfurman: Ughh good catch 🫠

gregfurman: > So if g.client.PingAndWarm fails - then we will retry but without creating a new client - is that right?

Upon second glance, I think this approach is fine. Afaik underlying client is backed by a grpc connection pool which should handle retries etc.

Like, the client being re-created shouldn't make a difference to whether `PingAndWarm` fail (unless the client is closed but this would need to be explicitly called via `.Close(ctx)`)

> Would it be better if we discarded that client before invocation of Connect again?

As mentioned, don't think this makes much a difference because of the fancy connection pooling under-the-hood

PR #724, opened 2026-03-06
docs: Add DuckDB examples and in-memory DSN to driver table

iamramtin:
Add DuckDB examples to `sql_insert` (output), `sql_raw` (input/processor) docs and update the DuckDB DSN table row to include `:memory:` for ephemeral in-process storage.

### DIFF
--- a/internal/impl/sql/processor_sql_raw.go
+++ b/internal/impl/sql/processor_sql_raw.go
@@ -104,6 +104,7 @@ pipeline:
 			"In-Memory Computation (DuckDB)",
 			"Run SQL expressions against message fields using a DuckDB `:memory:` database(no external file required).",
 			`
+# BENTO LINT DISABLE
 pipeline:
   processors:
     - sql_raw:
--- a/website/docs/components/processors/sql_raw.md
+++ b/website/docs/components/processors/sql_raw.md
@@ -165,6 +165,7 @@ pipeline:
 Run SQL expressions against message fields using a DuckDB `:memory:` database(no external file required).
 
 ```yaml
+# BENTO LINT DISABLE
 pipeline:
   processors:
     - sql_raw:


internal/impl/sql/input_sql_raw.go
```diff
@@ -52,6 +52,20 @@ input:
       root = [
         now().ts_unix() - 3600
       ]
+`,
+		).
+		Example("Aggregate Query (DuckDB)",
+			"Read aggregated results from a DuckDB file as a stream of messages.",
+			`
+input:
```
gregfurman, line 61: @iamramtin Since the CI is failing for these due to the linter not running with the `x_bento_extra` tag, we have two options:

1. Prepend a `# BENTO LINT DISABLE` to each of these to tell the `bento lint` command to ignore this. FYI this will be present in the docs but IMO that's not a huge deal.
2. Compile and run the lint with the `x_bento_extra` tag (am unsure if there are any side effects).

Thoughts?

iamramtin: I've gone with option 2, since it's a simple one line change and doesn't leak the comment in the docs. Maybe worth rethinking this solution down the line if the CI starts to slow down with more and more x_bento_extra docs being validated

iamramtin: Looks like that introduced a different error, looks like ZeroMQ needs `libzmq`, so it would need to be installed a system package in the CI, which feels like a code smell. I think the simplest fix is to remove `conn_fields_no_cgo.go`, what do you think?

iamramtin: @gregfurman I've reverted that change, and introduced a separate PR to address this instead: https://github.com/warpstreamlabs/bento/pull/726

LMK if you're happy with the approach, if so can merge that in then this one after rebasing

gregfurman:
@iamramtin Can you rebase again?

internal/impl/sql/processor_sql_raw.go
```diff
@@ -75,6 +75,43 @@ pipeline:
...
+                  ('Huey Duck','Junior Woodchuck'),('Launchpad McQuack','Pilot');
+              conn_max_open: 1
+        result_map: "root.occupation = this.index(0).occupation"
+`,
+		).
+		Example(
+			"In-Memory Computation (DuckDB)",
+			"Run SQL expressions against message fields using a DuckDB `:memory:` database(no external file required).",
+			`
+pipeline:
```
gregfurman, line 108: Can you add the sentinel here as well and regenerate?

iamramtin: Done

iamramtin:
@gregfurman CI finally passes! 🥳 Thanks for addressing the lint and test issues

PR #726, opened 2026-03-07
feat: Add lint rule for build constraint warnings

iamramtin:
Add `LintBuildConstraintRule` on `ConfigField` so docs **warn instead of error** when the required build flags aren't present.

This addresses [PR](https://github.com/warpstreamlabs/bento/pull/724) where config docs using `driver: duckdb` fail because they require the `x_bento_extra` build tag. The rule prevents those failures without breaking doc validation in CI.

### DIFF
--- a/cmd/tools/bento_docs_gen/schema_test.go
+++ b/cmd/tools/bento_docs_gen/schema_test.go
@@ -28,6 +28,9 @@ func TestComponentExamples(t *testing.T) {
 		lConf.RejectDeprecated = !deprecated
 		lints := confSpec.LintYAML(docs.NewLintContext(lConf), node)
 		for _, lint := range lints {
+			if lint.Level == docs.LintWarning && lint.Type == docs.LintBuildConstraint {
+				continue
+			}
 			t.Errorf("%v %v:%v:%v", lint, componentType, typeName, title)
 		}
 	}


gregfurman:
Hey! Am giving this a look. Will post a review a bit later with some ideas on this since I think including more lint rules that can be defined structurally instead of via bloblang would be a good idea.

gregfurman:
OK have given this another look and I think this functionality needs some more consideration. I'd be in favour of formalising adding more complicated config rules but think this new functionality is a bit _too_ hyperspecific.

For now, I'd be in favour of adding some functionality in the `website` that perhaps scrubs the `# BENTO LINT DISABLE` comment from the example config.

Also, if we set a YAML block using `yaml` instead of `yml`, then the linter will ignore it. So perhaps an idea here is to automate this when that comment/sentinel is present.

iamramtin:
> OK have given this another look and I think this functionality needs some more consideration. I'd be in favour of formalising adding more complicated config rules but think this new functionality is a bit _too_ hyperspecific.
> 
> For now, I'd be in favour of adding some functionality in the `website` that perhaps scrubs the `# BENTO LINT DISABLE` comment from the example config.
> 
> Also, if we set a YAML block using `yaml` instead of `yml`, then the linter will ignore it. So perhaps an idea here is to automate this when that comment/sentinel is present.

Thanks for the review. The use of `# BENTO LINT DISABLE` comment from the example config was not supported, so I will close this PR in favour of adding that support instead. Please see https://github.com/warpstreamlabs/bento/pull/729

iamramtin:
@gregfurman @jem-davies closing this in favour of https://github.com/warpstreamlabs/bento/pull/729

PR #729, opened 2026-03-09
feat: Support disabling linter on component examples

iamramtin:
Support use of `# BENTO LINT DISABLE` sentinel in component examples by emitting `yml` instead of `yaml` in the rendered docs to skip linting and then stripping the sentinel comment.

This addresses [PR](https://github.com/warpstreamlabs/bento/pull/724) where config docs using `driver: duckdb` fail because they require the `x_bento_extra` build tag. Supporting the sentinel in the example docs prevents these failures without breaking doc validation in CI.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -26,6 +26,9 @@ jobs:
         go-version: 1.25.x
         check-latest: true
 
+    - name: Ext Deps (x_bento_extra)
+      run: sudo apt-get update && sudo apt-get install -y libzmq3-dev
+
     - name: Deps
       run: make deps && git diff-index --quiet HEAD || { >&2 echo "Stale go.{mod,sum} detected. This can be fixed with 'make deps'."; exit 1; }
 


iamramtin:
@gregfurman LMK if you're happy with this, then I can close https://github.com/warpstreamlabs/bento/pull/726 and update https://github.com/warpstreamlabs/bento/pull/724 based on this.

gregfurman:
I'm kinda averse to changing the `component_markdown.go` unless totally necessary. Can we not just disable this for now in the `lintMDSnippets`?

https://github.com/warpstreamlabs/bento/blob/718e922ba68a10ccdb812c9fcef231f9da65b79e/internal/cli/lint.go#L52-L62

We can just add check that ignores linting if the sentinel is present and allow this to be included in the generated markdown i.e

```golang
	if bytes.HasPrefix(rawBytes, []byte("\n# BENTO LINT DISABLE")) {
		return pathLints
	}
```

@jem-davies Wdyt? Should we be adding an extra disabled field or kick the can down the road and think how best to handle the sentinel later

jem-davies:
> @jem-davies Wdyt? Should we be adding an extra disabled field or kick the can down the road and think how best to handle the sentinel later

I think it would be good if the lintMDSnippets behaved the same as other instances we Lint - i.e. lintMDSnippets should check for the existence of the `BENTO LINT DISABLE` prefix. 

But then it would be better if we stripped out the prefix from the actual docs...  just that replacing the ```yaml -> ```yml feels a bit of a hack. But then also realise that it feels somewhat insignificant to argue over so I would just roll with the PR as-is imo...

jem-davies:
Or you add `### BENTO LINT DISABLE` to examples that will fail linting -> update the lintMDSnippets -> docs are generated with the `### BENTO LINT DISABLE` inside them ... Then as another explicit step you remove the `### BENTO LINT DISABLE` ... Does that make sense/work? 

Either way - that idea could be added as a ticket and worked on afterwards imo ...

iamramtin:
@jem-davies @gregfurman Thanks for the review. This approach seems to be the simplest way with minimal changes to not lint the docs, and keeping the sentinel in seems a little clunky for user-facing docs, so I'd definitely think it best to strip it out and proceed this way if possible

iamramtin:
> Or you add `### BENTO LINT DISABLE` to examples that will fail linting -> update the lintMDSnippets -> docs are generated with the `### BENTO LINT DISABLE` inside them ... Then as another explicit step you remove the `### BENTO LINT DISABLE` ... Does that make sense/work?
> 
> Either way - that idea could be added as a ticket and worked on afterwards imo ...

Not sure I quite understand, would this be different to the current approach of disabling linting docs on `# BENTO LINT DISABLE` and then stripping them out?

jem-davies:
> Not sure I quite understand, would this be different to the current approach of disabling linting docs on `# BENTO LINT DISABLE` and then stripping them out?

So I mean: 

- update the func `lintMDSnippets` so it will skip over any snippets with the prefix `### BENTO LINT DISABLE`. 
- Add `### BENTO LINT DISABLE` to snippets that will fail because we lack build tags 
- Generate the website using `make docs` (at this point we still have `### BENTO LINT DISABLE` in the examples)
- Add another explicit step "Scrub bento lint disable" to the process that runs after calling `make docs` 

Avoiding the .yaml -> .yml trick

gregfurman:
@jem-davies @iamramtin I updated the approach to instead strip the rendering on the frontend MD snippets and skip configuration with the sentinel when linting said MD snippets.

iamramtin:
Looks good, thanks @gregfurman :shipit:

internal/cli/lint.go
```diff
@@ -59,6 +59,12 @@ func lintMDSnippets(path string, spec docs.FieldSpecs, lConf docs.LintConfig) (p
 		return
 	}
 
+	if bytes.HasPrefix(rawBytes, []byte("\n# BENTO LINT DISABLE")) {
+		// TODO(gregfurman): This magic comment approach is devastatingly hacky and should be re-considered
```
jem-davies, line 63: This is how it works in all other instances of linting, when starting a config via CLI , REST API etc.

PR #735, opened 2026-03-10
feat(output): Add datadog logs sink

gregfurman:
## Motivation

Adds an output component for writing logs to datadog.

### DIFF
--- a/internal/impl/datadog/output_datadog_logs.go
+++ b/internal/impl/datadog/output_datadog_logs.go
@@ -110,9 +110,8 @@ Set `+"`api_key`"+` explicitly or via the `+"`DD_API_KEY`"+` environment variabl
 }
 
 type datadogLogWriterConfig struct {
-	apiKey   string
-	site     string
-	endpoint string
+	apiKey string
+	site   string
 
 	contentEncoding datadogV2.ContentEncoding
 
@@ -312,7 +311,7 @@ func (d *datadogLogWriter) WriteBatch(ctx context.Context, batch service.Message
 		}
 		item := datadogV2.HTTPLogItem{
 			Message:              string(contents),
-			AdditionalProperties: make(map[string]interface{}, 2),
+			AdditionalProperties: make(map[string]any, 2),
 		}
 		if err := exec(i, ddsourceExec, item.SetDdsource); err != nil {
 			batchErrFailed(i, fmt.Errorf("message %d resolving ddsource: %w", i, err))


internal/impl/datadog/output_datadog_logs.go
```diff
@@ -0,0 +1,298 @@
...
+
+	contentEncoding datadogV2.ContentEncoding
+
+	ddsource *service.InterpolatedString
+	ddtags   *service.InterpolatedString
+	hostname *service.InterpolatedString
+	service  *service.InterpolatedString
+}
+
+func (c *datadogLogWriterConfig) GetContext(ctx context.Context) (context.Context, error) {
```
jem-davies: In the function signature we have an error - that we aren't using in the body of the function - so we should remove this?

internal/impl/datadog/output_datadog_logs.go
```diff
@@ -0,0 +1,298 @@
...
+		if err := exec(i, serviceExec, item.SetService); err != nil {
+			return fmt.Errorf("message %d resolving service: %w", i, err)
+		}
+		items[i] = item
+	}
+
+	submitCtx, err := d.writerConf.GetContext(ctx)
+	if err != nil {
+		return err
+	}
```
jem-davies: This GetContext can't actually return an error.

internal/impl/datadog/output_datadog_logs.go
```diff
@@ -0,0 +1,298 @@
...
+
+	opts := datadogV2.NewSubmitLogOptionalParameters().WithContentEncoding(d.writerConf.contentEncoding)
+
+	_, resp, err := d.client.SubmitLog(
+		submitCtx,
+		items,
+		*opts,
+	)
+	defer resp.Body.Close()
```
jem-davies:
```suggestion
    if resp != nil {
	    defer resp.Body.Close()
	}
```

internal/impl/datadog/output_datadog_logs.go
```diff
@@ -0,0 +1,298 @@
...
+	for i, msg := range batch {
+		contents, err := msg.AsBytes()
+		if err != nil {
+			return fmt.Errorf("message %d reading body: %w", i, err)
+		}
+		item := datadogV2.HTTPLogItem{
+			Message: string(contents),
+		}
+		if err := exec(i, ddsourceExec, item.SetDdsource); err != nil {
+			return fmt.Errorf("message %d resolving ddsource: %w", i, err)
```
jem-davies: Could we use a [BatchError](https://pkg.go.dev/github.com/warpstreamlabs/bento/public/service#BatchError) here?

internal/impl/datadog/output_datadog_logs.go
```diff
@@ -0,0 +1,298 @@
...
+
+:::warning
+Log events can be submitted with a timestamp that is up to 18 hours in the past.
+:::
+
+### Authentication
+
+Set `+"`api_key`"+` explicitly or via the `+"`DD_API_KEY`"+` environment variable.`).
+		Fields(
+			service.NewStringField(ddoFieldAPIKey).
```
jem-davies, line 73: I noticed in datadog.NewDefaultContext - it also looks for a [DD_APP_KEY](https://github.com/DataDog/datadog-api-client-go/blob/e199042bbf57d7b33f943965a87feb772487fa5d/api/datadog/configuration.go#L1081) - is that something we need to have a config field for - or mention in the docs?

internal/impl/datadog/output_datadog_logs.go
```diff
@@ -0,0 +1,298 @@
...
+### Limits
+
+- Maximum payload size (uncompressed): 5 MB
+- Maximum size for a single log: 1 MB
+- Maximum number of logs per batch: 1,000 entries
+
+Logs exceeding 1 MB are truncated by Datadog but still accepted (2xx). Payloads exceeding 5 MB are rejected with a 413.
+
+:::warning
+Log events can be submitted with a timestamp that is up to 18 hours in the past.
```
jem-davies: nit: I would reword this - it's warning me about something I can do - It would make more sense if it said what I cannot do?

internal/impl/datadog/output_datadog_logs.go
```diff
@@ -0,0 +1,298 @@
...
+				Description("The Datadog site to send logs to. If unset, falls back to the `DD_SITE` environment variable, then `datadoghq.com`.").
+				Optional().
+				Examples("datadoghq.com", "datadoghq.eu", "us3.datadoghq.com", "us5.datadoghq.com").
+				Advanced(),
+			service.NewInterpolatedStringField(ddoFieldDDSource).
+				Description("The source of the log, used for log processing rules.").
+				Optional(),
+			service.NewInterpolatedStringField(ddoFieldDDTags).
+				Description("A comma-separated list of tags to attach to the log.").
+				Optional(),
```
jem-davies, line 88:
```suggestion
				Description("A comma-separated list of tags to attach to the log.").
				Example(`env:${!json("environment")},version:${!json("version")}`).
				Optional(),
```

internal/impl/datadog/output_datadog_logs.go
gregfurman: How would we handle a partial batch then? I was thinking we reject the whole batch if something is wrong here since I'd imagine splitting up logs (i.e some get written, some dont) could have unintended side effects.

internal/impl/datadog/output_datadog_logs.go
gregfurman: Yeah it's just that it's unused in write calls, so didn't think to add it here.

internal/impl/datadog/output_datadog_logs.go
jem-davies: I was thinking about the [behaviour of a fallback with batched messages](https://warpstreamlabs.github.io/bento/docs/components/outputs/fallback#batching) - "Bento makes a best attempt at inferring which specific messages of the batch failed" - That is the purpose of the BatchError isn't it?  

Say you had a batch of 10 messages and one had an issue (too big) - then that one should carry on through any DLQ / Error handling pattern - if there isn't a DLQ it gets nacked. 

Without the batch error you will allow messages that have been delivered through the DLQ etc?

jem-davies (approved):

PR #736, opened 2026-03-10
add cache collector

### LINKED ISSUE #711: Add cache collector
For complexer work it would be nice to have something like this:

```yaml
cache_collector:
  resource: cache_resource
  key: ""
  init: root = []
  append_check: this.process == "work"
  append_map: root = this.old.append(this.new)
  flush_check: this.process == "end"
  flush_deletes: true
  flush_map: root = this
```

logic for each message (process):
- get cache by resource config
- check if append_check returns true on current message
- if append_check bloblang returns true
	- read the key from key config in cache as current value
	- if not exists call the bloblang init function which is now the current value
	- call append_map bloblang function with "this.old" (the current value) and "this.new" (the current message), the return value is then the current value
	- store the value into the cache
	- return nothing from processing (no message)
- if flush_check bloblang true
	- read the key from key config in cache as current value
	- if not exists call the bloblang init function which is now the current value
	- if flush_deletes is true the cache key gets removed
	- call flush_map bloblang function which returns from processing the new message (replaces the current message)
- if both checks are not true
	- just return the message as it is

lublak:
fixes #711


jem-davies (commented):
Thanks for this - but I am not convinced that we should add this processor to
Bento for a few reasons: 

 - Concurrency problems with append requiring two separate non-atomic operations on the cache_collector
 - The implementation doesn't require access to any packages in /internal/ and therefore could be implemented as a plug-in.
 - It would be possible to compose together already existing components to implement similar functionality

It is a shame that the current interface for cache doesn't allow locking/unlocking such it would be possible to implement 
an append operation...

lublak:
@jem-davies i understand, i use this myself and find it very usefull (and if locking for cache exists, some days, i can also improve it in the concurrency problems).
Yes it can be composed together (just as an example: it shrinks my config half and improves readability).
But i can also understand if it doesn't fit.

jem-davies:
> @jem-davies i understand, i use this myself and find it very usefull 

Ok thanks again 

> Yes it can be composed together (just as an example: it shrinks my config half and improves readability)

Yes but then I would say that it would be subjective because the complexity will either exist in the YAML config or in the Go Processor implementation.

PR #740, opened 2026-03-10
feat(websocket): add headers support to websocket output

### LINKED ISSUE #739: feat(websocket): add headers support to websocket output
The `websocket` output component does not currently support custom headers
on the WebSocket handshake, unlike the `websocket` input which already has
this capability.

This makes it impossible to connect to servers that require headers for
authentication or protocol negotiation (e.g. `Authorization`,
`Sec-WebSocket-Protocol`).

**Proposed solution:**
Add an interpolated string map `headers` field to the `websocket` output,
consistent with the existing implementation in the `websocket` input.
Static and Bloblang-interpolated values (e.g. `${! env("TOKEN") }`) would
both be supported.

I'm opening a PR for this now.

DackJempsey:
## Summary

- Adds a `headers` field to the `websocket` output component
- Supports static and Bloblang-interpolated header values (e.g. `${! env("TOKEN") }`)
- Mirrors the existing `headers` implementation in the `websocket` input
- Adds unit tests covering static, interpolated, and mixed header cases

## Example config
```
websocket:
  url: ws://example.com/ws
  headers:
    Authorization: 'Bearer ${! env("TOKEN") }'
    Sec-WebSocket-Protocol: graphql-ws
```

## Test plan
- [x] `make test` passes - failures unrelated
- [x] `make docs` run to regenerate docs for new config field

Closes #739

### DIFF
--- a/internal/impl/io/output_websocket_test.go
+++ b/internal/impl/io/output_websocket_test.go
@@ -2,6 +2,7 @@ package io
 
 import (
 	"context"
+	"fmt"
 	"net/http"
 	"net/http/httptest"
 	"net/url"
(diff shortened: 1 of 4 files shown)


internal/impl/io/output_websocket_test.go
```diff
@@ -70,6 +71,108 @@ websocket:
...
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			ctx, done := context.WithTimeout(context.Background(), time.Second*30)
+			defer done()
+
+			for k, v := range test.envVar {
+				t.Setenv(k, v)
+			}
```
gregfurman, line 125: Nice approach here!

Issue #737, opened 2026-03-10
Create dedicated DuckDB appender component for bulk inserts


gregfurman:
Currently, we allow for interacting with DuckDB via the SQL components -- which uses the duckdb sql driver under-the-hood.

However, the Appender API offers bulk insertions which would be useful as a sink.

See https://duckdb.org/docs/stable/clients/go#appender

gregfurman:
@iamramtin You keen to pick this up?

iamramtin:
> [@iamramtin](https://github.com/iamramtin) You keen to pick this up?

Would be an honour 🦆

PR #741, opened 2026-03-10
feat(duckdb): Add duckdb_append component

iamramtin:
## Add `duckdb_append` Component using Appender API

Adds a new `duckdb_append` batch output component that writes rows directly into DuckDB's columnar storage via `NewAppenderFromConn`, bypassing SQL parsing entirely.

#### Flush on every batch

`Flush()` is currently invoked at the end of every `WriteBatch`.

This makes rows **immediately visible** after each batch completes but this may result in row groups being smaller than optimal. DuckDB's `STANDARD_VECTOR_SIZE` is **2048 rows**, see https://duckdb.org/docs/stable/internals/vector.

If there's a need, we can update this in the future to flush only on `Close()`.

#### Type coercions

Bloblang deserialises JSON numbers as `json.Number`. These are converted to `int64` / `float64` before passed to `AppendRow`.

RFC3339 strings are parsed to `time.Time` for `TIMESTAMP` columns.

All other types pass through unchanged.

---

References issue https://github.com/warpstreamlabs/bento/issues/737


iamramtin:
To test this out, build Bento with the build tag via: `go build -o bento -tags x_bento_extra ./cmd/bento/main.go`
then run: `./bento -c ducks.yaml`.

Here's a basic config to test this out.

```
# ducks.yaml
input:
  generate:
    interval: 200ms
    count: 25
    mapping: |
      let ducks = [
        "Scrooge McDuck",
        "Donald Duck",
        "Howard the Duck",
        "Daffy Duck",
        "Psyduck",
      ]

      let items = [
        "Gold Coin",
        "Ancient Bread Crumb",
        "Bottle of Gin",
        "Lucky Magic Dime",
        "Half-Eaten Bento Box",
        "Can of Corn",
        "Shiny Pebble",
        "Mysterious Pond Key"
      ]

      let locations = [
        "Duckburg Vault",
        "Trash Can",
        "Secret Bank",
        "Random Pond",
        "Old Mill Coin Stash",
        "Kitchen Drain"
      ]

      root.deposit_id   = uuid_v4()
      root.item         = $items.index(random_int() % $items.length())
      root.location     = $locations.index(random_int() % $locations.length())
      root.quantity     = (random_int() % 99) + 1
      root.deposited_at = now().format_timestamp("2006-01-02T15:04:05Z07:00")

      let is_goose = (random_int() % 10) == 0

      root.duck   = if $is_goose { "Suspicious Goose" } else { $ducks.index(random_int() % $ducks.length()) }
      root.emoji  = if $is_goose { "🪿" } else { "🦆" }
      root.action = if $is_goose { "stole" } else { "deposited" }

pipeline:
  processors:
    - log:
        level: INFO
        message: |
          ${! this.emoji } ${! this.duck } waddled into ${! this.location }
          💰 ${! this.action } ${! this.quantity } × ${! this.item }
          🧾 receipt=${! this.deposit_id.slice(0,8) }  time=${! this.deposited_at }

output:
  broker:
    outputs:
      - duckdb_append:
          dsn: /tmp/vault.duckdb
          table: vault_deposits
          columns: [deposit_id, duck, item, quantity, deposited_at]
          args_mapping: |
            root = [
              this.deposit_id,
              this.duck,
              this.item,
              this.quantity,
              this.deposited_at,
            ]
          init_statement: |
            CREATE TABLE IF NOT EXISTS vault_deposits (
              deposit_id   VARCHAR PRIMARY KEY,
              duck         VARCHAR      NOT NULL,
              item         VARCHAR      NOT NULL,
              quantity     BIGINT       NOT NULL,
              deposited_at TIMESTAMP    NOT NULL
            )
          batching:
            count: 10
            period: 1s

      - stdout: {}
```

You should see an output like so:

```
🦆 Donald Duck waddled into Trash Can
💰 deposited 41 × Half-Eaten Bento Box
🧾 receipt=042de1ab  time=2026-03-14T09:05:33+02:00  @service=bento label="" path=root.pipeline.processors.0
{"action":"deposited","deposit_id":"042de1ab-9b89-4c9a-bbcb-2c7c28acef87","deposited_at":"2026-03-14T09:05:33+02:00","duck":"Donald Duck","emoji":"🦆","item":"Half-Eaten Bento Box","location":"OTrash Can","quantity":41}

🦆 Scrooge McDuck waddled into Random Pond
💰 deposited 85 × Mysterious Pond Key
🧾 receipt=f7517442  time=2026-03-14T09:00:15+02:00  @service=bento label="" path=root.pipeline.processors.0
{"action":"deposited","deposit_id":"f7517442-5fbc-4683-ae22-b716a91edf34","deposited_at":"2026-03-14T09:00:15+02:00","duck":"Scrooge McDuck","emoji":"🦆","item":"Mysterious Pond Key","location":"Random Pond","quantity":85}

🦆 Howard the Duck waddled into Old Mill Coin Stash
💰 deposited 85 × Bottle of Gin
🧾 receipt=2ff9390f  time=2026-03-14T09:00:12+02:00  @service=bento label="" path=root.pipeline.processors.0
{"action":"deposited","deposit_id":"2ff9390f-4a1f-431a-9d8e-353eccf08a98","deposited_at":"2026-03-14T09:00:12+02:00","duck":"Howard the Duck","emoji":"🦆","item":"Bottle of Gin","location":"Old Mill Coin Stash","quantity":85}

🪿 Suspicious Goose waddled into Secret Bank
💰 stole 60 × Gold Coin
🧾 receipt=3f51aac0  time=2026-03-14T09:00:14+02:00  @service=bento label="" path=root.pipeline.processors.0
{"action":"stole","deposit_id":"3f51aac0-2312-4ff9-b9bf-00016a6dadde","deposited_at":"2026-03-14T09:00:14+02:00","duck":"Suspicious Goose","emoji":"🪿","item":"Gold Coin","location":"Secret Bank","quantity":60}
```

iamramtin:
CI passes now after rebasing off https://github.com/warpstreamlabs/bento/pull/729 to fix the doc issues. @gregfurman @jem-davies

Ready for review 👀

internal/impl/sql/output_duckdb_insert.go
```diff
@@ -0,0 +1,343 @@
...
+	case <-ctx.Done():
+		return ctx.Err()
+	}
+
+	return nil
+}
+
+func (d *duckdbAppendOutput) WriteBatch(ctx context.Context, batch service.MessageBatch) error {
+	d.mu.Lock()
+	defer d.mu.Unlock()
```
gregfurman: This will place a write-lock around the entire `WriteBatch` call. Think we should rather just lock once with a read-lock, retrieve a reference to the executor & appender, and then release the lock and proceed.

i.e

```golang
d.mu.RLock()
if d.appender == nil {
        d.mu.RUnlock()
}
appender := d.appender 

var executor *service.MessageBatchBloblangExecutor
if d.argsMapping != nil {
   executor = batch.BloblangExecutor(d.argsMapping)
}
d.mu.RUnlock()
```

internal/impl/sql/output_duckdb_append.go
```diff
@@ -0,0 +1,343 @@
...
+		return nil, err
+	}
+
+	return d, nil
+}
+
+func (d *duckdbAppendOutput) Connect(ctx context.Context) error {
+	d.mu.Lock()
+	defer d.mu.Unlock()
+
```
gregfurman, line 224: Again, I don't think we should be locking the entire `Connect` call if it can be avoided 👀

internal/impl/sql/output_duckdb_insert.go
```diff
@@ -0,0 +1,343 @@
...
+		d.shutSig.TriggerHasStopped()
+	}()
+
+	return nil
+}
+
+func (d *duckdbAppendOutput) Close(ctx context.Context) error {
+	d.shutSig.TriggerHardStop()
+
+	d.mu.Lock()
```
gregfurman: Image

internal/impl/sql/output_duckdb_append_test.go
gregfurman, line 1: We have both a duckdb reader and writer now, correct?

Could be worth writing an integration test using:

https://github.com/warpstreamlabs/bento/blob/2032cb588e6fbbf7d207fdba4bc957b8f6b96b31/public/service/integration/stream_test_helpers.go#L232-L234

with a duckdb input and a duckdb appender output 👀 

Here's a GCP example using this util but you can adapt it to ingest from a `sql_input` and output to a `duckdb_append` 

https://github.com/warpstreamlabs/bento/blob/2032cb588e6fbbf7d207fdba4bc957b8f6b96b31/internal/impl/gcp/integration_test.go#L114-L141

gregfurman (changes requested):

internal/impl/sql/output_duckdb_insert.go
iamramtin: Have updated, thanks for pointing this out!

internal/impl/sql/output_duckdb_append.go
iamramtin: Are you sure we don't need to acquire a full lock here on the `Connect` call? I see similar behaviour in existing SQL components; if multiple goroutines call `Connect()` simultaneously would this be safe or would they race to create `d.db`, `d.sqlConn`, and `d.appender`?

gregfurman: Fair point. A single write lock here simplifies things... but I'm just concerned that a `Connect` call hanging could cause a `Close` call to block.

However, since we do pass a `ctx` in the param, I'd expect it being cancelled to allow us to exit early -- which should hopefully be handled by bento's internal component manager but this is definitely out-of-scope for this PR 🤷

internal/impl/sql/output_duckdb_append.go
```diff
@@ -0,0 +1,348 @@
...
+					driverArgs[j] = v
+				}
+			default:
+				driverArgs[j] = v
+			}
+		}
+
+		if err := appender.AppendRow(driverArgs...); err != nil {
+			return fmt.Errorf("appending row %d: %w", i, err)
+		}
```
gregfurman: suggestion(non-blocking): if we can assume that each row-operation in the batch is independant, this could be a nice use-case for a `service.BatchError` which allows us to specify exactly which messages in a batch fail to be appended.

e.g the `WriteBatch` call of `outputs/couchbase`:

https://github.com/warpstreamlabs/bento/blob/f570aa4612a7aca585456f29e3931388d398b538/internal/impl/couchbase/output.go#L61-L81

internal/impl/sql/output_duckdb_append.go
```diff
@@ -0,0 +1,348 @@
...
+			var ok bool
+			if args, ok = iargs.([]any); !ok {
+				return fmt.Errorf("mapping returned non-array result: %T", iargs)
+			}
+			if len(args) != len(d.columns) {
+				return fmt.Errorf("mapping returned %d values for %d columns", len(args), len(d.columns))
+			}
+		}
+
+		driverArgs := make([]driver.Value, len(args))
```
gregfurman: q: what is the purpose of these `driverArgs`?

internal/impl/sql/output_duckdb_append.go
```diff
@@ -0,0 +1,348 @@
...
+
+func (d *duckdbAppendOutput) startShutdownListener() {
+	go func() {
+		<-d.shutSig.HardStopChan()
+
+		d.mu.Lock()
+		defer d.mu.Unlock()
+
+		if d.appender != nil {
+			if err := d.appender.Close(); err != nil {
```
gregfurman, line 197: suggestion(non-blocking): Thoughts of a `DEBUG` log here saying we're closing the DDB appender early and that this will trigger a flush of any pending operations previously loaded? Either that or a note in the component spec would be good.

From the `Close` docs:

> Close the appender. This will flush the appender to the underlying table.

internal/impl/sql/output_duckdb_append.go
```diff
@@ -0,0 +1,348 @@
...
+					driverArgs[j] = n.String()
+				}
+			case string:
+				if t, err := time.Parse(time.RFC3339, n); err == nil {
+					driverArgs[j] = t
+				} else {
+					driverArgs[j] = v
+				}
+			default:
+				driverArgs[j] = v
```
gregfurman: What if we have a nested value for `v`? Should we not be running this mapping functionality recursively to build up what these args should look like?

internal/impl/sql/output_duckdb_append.go
```diff
@@ -0,0 +1,348 @@
...
+				return fmt.Errorf("mapping returned non-array result: %T", iargs)
+			}
+			if len(args) != len(d.columns) {
+				return fmt.Errorf("mapping returned %d values for %d columns", len(args), len(d.columns))
+			}
+		}
+
+		driverArgs := make([]driver.Value, len(args))
+		for j, v := range args {
+			switch n := v.(type) {
```
gregfurman: Can we add a test where we attempt to write in composite values to a ddb table i.e `STRUCT`, `LIST`, and `JSON` types

i.e https://github.com/duckdb/duckdb-go/blob/a2e25da13c75b87eb96da81bdca1b57bac8e4bc0/types_test.go#L638-L650

Also, it could be worth looking at how Cassandra handles mapping 👀 

https://github.com/warpstreamlabs/bento/blob/f570aa4612a7aca585456f29e3931388d398b538/internal/impl/cassandra/output.go#L304

internal/impl/sql/output_duckdb_append.go
```diff
@@ -0,0 +1,348 @@
...
+	var executor *service.MessageBatchBloblangExecutor
+	if d.argsMapping != nil {
+		executor = batch.BloblangExecutor(d.argsMapping)
+	}
+
+	for i := range batch {
+		var args []any
+
+		if executor != nil {
+			resMsg, err := executor.Query(i)
```
gregfurman: Q: If my args mapping explicitly maps to the correct type i.e

```coffee
[ this.foo.int(), this.bar.float() ]
```

Would those below type switches be necessary?

gregfurman (changes requested):

internal/impl/sql/output_duckdb_append.go
iamramtin: No wouldn't be necessary as the coercion would have been applied on the Bloblang side first. I tested it using `.int64()`,  `.float64()`, etc. I assume same applies with `.ts_parse()` for timestamps

internal/impl/sql/output_duckdb_append_test.go
iamramtin: wasn't entirely sure how `StreamTests` works but it seems to be for roundtrip testing where it sends its own payloads to the output and then reads them back from the input. Is that correct? DuckDB appender expects structured columns not raw strings, and has no input component. So I dont think `StreamTests` fits, correct me if I'm wrong. 

But I saw some other components like couchbase and dynamodb use `service.NewStreamBuilder`. So I sued that instead to append multiple rows and then open a new SQL connection to the same file to query results.

LMK if you're happy with this approach

internal/impl/sql/integration_test.go
```diff
@@ -1683,3 +1683,52 @@ output:
 		return countAfterRestart > count+100
 	}, time.Minute, time.Second)
 }
+
+func TestIntegrationDuckDBAppender(t *testing.T) {
```
gregfurman: This should be moved into the `internal/impl/sql/integration_cgo_test.go` instead since ddb wont work in the normal distro, hence the test will fail

internal/impl/sql/output_duckdb_append_test.go
gregfurman: Since `duckdb_append` resides in the `sql` package, I was suggesting we use a `sql_select` as the input layer with the new `duckdb_append` as the output i.e

```yaml
input:
  sql_select:
    driver: duckdb
    dsn: /path/to/test.duckdb
    table: vault_deposits
    columns: [ "*" ]

output:
  duckdb_append:
    dsn: /path/to/test.duckdb
    table: vault_deposits
    columns: [duck, gold_coins]
    args_mapping: "root = [this.duck, this.gold_coins]"
    init_statement: |
      CREATE TABLE IF NOT EXISTS vault_deposits (
        duck       VARCHAR,
        gold_coins BIGINT
      )
```

However, what you've done with the `generate` is fine 👍 The integration test fixtures just allow for more exhaustive testing of a component i.e

```go
// ...
integration.StreamTests(
	integration.StreamTestOpenClose(), // can the component open and close correctly
	integration.StreamTestStreamSequential(20), // can the component ingest data sequentially
	integration.StreamTestSendBatchCount(10), // can the component 
	integration.StreamTestStreamParallelLossyThroughReconnect(20),
).Run(
// ...
```

internal/impl/sql/output_duckdb_append_test.go
```diff
@@ -0,0 +1,451 @@
+//go:build x_bento_extra
```
gregfurman, line 1: nit(non-blocking): Using test-tables would be nice here since there is a lot of overlapping functionality between tests but we can add an inline-TODO instead

internal/impl/sql/output_duckdb_append.go
```diff
@@ -0,0 +1,387 @@
...
+
+	argsMapping  *bloblang.Executor
+	connSettings *connSettings
+
+	shutSig *shutdown.Signaller
+	logger  *service.Logger
+}
+
+func DuckDBAppendOutputConfig() *service.ConfigSpec {
+	return duckdbAppendOutputConfig()
```
gregfurman: maybe remove the private one and just define the configspec in this public func?

internal/impl/sql/output_duckdb_append.go
```diff
@@ -0,0 +1,387 @@
...
+	select {
+	case <-d.shutSig.HasStoppedChan():
+	case <-ctx.Done():
+		return ctx.Err()
+	}
+
+	return nil
+}
+
+func (d *duckdbAppendOutput) WriteBatch(ctx context.Context, batch service.MessageBatch) error {
```
gregfurman, line 283: Think it's worth giving these guidelines a lookthrough https://github.com/marcboeker/go-duckdb/discussions/277

Seems they suggest we create a new appender + db conn for each append call. This would also allow us to parallelize calls to `WriteBatch` since each call could have it's own appender spun up 🤔 

Anyway, think we can leave this as-is for now but future implementation could make use of a `sync.Pool` to spin up and teardown appenders dynamically while minimizing overhead of creation (due to resource pooling + object re-use).

gregfurman (changes requested):
Think once we move the integration test to the correct place, am happy to approve!

internal/impl/sql/integration_test.go
iamramtin: Done

internal/impl/sql/output_duckdb_append.go
iamramtin: Ah thanks for pointing this out. I've added a TODO to address this, happy to follow up with this once we merge this in. Feel free to create an issue for it and I'll grab it when I have a chance.

gregfurman (approved):
LGTM! Let's ship it 🦆 :shipit:

PR #755, opened 2026-03-15
chore(ci): Improve govulncheck visibility

gregfurman:
This PR simplifies the `govulncheck` CI job to report a job failure if vulnerabilities are found.

In addition:
- This job will run when a PR changes the `go.mod` or `go.sum` -- allowing us to check if a changes will introduce a vuln.
- Runs every day at 8am UTC
- Introduces a badge on the README indicating whether the `govulncheck` is failing -- helping with visibility of the vuln status instead of silently failing.

### DIFF
--- a/.github/workflows/govulncheck.yml
+++ b/.github/workflows/govulncheck.yml
@@ -20,7 +20,7 @@ jobs:
   govulncheck:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
         with:
           persist-credentials: false
       - uses: actions/setup-go@v6


.github/workflows/govulncheck.yml
```diff
@@ -1,35 +1,30 @@
...
-  check-for-vulnerabilities:
-    name: Check for vulnerabilities using `govulncheck`
+  govulncheck:
     runs-on: ubuntu-latest
-    permissions:
-      security-events: write
-      contents: read
     steps:
-      - uses: golang/govulncheck-action@v1.0.4
+      - uses: actions/checkout@v5
```
miparnisari:
```suggestion
      - uses: actions/checkout@v6
```

gregfurman:
@miparnisari Thanks for the extra eyes! Once CI passes I'll merge :shipit:

PR #758, opened 2026-03-16
kafka+oauth2 implementation

### LINKED ISSUE #437: Add OAuth to Kafka
Hi, I am trying to connect to my kafka instance which manages it's authentication and authorization via keycloak. I see that the kafka component allows for oauthbearer mechanism, but only accepts an oauth token. I would love for it to fetch its token autonomously from keycloak.

Currently I am working on building a soulution via the cache. Yet I've seen in the [doc](https://warpstreamlabs.github.io/bento/docs/components/inputs/http_client) that the http-client allows for oauth2 authentication by client-id ("client-key") and client-secret like so
```yaml
input:
  label: ""
  http_client:
    ...
    oauth2:
      enabled: false
      client_key: ""
      client_secret: ""
      token_url: ""
      scopes: []
      endpoint_params: {}
    ...
```
I guess the implementation is this one: https://github.com/warpstreamlabs/bento/blob/main/internal/httpclient/auth_oauth2.go.

Would it be possible to fix something similar for the [kafka component](https://github.com/warpstreamlabs/bento/tree/main/internal/impl/kafka)? I think it would fit nicely in the sasl definition as an option to specifying the token directly (access_token) or in cache (token_cache, token_key):
```yaml
input:
  label: ""
  kafka:
    ...
    sasl:
      mechanism: OAUTHBEARER
      access_token: undefined
      token_cache: undefined
      token_key: undefined
      oauth2:
        client_key: <cliend-id>
        client_secret: <client-secret>
        token_url: <oauth_token_uri>
        scopes: []
        endpoint_params: {}
    ...
```

vsl86:
resolves https://github.com/warpstreamlabs/bento/issues/437.
config for both kafka/franz:
  
      oauth2:
        enabled: <bool>
        client_key: <id>
        client_secret: <secret>
        token_url: <://your/oidc/realm/token>
        scopes: [] #optional
        endpoint_params: {} #optional

### DIFF
--- a/internal/impl/kafka/sasl.go
+++ b/internal/impl/kafka/sasl.go
@@ -14,6 +14,8 @@ import (
 	"github.com/twmb/franz-go/pkg/sasl/oauth"
 	"github.com/twmb/franz-go/pkg/sasl/plain"
 	"github.com/twmb/franz-go/pkg/sasl/scram"
+	"golang.org/x/oauth2"
+	"golang.org/x/oauth2/clientcredentials"
 )
 
 func notImportedAWSFn(c *service.ParsedConfig) (sasl.Mechanism, error) {
@@ -50,6 +52,30 @@ func saslField() *service.ConfigField {
 		service.NewStringField("token").
 			Description("The token to use for a single session's OAUTHBEARER authentication.").
 			Default(""),
+		service.NewObjectField("oauth2",
+			service.NewBoolField("enabled").
+				Description("Whether to use OAuth version 2 in requests.").
+				Default(false),
+			service.NewStringField("client_key").
+				Description("A value used to identify the client to the token provider.").
+				Default(""),
+			service.NewStringField("client_secret").
+				Description("A secret used to establish ownership of the client key.").
+				Default("").Secret(),
+			service.NewURLField("token_url").
+				Description("The URL of the token provider.").
+				Default(""),
+			service.NewStringListField("scopes").
+				Description("A list of optional requested permissions.").
+				Default([]any{}).
+				Advanced(),
+			service.NewAnyMapField("endpoint_params").
+				Description("A list of optional endpoint parameters, values should be arrays of strings.").
+				Advanced().
+				Optional(),
+		).
+			Description("Allows you to specify open authentication via OAuth version 2 using the client credentials token flow.").
+			Optional().Advanced(),
 		service.NewStringMapField("extensions").
 			Description("Key/value pairs to add to OAUTHBEARER authentication requests.").
 			Optional(),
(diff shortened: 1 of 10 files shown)


jem-davies:
I would like to add an integration test for this.

jem-davies:
@vsl86 - hey I have got a couple of integration tests over on the branch: https://github.com/warpstreamlabs/bento/pull/847 - I noticed that there is an issue with the way the `kafka` (sarama) components parse the provided SASL OAUTHBEARER config. 

I will get it working tomorrow but wanted to give this update - sorry it's been a while - thanks for the effort on this PR

internal/impl/kafka/sasl.go
```diff
@@ -136,6 +162,63 @@ func plainSaslFromConfig(c *service.ParsedConfig) (sasl.Mechanism, error) {
...
+					return nil, err
+				}
+				for k, v := range params {
+					if endpointParams[k], err = v.FieldStringList(); err != nil {
+						return nil, err
+					}
+				}
+			}
+			var extensions map[string]string
+			if c.Contains("extensions") {
```
jem-davies: Should we move this into the "oauth2" object field? - It's not referenced from any other SASL mechanisms?

internal/impl/kafka/sasl.go
```diff
@@ -239,6 +324,34 @@ func SaramaSASLField() *service.ConfigField {
 		service.NewStringField(saramaFieldSASLTokenKey).
 			Description("Required when using a `token_cache`, the key to query the cache with for tokens.").
 			Default(""),
+		service.NewAnyMapField(saramaFieldSASLExtensions).
+			Description("A list of optional endpoint parameters, values should be arrays of strings.").
+			Advanced().
+			Optional(),
```
jem-davies: Should we move this into the "oauth2" object field? - It's not referenced from any other SASL mechanisms? 

Also the Description text here is the same as "oauth2.endpoint_params" - are these the same?

jem-davies (dismissed):
Looks good thanks 😄  - just a couple of small things to consider: 

- "extensions" in the `kafka` sarama components, this field is only referenced inside the client credentials method of OAUTHBEARER SASL mechanism, so should we move that into the `sasl.oauth2` config group? 
- It has the same description as `endpoint_params` - in the docstring from sarama: `Extensions is a optional map of arbitrary key-value pairs that can be sent with the SASL/OAUTHBEARER initial client response.` - Are they sent as endpoint_params 🤔 

Also perhaps worth some documentation in both `kafka_franz` & `kafka` to explain the 3 ways of OAUTHBEARER set-ups: 

 - `token` (franz) & `access_token` (sarama) which allow for static tokens
 - `token_cache` (sarama) which allow for fetching tokens from a cache
 - `sasl.oauth2` (franz & sarama) 

And I have added an int. test on: https://github.com/vsl86/bento/pull/1 & a fix too.

Thanks for the effort on this 🙏

internal/impl/kafka/sasl.go
```diff
@@ -239,6 +324,34 @@ func SaramaSASLField() *service.ConfigField {
...
+		service.NewObjectField(saramaFieldSASLOAuth2,
```
jem-davies, line 340: Can add `.Version("1.18.0")` to these fields as well,

PR #759, opened 2026-03-16
Add handling for authentication errors in SQL output modules

adrianhaj:
Add handling for authentication errors in SQL output modules:

- Added `isAuthError` helper function to identify authentication-related connection errors.
- Modified SQL output modules to trigger a reconnection on authentication errors.
- Introduced a test to verify behavior when PAM authentication errors occur.

### DIFF
--- a/internal/impl/sql/conn_fields_test.go
+++ b/internal/impl/sql/conn_fields_test.go
@@ -1,93 +1,365 @@
-package sql
+package sql_test
 
 import (
+	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
+	"os"
+	"path/filepath"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	baws "github.com/warpstreamlabs/bento/internal/impl/sql/aws"
+
+	"github.com/warpstreamlabs/bento/public/service"
+
+	_ "github.com/warpstreamlabs/bento/public/components/pure"
+	_ "github.com/warpstreamlabs/bento/public/components/sql"
 )
 
-type fakeSQLStateError struct {
-	state string
-	msg   string
+func TestConnSettingsInitStmt(t *testing.T) {
+	tCtx, done := context.WithTimeout(context.Background(), time.Second*30)
+	defer done()
+
+	tmpDir, err := os.MkdirTemp("", t.Name())
+	require.NoError(t, err)
+	t.Cleanup(func() {
+		os.RemoveAll(tmpDir)
+	})
+
+	outputConf := fmt.Sprintf(`
+sql_insert:
+  driver: sqlite
+  dsn: file:%v/foo.db
+  table: things
+  columns: [ foo, bar, baz ]
+  args_mapping: 'root = [ this.foo, this.bar, this.baz ]'
+  init_statement: |
+    CREATE TABLE IF NOT EXISTS things (
+      foo varchar(50) not null,
+      bar varchar(50) not null,
+      baz varchar(50) not null,
+      primary key (foo)
+    ) WITHOUT ROWID;
+`, tmpDir)
+
+	streamInBuilder := service.NewStreamBuilder()
+	require.NoError(t, streamInBuilder.SetLoggerYAML(`level: OFF`))
+	require.NoError(t, streamInBuilder.AddOutputYAML(outputConf))
+
+	inFn, err := streamInBuilder.AddBatchProducerFunc()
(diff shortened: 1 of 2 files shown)


internal/impl/sql/output_sql_raw_test.go
```diff
@@ -0,0 +1,82 @@
...
+		dsn:          "test",
+		db:           db,
+		queryStatic:  "INSERT INTO foo VALUES (1)",
+		connSettings: &connSettings{},
+		logger:       service.MockResources().Logger(),
+		shutSig:      shutdown.NewSignaller(),
+	}
+}
+
+func TestSQLRawOutputReconnectsOnPAMAuthError(t *testing.T) {
```
gregfurman: Can you not just set the function here to return an auth error instead of this mock approach?

https://github.com/warpstreamlabs/bento/blob/018a0137fa650fe84ad786998f5104994af04c2a/internal/impl/sql/conn_aws.go#L46-L48

Perhaps just wrap the original with one that always returns an error overriding what is done in:

https://github.com/warpstreamlabs/bento/blob/018a0137fa650fe84ad786998f5104994af04c2a/internal/impl/sql/aws/aws.go#L82

internal/impl/sql/output_sql_raw_test.go
gregfurman, line 1: Should this test (or some variation of it) not perhaps rather be in internal/impl/sql/conn_fields_test.go?

internal/impl/sql/conn_fields.go
```diff
@@ -401,3 +401,15 @@ func sqlOpenWithReworks(ctx context.Context, logger *service.Logger, driver, dsn
 
 	return db, nil
 }
+
+func isAuthError(err error) bool {
+	if err == nil {
+		return false
+	}
+	msg := strings.ToLower(err.Error())
+	return strings.Contains(msg, "authentication failed") ||
```
gregfurman: I'm not sold on this approach. Perhaps we add a specific case for this PAM error instead of the string matching like this. 

```
PAM authentication failed for user ...
```

Perhaps the Postgres library or AWS sdk has an error defined somewhere for this case?

internal/impl/sql/conn_fields.go
```diff
@@ -401,3 +401,15 @@ func sqlOpenWithReworks(ctx context.Context, logger *service.Logger, driver, dsn
 
 	return db, nil
 }
+
+func isAuthError(err error) bool {
```
gregfurman: Could be worth adding an additional param here for `driver` that allows us to run specific auth checks depending on the SQL driver in use?

Like, postgres has different error messages to mysql. which will have different errors to duckdb etc.

then we can do a switch case based on driver + error strings (or ideally `errors.Is` checks).

internal/impl/sql/output_sql_raw_test.go
adrianhaj: Done. Thank you for your comment.

internal/impl/sql/conn_fields.go
adrianhaj: Thank you. Done.

internal/impl/sql/conn_fields.go
adrianhaj: Good point. I changed the approach because only two databases are supported for IAM: PostgreSQL and MySQL. https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.IAMDBAuth.Enabling.html

adrianhaj:
@gregfurman Everything has been updated in accordance with the review. Please check it. Thank you!

internal/impl/sql/conn_fields_test.go
jem-davies, line 1: How come we are getting rid of all these tests?

adrianhaj: @jem-davies my bad :( sorry for this. I will revert it.

adrianhaj: @jem-davies I'm using different file for testing because `conn_fields_test.go` is using `sql_test` package. Please check it.

adrianhaj:
@gregfurman @jem-davies I moved my test do different file because `conn_fields_test.go` is using `sql_test` and I don't want to expose`isAuthError`.

internal/impl/sql/output_sql_raw_test.go
adrianhaj: done

adrianhaj:
Could you please re-review it again @gregfurman @jem-davies 🙏

PR #764, opened 2026-03-17
Grpc client input

jem-davies:
Changes: 

 - Adds a new input component `grpc_client` 
 - Refactors the existing `internal/impl/grpc_client` package to `internal/impl/grpc`
 
 ------
 
 TODO 
 
 - [ ] ~~Add more tests~~ 
 - [X] Consider shutdown procedure with rpc_type `bidi`
 - [ ] ~~Consider long-lived server streams: currently we await closure of the stream prior to creation of Bento message~~ -> will raise an issue to implement a new config field that will enable server side streams be marshalled into separate bento message


internal/impl/grpc/grpc_test_server/helloworld/testdata/certs/ca.key
gregfurman, line 1: Can all of these go in a `testdata` directory? That way, we ensure they're never included in any of the builds.

internal/impl/grpc/grpc_test_server/test_server/test_server.go
```diff
@@ -0,0 +1,296 @@
...
+}
+
+func StartGRPCServer(t *testing.T, opts ...TestServerOpt) *TestServer {
+	t.Helper()
+
+	testServer := &TestServer{}
+
+	for _, o := range opts {
+		o(testServer)
+	}
```
gregfurman: nit: 
```suggestion
	for _, opt := range opts {
		opt(testServer)
	}
```

internal/impl/grpc/grpc_test_server/test_server/test_server.go
```diff
@@ -0,0 +1,296 @@
...
+			"./grpc_test_server/helloworld/certs/server.key",
+		)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		caCert, err := os.ReadFile("./grpc_test_server/helloworld/certs/ca.pem")
+		if err != nil {
+			t.Fatal(err)
+		}
```
gregfurman, line 79: nit: wdyt of including these assets using the `embed` directory? That way, we can ensure they're always present -- making many of these checks redundant since we know the files/data exists i.e

```golang
var (
   //go:embed grpc_test_server/helloworld/certs/ca.pem
   serverPEM []byte
)
```

internal/impl/grpc/grpc_test_server/test_server/test_server.go
```diff
@@ -0,0 +1,296 @@
...
+		tsOAuth2 := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+
+			username, password, ok := r.BasicAuth()
+			require.True(t, ok)
+			require.Equal(t, "fookey", username)
+			require.Equal(t, "foosecret", password)
+
+			b, err := io.ReadAll(r.Body)
+			require.NoError(t, err)
+			assert.Equal(t, "grant_type=client_credentials", string(b))
```
gregfurman: super-nit: can we align the usage of `require` and `assert`? I personally prefer `require` but 🤷

internal/impl/grpc/common.go
```diff
@@ -0,0 +1,343 @@
...
+				"bar": []string{"woof"},
+			}).
+			Default(map[string]any{}).
+			Optional(),
+	).
+		Description("Allows you to specify open authentication via OAuth version 2 using the client credentials token flow.").
+		Optional().Advanced()
+}
+
+type grpcCommon struct {
```
gregfurman: Can this perhaps be renamed to `grpcConfig`?

internal/impl/grpc/common.go
```diff
@@ -0,0 +1,343 @@
...
+	grpcClientTLS                    = "tls"
+	grpcClientHealthCheck            = "health_check"
+	grpcClientHealthCheckToggle      = "enabled"
+	grpcClientHealthCheckServiceName = "service"
+
+	rpcTypeUnary        = "unary"
+	rpcTypeClientStream = "client_stream"
+	rpcTypeServerStream = "server_stream"
+	rpcTypeBidi         = "bidi"
+
```
gregfurman, line 38: nit: Can we consider making this an enum? Would make the relationship between the `rpcType`attribute in the `grpcCommon` struct and these values clearer IMO i.e

```golang
type rpcType string
var unary rpcType = "unary"
```

internal/impl/grpc/common.go
```diff
@@ -0,0 +1,343 @@
...
+	healthCheckEnabled, err := healthCheckConf.FieldBool(grpcClientHealthCheckToggle)
+	if err != nil {
+		return grpcCommon{}, err
+	}
+	healthCheckServiceName, err := healthCheckConf.FieldString(grpcClientHealthCheckServiceName)
+	if err != nil {
+		return grpcCommon{}, err
+	}
+
+	oauthConf := conf.Namespace(oa2FieldOAuth2)
```
gregfurman: If no oauth config is provided, should we care about setting all of this?

internal/impl/grpc/common.go
```diff
@@ -0,0 +1,343 @@
...
+		}
+
+		perRPC := oauth.TokenSource{TokenSource: tokenSource}
+		dialOpts = append(dialOpts, grpc.WithPerRPCCredentials(perRPC))
+	}
+
+	if gcc.healthCheckEnabled {
+		serviceConf := fmt.Sprintf(`{"healthCheckConfig": {"serviceName": "%v"}}`, gcc.healthCheckServiceName)
+		dialOpts = append(dialOpts, grpc.WithDefaultServiceConfig(serviceConf))
+	}
```
gregfurman, line 315: Can you clarify this JSON config and why we expect it in this format? Perhaps an inline comment could be useful since this doesn't make it super clear whether this is something we define or whether it's needed by the upstream dep

internal/impl/grpc/common.go
```diff
@@ -0,0 +1,343 @@
...
+		serviceDescriptor, err := gcc.reflectClient.ResolveService(gcc.serviceName)
+		if err != nil {
+			return err
+		}
+
+		if method := serviceDescriptor.FindMethodByName(gcc.methodName); method != nil {
+			gcc.method = method
+		} else {
+			return fmt.Errorf("method: %v not found", gcc.methodName)
+		}
```
gregfurman:
```suggestion
		if gcc.method = serviceDescriptor.FindMethodByName(gcc.methodName); method == nil {
			return fmt.Errorf("method: %v not found", gcc.methodName)
		}
```

internal/impl/grpc/common.go
```diff
@@ -0,0 +1,343 @@
...
+		if err != nil {
+			return err
+		}
+
+	Found:
+		for _, fileDescriptor := range fileDescriptors {
+			for _, service := range fileDescriptor.GetServices() {
+				if service.GetFullyQualifiedName() == gcc.serviceName || service.GetName() == gcc.serviceName {
+					if method := service.FindMethodByName(gcc.methodName); method != nil {
+						gcc.method = method
```
gregfurman, line 361: So we override the `gcc.method` if `protofiles` are provided. Should we document this somewhere and indicate that providing protobuf definitions take precedence over enabling the reflection server?

internal/impl/grpc/common.go
```diff
@@ -0,0 +1,343 @@
...
+						break Found
+					}
+				}
+			}
+		}
+	}
+
+	if gcc.method == nil {
+		return fmt.Errorf("unable to find method: %s in provided proto files", gcc.methodName)
+	}
```
gregfurman: should this not be placed inside the `if len(gcc.protofiles) {...}` block?

internal/impl/grpc/common.go
```diff
@@ -0,0 +1,343 @@
...
+		perRPC := oauth.TokenSource{TokenSource: tokenSource}
+		dialOpts = append(dialOpts, grpc.WithPerRPCCredentials(perRPC))
+	}
+
+	if gcc.healthCheckEnabled {
+		serviceConf := fmt.Sprintf(`{"healthCheckConfig": {"serviceName": "%v"}}`, gcc.healthCheckServiceName)
+		dialOpts = append(dialOpts, grpc.WithDefaultServiceConfig(serviceConf))
+	}
+
+	gcc.conn, err = grpc.NewClient(gcc.address, dialOpts...)
```
gregfurman, line 317: Since `Connect` will re-run if a non-nil `err` is returned, I'm wondering whether it's desirable to re-create the entire `gcc.conn` each time, at least without closing the previous one.

Wdyt?

internal/impl/grpc/common.go
```diff
@@ -0,0 +1,343 @@
...
+		}
+	}
+
+	if len(gcc.protoFiles) != 0 {
+		var parser protoparse.Parser
+
+		fileDescriptors, err := parser.ParseFiles(gcc.protoFiles...)
+		if err != nil {
+			return err
+		}
```
gregfurman, line 354: FYI there's a potential bug with this approach.

Since a `Connect` method will retry (with backoff) until it returns a `nil` error, if `gcc.reflection` is enabled, `gcc.method` AND `gcc.conn` will have been defined.

This means if `parser.ParseFiles` fails with an error, the subsequent `Connect` retry will always return without error due to the [idempotency guard atop the method](https://github.com/jem-davies/bento/blob/49507bc93f8b12a1d49f8f28138262d6cea1079a/internal/impl/grpc/common.go#L244-L246):

```golang
	if gcc.conn != nil && gcc.method != nil {
		return nil
	}
```

Do we ever expect both `protofiles` to be provided and the reflection server to be enabled? Perhaps we should be guarding against this at the config level.

Lmk if this makes sense.

internal/impl/grpc/input_grpc_client.go
```diff
@@ -0,0 +1,384 @@
...
+
+Either the field ` + "`reflection` or `proto_files`" + ` must be supplied, which will provide the protobuf schema Bento will use to marshall the Bento message into protobuf.
+
+`
+
+const grpcClientInputPayloadDescription = `
+For ` + "`rpc_type`" + ` values: unary, client_stream & server_stream, the payload field defines the data that is sent as the request.
+In the instance of the client_stream rpc_type - this is expected to resolve to an array. The array will be sent as a stream, with the elements
+making the individual messages.
+`
```
gregfurman, line 31:
```suggestion
const grpcClientInputPayloadDescription = `
For ` + "`rpc_type`" + ` values: ` + "`unary`" + `, ` + "`client_stream`" + ` & ` + "`server_stream`" + `, the ` + "`payload`" + ` field defines the data that is sent as the request.
In the instance of the ` + "`client_stream`" + ` rpc_type - this is expected to resolve to an array. The array will be sent as a stream, with the elements
making the individual messages.
`
```

internal/impl/grpc/input_grpc_client.go
```diff
@@ -0,0 +1,384 @@
...
+func grpcClientInputSpec() *service.ConfigSpec {
+	return service.NewConfigSpec().
+		Summary("Receives messages from a gRPC server.").
+		Description(grpcClientInputDescription).
+		Fields(grpcCommonFieldSpec()...).
+		Fields(
+			service.NewInterpolatedStringField(grpcClientInputPayload).
+				Description(grpcClientInputPayloadDescription).
+				Optional().
+				Example("TODO"),
```
gregfurman: 🧹 Should probably be removed

```suggestion
```

internal/impl/grpc/input_grpc_client.go
```diff
@@ -0,0 +1,384 @@
...
+			service.NewStringField(grpcClientInputRateLimit).
+				Description("An optional [rate limit](/docs/components/rate_limits/about) to throttle requests by.").
+				Optional(),
+		).
+		Field(service.NewAutoRetryNacksToggleField()).
+		LintRule(
+			`root = match {
+  this.reflection == false && (!this.exists("proto_files") || this.proto_files.length() == 0) => "reflection must be true or proto_files must be populated"
```
gregfurman: Already mentioned above, but what if both are `true`?

internal/impl/grpc/input_grpc_client.go
```diff
@@ -0,0 +1,384 @@
...
+  (this.rpc_type == "unary" || this.rpc_type == "client_stream" || this.rpc_type == "server_stream") && this.payload == null => "payload must be set for rpc_types: unary, client_stream, server_stream",
```
gregfurman, line 51:
```suggestion
  ["unary","client_stream","client_stream"].contains(this.rpc_type) && this.payload == null => "payload must be set for rpc_types: unary, client_stream, server_stream",
```

internal/impl/grpc/input_grpc_client.go
```diff
@@ -0,0 +1,384 @@
...
+		})
+	if err != nil {
+		panic(err)
+	}
+}
+
+//------------------------------------------------------------------------------
+
+type grpcClientInput struct {
+	grpcCommon
```
gregfurman: WDYT of making this an attribute instead of embedding? I got confused for a moment trying to find the `Connect` method of the component 😅

internal/impl/grpc/input_grpc_client.go
```diff
@@ -0,0 +1,384 @@
...
+	responseBatch := service.MessageBatch{}
+
+	for {
+		resProtoMessage, err := serverStream.RecvMsg()
+		if err == io.EOF {
+			break
+		}
+		if err != nil {
+			return nil, nil, err
+		}
```
gregfurman, line 284: super-nit: for-loops without explicit conditions scare me. Feel free to ignore haha

```suggestion
	for resProtoMessage, err := serverStream.RecvMsg(); err != io.EOF; {
		if err != nil {
			return nil, nil, err
		}
```

internal/impl/grpc/input_grpc_client.go
```diff
@@ -0,0 +1,384 @@
...
+	payload, err := gci.payloadExpr.TryString(service.NewMessage([]byte("")))
+	if err != nil {
+		err = fmt.Errorf("payload interpolation error: %w", err)
+		return nil, nil, err
+	}
+
+	err = request.UnmarshalJSON([]byte(payload))
+	if err != nil {
+		return nil, nil, err
+	}
```
gregfurman: Not sure I understand the purpose of this. Why do we need to interpolate on empty bytes and then unmarshal these to the protobuf request?

internal/impl/grpc/input_grpc_client.go
```diff
@@ -0,0 +1,384 @@
...
+}
+
+func (gci *grpcClientInput) serverStreamHandler(ctx context.Context) (service.MessageBatch, service.AckFunc, error) {
+
+	request := dynamic.NewMessage(gci.method.GetInputType())
+
+	if gci.payloadExpr == nil {
+		return nil, nil, fmt.Errorf("payload is required when rpc_type: %v", gci.rpcType)
+	}
+	payload, err := gci.payloadExpr.TryString(service.NewMessage([]byte("")))
```
gregfurman:
```suggestion
	payload, err := gci.payloadExpr.TryBytes(service.NewMessage([]byte("")))
```

internal/impl/grpc/input_grpc_client.go
```diff
@@ -0,0 +1,384 @@
...
+	}
+
+	for _, payload := range payloads {
+		request := dynamic.NewMessage(gci.method.GetInputType())
+		if err := request.UnmarshalJSON(payload); err != nil {
+			return nil, nil, err
+		}
+
+		err = clientStream.SendMsg(request)
+		if err != nil {
```
gregfurman: nit: make consistent with above inlined error check

```suggestion
		
		if err := clientStream.SendMsg(request); err != nil {
```

internal/impl/grpc/input_grpc_client.go
```diff
@@ -0,0 +1,384 @@
...
+		grpcCommon: gcc,
+
+		payloadExpr: payloadExpr,
+		rateLimit:   rateLimit,
+		mgr:         mgr,
+		log:         mgr.Logger(),
+	}, nil
+}
+
+func (gci *grpcClientInput) ReadBatch(ctx context.Context) (service.MessageBatch, service.AckFunc, error) {
```
gregfurman, line 123: Why is this a `BatchInput` when all responses (besides the `serverStreamHandler`) are just returning batches with 1 message?

internal/impl/grpc/input_grpc_client.go
```diff
@@ -0,0 +1,384 @@
...
+		}
+
+	default:
+		return nil, nil, fmt.Errorf("rpc_type: %v not supported", gci.rpcType)
+	}
+}
+
+func (gci *grpcClientInput) Close(ctx context.Context) error {
+	return nil
+}
```
gregfurman, line 183: Should we not be closing the `grpcCommon.conn`, bidirectional channels, etc. here?

internal/impl/grpc/input_grpc_client.go
```diff
@@ -0,0 +1,384 @@
...
+			return service.AutoRetryNacksBatchedToggled(conf, rdr)
+		})
+	if err != nil {
+		panic(err)
+	}
+}
+
+//------------------------------------------------------------------------------
+
+type grpcClientInput struct {
```
gregfurman, line 73: Thoughts on creating a special case for when the data being ingested is already in bento-protobuf form?

https://github.com/warpstreamlabs/bento/blob/07de34111fc9bb1f6ef1d3e30bd56426c5c74e13/internal/message/protobuf.go#L12

gregfurman (changes requested):
Left quite a few comments. Many of them are "stylistic" but I have a couple of concerns.

Otherwise, I'd be really keen to integrate this approach with our [Bento protobuf implementation](https://github.com/warpstreamlabs/bento/blob/07de34111fc9bb1f6ef1d3e30bd56426c5c74e13/internal/message/protobuf.go) 👀 

That way, we could wrap a component to return/ingest our protobuf form and have this `grpc_client` be able to ingest bento data from components not running within the same process 🤩 

```golang
inputProto := service.WrapWithProtobuf(input)
```

internal/impl/grpc/grpc_test_server/test_server/test_server.go
jem-davies: Hmm seems fine as-is.

internal/impl/grpc/common.go
jem-davies: Thinking that it is clear because we have them referenced under `service.NewStringEnumField()` in the field spec anyway. 

(Think most of the rest of the code base doesn't bother to have consts, just has a string literal for the enum options, as the variadic arguments to NewStringEnumField).

internal/impl/grpc/common.go
jem-davies: Updated the lint rules such it won't be possible to set both `proto_files` and `reflection`.

internal/impl/grpc/input_grpc_client.go
jem-davies: Added a case to stop setting both reflection & proto_files.

internal/impl/grpc/input_grpc_client.go
jem-davies: It gives the user the ability to provide a Bloblang expression for `payload` - we need a message to operate on, so we make a blank message. Similar to other inputs that have a interpolatedStringField.

internal/impl/grpc/input_grpc_client.go
jem-davies: gonna leave as-is

internal/impl/grpc/input_grpc_client.go
jem-davies: Added a comment to mention that it impl Connect()

internal/impl/grpc/common.go
jem-davies: Yeah I see - added a guard such that you cannot set both `reflection` & `proto_files` 👍

internal/impl/grpc/grpc_test_server/test_server/test_server.go
jem-davies: also unsure about not wanting them in the production binary as they are for tests, which we have now moved to a /testdata/ directory.

internal/impl/grpc/common.go
jem-davies: Added a defer to close and reset these struct fields if we encounter an error in the Connect func 👍

internal/impl/grpc/input_grpc_client.go
jem-davies: It was just because the serverStreamHandler is more than one message / there could be opportunity to add batching to this in future as well.

internal/impl/grpc/input_grpc_client.go
jem-davies: Yeah it could work - think that it could be a separate github PR / Issue

jem-davies:
@gregfurman ready for another look when you can 🙏

internal/impl/grpc/common.go
```diff
@@ -0,0 +1,361 @@
...
+type oauth2Config struct {
+	enabled        bool
+	clientKey      string
+	clientSecret   string
+	tokenURL       string
+	scopes         []string
+	endpointParams map[string][]string
+}
+
+func oAuth2FieldSpec() *service.ConfigField {
```
gregfurman, line 96: nit(non-blocking): Out of scope for this PR, but it feels like oauth2 config spec should be a common util since there are a couple of components that redefine it (i.e. http client, pulsar, etc.).

internal/impl/grpc/common.go
```diff
@@ -0,0 +1,361 @@
...
+			Advanced(),
+		service.NewAnyMapField(oa2FieldEndpointParams).
+			Description("A list of optional endpoint parameters, values should be arrays of strings.").
+			Advanced().
+			Example(map[string]any{
+				"foo": []string{"meow", "quack"},
+				"bar": []string{"woof"},
+			}).
+			Default(map[string]any{}).
+			Optional(),
```
gregfurman: suggestion(non-blocking): a lint-rule here could be useful e.g

https://github.com/warpstreamlabs/bento/blob/665eb7914ac9d8f196fcbf9bd87d2169bdaebea6/internal/httpclient/auth_oauth2.go#L96-L105

internal/impl/grpc/common.go
```diff
@@ -0,0 +1,361 @@
...
+		if err != nil {
+			return grpcConfig{}, err
+		}
+		ep, err := oauthConf.FieldAnyMap(oa2FieldEndpointParams)
+		if err != nil {
+			return grpcConfig{}, err
+		}
+		endpointParams := map[string][]string{}
+		for k, v := range ep {
+			if endpointParams[k], err = v.FieldStringList(); err != nil {
```
gregfurman, line 228: Wdyt about also accepting a list of comma-separated values in addition to a map-style approach?

i.e

```yaml
endpoint_params: "foo,bar,baz"
```

internal/impl/grpc/common.go
```diff
@@ -0,0 +1,361 @@
...
+	success := false
+	defer func() {
+		if !success {
+			if gcc.conn != nil {
+				gcc.conn.Close()
+			}
+			gcc.conn = nil
+			gcc.method = nil
+			gcc.reflectClient = nil
+		}
```
gregfurman: nit: Can we unnest this?

```golang
if success {
  return
}
// ...
```

internal/impl/grpc/common.go
```diff
@@ -0,0 +1,361 @@
...
+			}
+		}
+		if gcc.method == nil {
+			return fmt.Errorf("unable to find method: %s in provided proto files", gcc.methodName)
+		}
+	}
+
+	gcc.stub = grpcdynamic.NewStub(gcc.conn)
+
+	success = true
```
gregfurman, line 374: praise: Nice 😎

internal/impl/grpc/input_grpc_client.go
gregfurman: Bump on this suggestion. Feel free to resolve this if you don't think it's necessary 🙂

internal/impl/grpc/input_grpc_client.go
```diff
@@ -0,0 +1,403 @@
...
+func (gci *grpcClientInput) startBidi(ctx context.Context) (err error) {
+	gci.bidiChan = make(chan service.MessageBatch)
+	gci.bidiErrChan = make(chan error)
+
+	bidiCtx, bidiCancel := context.WithCancel(ctx)
+	gci.bidiCancel = bidiCancel
+
+	bidi, err := gci.stub.InvokeRpcBidiStream(bidiCtx, gci.method)
+	if err != nil {
+		bidiCancel()
```
gregfurman, line 363: Q: confirming that we want to keep the `bidiChan` open in this case?

internal/impl/grpc/grpc_test_server/test_server/test_server.go
gregfurman: the `testdata` directory is not included in the production binary. See `go help package`:

> [...]
> Directory and file names that begin with "." or "_" are ignored
> by the go tool, as are directories named "testdata".

internal/impl/grpc/input_grpc_client_test.go
```diff
@@ -0,0 +1,474 @@
...
+				[]byte(`{"message":"Hello Alice"}`),
+			},
+		},
+	}
+
+	for name, test := range tests {
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+
+			testServer := test_server.StartGRPCServer(t, test.grpcServerOpts...)
```
gregfurman, line 163: nit: a comment mentioning that this `StartGRPCServer` fixture randomly allocates a port could be useful -- since `localhost:0` is used under-the-hood

gregfurman (approved):
LGTM! Perhaps a more complete example in the docs or a cookbook here could be useful just to get the full picture.

Otherwise, nothing blocking. Great stuff!

internal/impl/grpc/common.go
jem-davies: https://github.com/warpstreamlabs/bento/issues/898

internal/impl/grpc/input_grpc_client.go
jem-davies: moved the creation of the channel until after this call that could error 👍

Issue #770, opened 2026-03-17
Potential nil pointer panic and race conditions in `sql_insert` and `sql_raw` outputs


erkattak:
There are several race conditions in the `sql_insert` and `sql_raw` batch output implementations that can cause a nil pointer dereference panic at runtime. We hit this in production during a database cluster upgrade (when connections were being dropped), and after reviewing the source we believe the underlying issues are still present in the latest version (v1.15.2).

## Observed panic (v1.12.1)

```internal/impl/sql/output_sql_insert.go
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x74c315]

goroutine 289 [running]:
database/sql.(*DB).conn(0x0, {0x793d020, 0xc003112000}, 0x1)
        database/sql/sql.go:1317 +0x55
database/sql.(*DB).exec(0x0, ...)
        database/sql/sql.go:1689 +0x54
...
github.com/warpstreamlabs/bento/internal/impl/sql.(*sqlInsertOutput).WriteBatch(...)
        internal/impl/sql/output_sql_insert.go:296
```

The panic originates from `s.db` being nil at the point of the `ExecContext` call inside `writeBatch`, despite having passed a nil-guard at the top of the same function.

### 1. `sql_insert` — nil-DB race between `writeBatch` and `WriteBatch` / `Close`

`writeBatch` acquires only a **read lock** (`RLock`) for its entire duration and performs a nil check on `s.db` near the top:

```internal/impl/sql/output_sql_insert.go#L242-L248
func (s *sqlInsertOutput) writeBatch(ctx context.Context, batch service.MessageBatch) error {
	s.dbMut.RLock()
	defer s.dbMut.RUnlock()

	if s.db == nil {
		return service.ErrNotConnected
	}
```

However, two other code paths can set `s.db = nil` under a **write lock** while `writeBatch` is already past its nil check and still holding the read lock:

- **`WriteBatch` on bad connection:**
```internal/impl/sql/output_sql_insert.go#L231-L239
func (s *sqlInsertOutput) WriteBatch(ctx context.Context, batch service.MessageBatch) error {
	err := s.writeBatch(ctx, batch)
	...
	if errors.Is(err, driver.ErrBadConn) {
		s.dbMut.Lock()
		s.db = nil
		s.dbMut.Unlock()
		return service.ErrNotConnected
	}
```

- **The shutdown goroutine in `Connect`:**
[...]

PR #771, opened 2026-03-17
fix(sql): prevent nil pointer panic in `sql_insert` and `sql_raw` outputs during connection drops

erkattak:
Addresses #770 

Both outputs protect `s.db` with a `sync.RWMutex`, but `writeBatch` held only an `RLock` for its entire duration. Because `RWMutex` allows concurrent readers, a writer could legally acquire the write lock and zero `s.db` — via the bad-connection reset in `WriteBatch` or the shutdown goroutine in `Connect` — between the nil check and the subsequent use of `s.db`. This TOCTOU window caused a nil pointer panic under load when connections were dropped.

**Fixes:**
- In both outputs, `writeBatch` now captures `s.db` into a local variable immediately after acquiring the `RLock`. The check and all subsequent uses refer to the same pointer value, closing the TOCTOU window with no lock-mode changes.
- `sql_raw`'s shutdown goroutine now sets `s.db = nil` after `Close()`, consistent with `sql_insert`. Previously a closed but non-nil `*sql.DB` would produce a generic driver error that bypassed the `driver.ErrBadConn` reconnect path, leaving the output permanently broken.
- Tests for both outputs are added to their respective `_test.go` files, covering the bad-connection and shutdown write paths with a barrier-synchronised burst of 50 goroutines to maximise the TOCTOU window. Because this is a logical race rather than an unsynchronised memory access, `go test -race` does not flag it; the tests reproduce the failure mode directly.

### DIFF
--- a/internal/impl/sql/output_sql_insert.go
+++ b/internal/impl/sql/output_sql_insert.go
@@ -273,14 +273,19 @@ func (s *sqlInsertOutput) writeBatch(ctx context.Context, batch service.MessageB
 		if tx, err = s.db.Begin(); err != nil {
 			return err
 		}
+		defer func() {
+			_ = tx.Rollback()
+		}()
 		sqlStr, _, err := insertBuilder.ToSql()
 		if err != nil {
 			return err
 		}
 		if stmt, err = tx.Prepare(sqlStr); err != nil {
-			_ = tx.Rollback()
 			return err
 		}
+		defer func() {
+			_ = stmt.Close()
+		}()
 	}
 
 	for i := range batch {
@@ -305,7 +310,6 @@ func (s *sqlInsertOutput) writeBatch(ctx context.Context, batch service.MessageB
 		if tx == nil {
 			insertBuilder = insertBuilder.Values(args...)
 		} else if _, err := stmt.Exec(args...); err != nil {
-			_ = tx.Rollback()
 			return err
 		}
 	}
--- a/internal/impl/sql/cache_sql.go
+++ b/internal/impl/sql/cache_sql.go
@@ -5,6 +5,7 @@ import (
 	"database/sql"
 	"errors"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/Masterminds/squirrel"
@@ -95,6 +96,7 @@ type sqlCache struct {
 	driver string
 	dsn    string
 	db     *sql.DB
+	dbMut  sync.RWMutex
 
 	keyColumn string
 
@@ -182,6 +184,8 @@ func newSQLCacheFromConfig(conf *service.ParsedConfig, mgr *service.Resources) (
 		}
 	}
 
+	s.dbMut.Lock()
+	defer s.dbMut.Unlock()
 	if s.db, err = sqlOpenWithReworks(context.Background(), s.logger, s.driver, s.dsn, connSettings); err != nil {
 		return nil, err
 	}
@@ -189,13 +193,18 @@ func newSQLCacheFromConfig(conf *service.ParsedConfig, mgr *service.Resources) (
 
 	go func() {
 		<-s.shutSig.HardStopChan()
+		s.dbMut.Lock()
 		_ = s.db.Close()
+		s.db = nil
+		s.dbMut.Unlock()
 		s.shutSig.TriggerHasStopped()
 	}()
 	return s, nil
 }
 
 func (s *sqlCache) Get(ctx context.Context, key string) (value []byte, err error) {
+	s.dbMut.RLock()
+	defer s.dbMut.RUnlock()
 	err = s.selectBuilder.
 		Where(squirrel.Eq{s.keyColumn: key}).
 		RunWith(s.db).QueryRowContext(ctx).
@@ -207,11 +216,15 @@ func (s *sqlCache) Get(ctx context.Context, key string) (value []byte, err error
 }
 
 func (s *sqlCache) Set(ctx context.Context, key string, value []byte, ttl *time.Duration) error {
+	s.dbMut.RLock()
+	defer s.dbMut.RUnlock()
 	_, err := s.upsertBuilder.Values(key, value).RunWith(s.db).ExecContext(ctx)
 	return err
 }
 
 func (s *sqlCache) Add(ctx context.Context, key string, value []byte, ttl *time.Duration) error {
+	s.dbMut.RLock()
+	defer s.dbMut.RUnlock()
 	_, err := s.insertBuilder.Values(key, value).RunWith(s.db).ExecContext(ctx)
 	if err != nil {
 		// This is difficult, ideally we need to translate any error that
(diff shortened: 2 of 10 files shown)


internal/impl/sql/output_sql_insert.go
```diff
@@ -254,7 +254,8 @@ func (s *sqlInsertOutput) writeBatch(ctx context.Context, batch service.MessageB
 	s.dbMut.RLock()
 	defer s.dbMut.RUnlock()
 
-	if s.db == nil {
+	db := s.db
+	if db == nil {
```
gregfurman: Perhaps I'm misunderstanding but considering this happens within an `RLock`, why do we need to assign this DB to a seperate variable?

erkattak:
Yes, if you checkout only the two changed sql output files from main, my tests induce the panic:

```
git checkout main -- internal/impl/sql/output_sql_insert.go
git checkout main -- internal/impl/sql/output_sql_raw.go
go test -race ./internal/impl/sql/...
```

internal/impl/sql/output_sql_insert.go
erkattak: The local `db := s.db` assignment is a defensive copy made while holding the read lock.

Basically, the rest of the function works against a stable, local reference that can't be changed out from under it by another goroutine. Even if `s.db` were later set to `nil`. The nil check and all subsequent uses of `db` are guaranteed to refer to the same object.

erkattak: The read lock doesn't prevent another write lock from doing something to the underlying `s.db`

erkattak: Hmm... I think you're right about questioning this. I've just written things this way a bunch, I guess

internal/impl/sql/output_sql_insert.go
```diff
@@ -292,14 +273,19 @@ func (s *sqlInsertOutput) writeBatch(ctx context.Context, batch service.MessageB
 		if tx, err = s.db.Begin(); err != nil {
 			return err
 		}
+		defer func() {
+			_ = tx.Rollback()
+		}()
```
jem-davies, line 278: Seems strange to defer a Rollback - shouldn't we only rollback if there has been an error like before?

jem-davies:
> Yes, if you checkout only the two changed sql output files from main, my tests induce the panic:
> 
> ```
> git checkout main -- internal/impl/sql/output_sql_insert.go
> git checkout main -- internal/impl/sql/output_sql_raw.go
> go test -race ./internal/impl/sql/...
> ```

Thanks I have done that and seen that you get a nil pointer panic with the existing impl code 

TestSQLRawOutputWriteBatchWithNilDB

```go
func TestSQLRawOutputWriteBatchWithNilDB(t *testing.T) {
	output := newSQLRawOutput(
		service.MockResources().Logger(),
		"sqlite",
		"",
		"INSERT INTO things (foo) VALUES (?)",
		nil,
		nil,
		defaultConnSettings(),
		aws.Config{},
	)

	batch := service.MessageBatch{
		service.NewMessage([]byte(`hello`)),
	}

	err := output.WriteBatch(context.Background(), batch)
	require.ErrorIs(t, err, service.ErrNotConnected)
}
```

It shouldn't be possible in the execution model of Bento that .WriteBatch() could be called without first successfully calling .Connect(). So I don't think that would emulate the panic you had on #770 (because also on 770 it's the sqlInsertOutput that panic'd which does have a nil pointer check) 

But yeah we should add a `if s.db == nil` check to SQLRawOutput 👍

internal/impl/sql/output_sql_insert.go
erkattak: Deferring a rollback is probably the easiest way to ensure that things roll back if an error is encountered. If the tx was already committed, you'd receive back a `sql.ErrTxDone`

erkattak:
The issue arose for us in a production environment because we were doing a database upgrade (YugabyteDB Cluster, but treated as Postgres). As part of the upgrade, connections are closed by the server. So we ended up receiving this panic:

```
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x74c315]

goroutine 289 [running]:
internal/sync.(*Mutex).Lock(...)
	internal/sync/mutex.go:63
sync.(*Mutex).Lock(...)
	sync/mutex.go:46
database/sql.(*DB).conn(0x0, {0x793d020, 0xc003112000}, 0x1)
	database/sql/sql.go:1317 +0x55
database/sql.(*DB).exec(0x0, {0x793d020, 0xc003112000}, {0xc002268a00, 0x97}, {0xc002da2280, 0x7, 0x8}, 0x97?)
	database/sql/sql.go:1689 +0x54
database/sql.(*DB).ExecContext.func1(0x1?)
	database/sql/sql.go:1672 +0x4f
database/sql.(*DB).retry(0x5de4e60?, 0xc00223b9f8)
	database/sql/sql.go:1576 +0x42
database/sql.(*DB).ExecContext(0x6b08d20?, {0x793d020?, 0xc003112000?}, {0xc002268a00?, 0x7?}, {0xc002da2280?, 0x6b08d20?, 0xc002da2280?})
	database/sql/sql.go:1671 +0xc8
github.com/Masterminds/squirrel.ExecContextWith({0x793d020, 0xc003112000}, {0x7f46d02452f0, 0xc001444f30}, {0x78d10c0?, 0xc0027f6f00?})
	github.com/Masterminds/squirrel@v1.5.4/squirrel_ctx.go:77 +0x73
github.com/Masterminds/squirrel.(*insertData).ExecContext(0xc0027f6f00?, {0x793d020?, 0xc003112000?})
	github.com/Masterminds/squirrel@v1.5.4/insert_ctx.go:20 +0xd0
github.com/Masterminds/squirrel.InsertBuilder.ExecContext({{0x79737d8, 0xc0029bb8f0}}, {0x793d020, 0xc003112000})
	github.com/Masterminds/squirrel@v1.5.4/insert_ctx.go:51 +0x132
github.com/warpstreamlabs/bento/internal/impl/sql.(*sqlInsertOutput).WriteBatch(0xc002dc3188, {0x793d020, 0xc003112000}, {0xc002b00010, 0x1, 0x1})
	github.com/warpstreamlabs/bento@v1.12.1/internal/impl/sql/output_sql_insert.go:296 +0x7f4
github.com/warpstreamlabs/bento/public/service.(*airGapBatchWriter).WriteBatch(0xc002453270, {0x793d020, 0xc003112000}, {0xc002cfae48, 0x1, 0x1?})
	github.com/warpstreamlabs/bento@v1.12.1/public/service/output.go:117 +0x183
github.com/warpstreamlabs/bento/internal/component/output.(*AsyncWriter).latencyMeasuringWrite(0xc002f61400, {0x793d020, 0xc003112000}, {0xc002cfae48, 0x1, 0x1})
	github.com/warpstreamlabs/bento@v1.12.1/internal/component/output/async_writer.go:81 +0x85
github.com/warpstreamlabs/bento/internal/component/output.(*AsyncWriter).loop.func4()
	github.com/warpstreamlabs/bento@v1.12.1/internal/component/output/async_writer.go:209 +0x376
created by github.com/warpstreamlabs/bento/internal/component/output.(*AsyncWriter).loop in goroutine 244
	github.com/warpstreamlabs/bento@v1.12.1/internal/component/output/async_writer.go:247 +0x8f6
```

This is what the test reproduces

erkattak:
I'm closing this and will reopen. The scope changed quite a bit when I looked through the rest of the impl/sql code and found more race conditions, logic bugs, and resource leaks

Issue #38, opened 2024-06-09
support different encodings in `parquet` encode processor


ryanworl:
Relevant docs: https://warpstreamlabs.github.io/bento/docs/components/processors/parquet_encode/

Today Bento only supports PLAIN and DELTA_LENGTH_BYTE_ARRAY encodings, and you can only use a single encoding for all data types.

We should instead support configuring a different encoding for each field.

For example, in a use case where you are transforming a Kafka topic into a set of Parquet files, you would want to use RLE_DICTIONARY for the topic names and partitions. You'd probably also want to use DELTA_BINARY_PACKED for the offsets and record timestamps.

PR #773, opened 2026-03-18
feat(parquet): add RLE_DICTIONARY support to default_encoding

georgeTobia:
## Summary

- Adds `RLE_DICTIONARY` as a supported value for the `default_encoding` field in the `parquet_encode` processor
- Maps to the `dict` struct tag in parquet-go, enabling dictionary encoding on all leaf column types
- Applies dictionary encoding to all leaf types (unlike `PLAIN` which only targets string types) while skipping container types (MAP, LIST, STRUCT)

## Changes

- `processor_encode.go`: Added `RLE_DICTIONARY` to the enum and a switch case mapping it to the `dict` struct tag
- `schema.go`: Added `isLeafType` helper and logic to apply the `dict` tag to all leaf types
- `processor_encode_test.go`: Added encoding verification test (with nested STRUCT), round-trip test, and MAP/LIST round-trip test

## Test plan

- [x] `TestParquetEncodeDefaultEncodingRLEDictionary` — verifies all leaf fields (including nested) have `RLE_DICTIONARY` encoding
- [x] `TestParquetEncodeDecodeRoundTripRLEDictionaryEncoding` — full encode/decode round-trip with all supported types
- [x] `TestParquetEncodeDecodeRoundTripRLEDictionaryMapList` — round-trip with MAP and LIST container types
- [x] All existing tests continue to pass

Partial fix for #38


gregfurman:
Think you need to run a `make docs` again and commit that since I see this branch has conflicts 👀 

Otherwise, will review this week! Apologies for the delay.

gregfurman (approved):
LGTM. I manually updated the docs since they were outdated.

PR #774, opened 2026-03-18
amqp1: expose AMQP  property as input metadata

### LINKED ISSUE #754: AMQP 1.0 input: expose message `to` property as metadata
**Problem**
When consuming messages via `amqp_1`, Bento exposes some AMQP metadata such as content type, content encoding, creation time, optional header fields, and string annotations. However, it does not expose the AMQP message `to` property.

In our Solace setup, we are able to access the original topic information from `message.address` when consuming the same messages with Python Proton. Exposing the corresponding value as Bento metadata would make it available for routing and downstream processing there as well.

**Why this matters**
This makes it hard to route records dynamically based on the original topic/address when using Bento as an AMQP 1.0 consumer.

**Source code observations**

* [`internal/impl/amqp1/input.go`](https://github.com/warpstreamlabs/bento/blob/90ab505706184b642a0e77411d0afb67087f953d/internal/impl/amqp1/input.go#L252-L256) already reads `amqpMsg.Properties`
* metadata is currently populated from `ContentType`, `ContentEncoding`, and `CreationTime`
* Bento uses `github.com/Azure/go-amqp`, and the underlying AMQP message properties include [`To *string`](https://pkg.go.dev/github.com/Azure/go-amqp#MessageProperties). So the field appears available in the client library but not currently mapped into Bento metadata.

**Proposed change**
Add the AMQP `to` property as metadata, for example `amqp_to`.

Possible implementation in [`internal/impl/amqp1/input.go`](https://github.com/warpstreamlabs/bento/blob/90ab505706184b642a0e77411d0afb67087f953d/internal/impl/amqp1/input.go#L252-L256):

```go
if amqpMsg.Properties != nil {
    amqpSetMetadata(part, "amqp_content_type", amqpMsg.Properties.ContentType)
    amqpSetMetadata(part, "amqp_content_encoding", amqpMsg.Properties.ContentEncoding)
    amqpSetMetadata(part, "amqp_creation_time", amqpMsg.Properties.CreationTime)
    amqpSetMetadata(part, "amqp_to", amqpMsg.Properties.To)
}
```

Also update the input docs and add a corresponding test.

If this sounds reasonable, I’d be happy to open a PR for it.

msamadony:
This exposes the AMQP message `to` property as metadata in the `amqp_1` input as `amqp_to`.

The change adds the mapping in the input, updates the docs, and extends the existing integration test.

Fixes #754

### DIFF
--- a/internal/impl/amqp1/testdata/docker-compose.yaml
+++ b/internal/impl/amqp1/testdata/docker-compose.yaml
@@ -1,0 +1,72 @@
+# Adapted docker-compose.yaml from https://github.com/Azure/azure-service-bus-emulator-installer/blob/main/Docker-Compose-Template/docker-compose-default.yml
+# and Config.json pattern from https://jeremydmiller.com/2026/02/19/using-the-azure-service-bus-emulator-for-local-wolverine-development/
+
+# Ran:
+#    docker compose -f internal/impl/amqp1/testdata/docker-compose.yaml up -d
+#    TEST_SB_URL="amqp://localhost:5672" TEST_SB_SOURCE_ADDRESS="/test" go test -v -timeout 120s -run TestIntegrationAzureServiceBus ./internal/impl/amqp1
+#    docker compose -f internal/impl/amqp1/testdata/docker-compose.yaml down
+
+name: microsoft-azure-servicebus-emulator
+
+configs:
+  servicebus_config:
+    content: |
+      {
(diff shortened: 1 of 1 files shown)


msamadony:
Thanks a lot for taking the time to test it and create the docker-compose file. I added it to a `testdata` folder and committed it.

PR #775, opened 2026-03-18
fix(sql): nil pointer panics, race conditions, and resource leaks in SQL outputs

### LINKED ISSUE #770: Potential nil pointer panic and race conditions in `sql_insert` and `sql_raw` outputs

erkattak:
This fixes a collection of bugs in the SQL output, processor, and cache components. Most are latent race conditions or resource leaks that show up under load or during shutdown.

We experienced a bug that would be fixed by these changes in production during a database upgrade.

Fixes #770 

### Changes

- `output_sql_insert`, `processor_sql_insert`: Replaced scattered `tx.Rollback()` calls with `defer tx.Rollback()` so rollback always fires on early return. Added `defer stmt.Close()` to ensure prepared statements are released.
- `output_sql_raw`, `processor_sql_raw`, `processor_sql_select`, `input_sql_select`: Set `db = nil` after `db.Close()` in shutdown goroutines. Added a nil-guard in `writeBatch` - previously a write after shutdown would panic rather than return `ErrNotConnected`.
- `processor_sql_raw`, `processor_sql_select`: Added `rows.Close()` after iterating result sets. Without this, DB connections were held open longer than needed.
- `cache_sql`: Added a `sync.RWMutex` around the `db` field. The shutdown goroutine and all cache methods (`Get`, `Set`, `Add`, `Delete`) were accessing it without synchronization.
- `conn_fields`: Fixed an inverted condition in `reworkDSN` for ClickHouse legacy TCP DSNs - the `password == ""` branch was backwards, so username-only connections got the wrong user info.

### Explanation

The shutdown race could cause a nil dereference mid-write. Prepared statements leaked on partial failures. Rows from raw/select processors were never explicitly closed, holding connections open. The DSN bug silently misconfigured ClickHouse connections that had a username but no password.

Tests cover the `reworkDSN` logic (including the fixed inversion) and nil-connection behavior for `sql_insert` and `sql_raw` outputs.

### DIFF
--- a/internal/impl/sql/cache_sql.go
+++ b/internal/impl/sql/cache_sql.go
@@ -194,8 +194,10 @@ func newSQLCacheFromConfig(conf *service.ParsedConfig, mgr *service.Resources) (
 	go func() {
 		<-s.shutSig.HardStopChan()
 		s.dbMut.Lock()
-		_ = s.db.Close()
-		s.db = nil
+		if s.db != nil {
+			_ = s.db.Close()
+			s.db = nil
+		}
 		s.dbMut.Unlock()
 		s.shutSig.TriggerHasStopped()
 	}()
--- a/internal/impl/sql/cache_sql.go
+++ b/internal/impl/sql/cache_sql.go
@@ -5,6 +5,7 @@ import (
 	"database/sql"
 	"errors"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/Masterminds/squirrel"
@@ -95,6 +96,7 @@ type sqlCache struct {
 	driver string
 	dsn    string
 	db     *sql.DB
+	dbMut  sync.RWMutex
 
 	keyColumn string
 
@@ -182,6 +184,8 @@ func newSQLCacheFromConfig(conf *service.ParsedConfig, mgr *service.Resources) (
 		}
 	}
 
+	s.dbMut.Lock()
+	defer s.dbMut.Unlock()
 	if s.db, err = sqlOpenWithReworks(context.Background(), s.logger, s.driver, s.dsn, connSettings); err != nil {
 		return nil, err
 	}
@@ -189,13 +193,18 @@ func newSQLCacheFromConfig(conf *service.ParsedConfig, mgr *service.Resources) (
 
 	go func() {
 		<-s.shutSig.HardStopChan()
+		s.dbMut.Lock()
 		_ = s.db.Close()
+		s.db = nil
+		s.dbMut.Unlock()
 		s.shutSig.TriggerHasStopped()
 	}()
 	return s, nil
 }
 
 func (s *sqlCache) Get(ctx context.Context, key string) (value []byte, err error) {
+	s.dbMut.RLock()
+	defer s.dbMut.RUnlock()
 	err = s.selectBuilder.
 		Where(squirrel.Eq{s.keyColumn: key}).
 		RunWith(s.db).QueryRowContext(ctx).
@@ -207,11 +216,15 @@ func (s *sqlCache) Get(ctx context.Context, key string) (value []byte, err error
 }
 
 func (s *sqlCache) Set(ctx context.Context, key string, value []byte, ttl *time.Duration) error {
+	s.dbMut.RLock()
+	defer s.dbMut.RUnlock()
 	_, err := s.upsertBuilder.Values(key, value).RunWith(s.db).ExecContext(ctx)
 	return err
 }
 
 func (s *sqlCache) Add(ctx context.Context, key string, value []byte, ttl *time.Duration) error {
+	s.dbMut.RLock()
+	defer s.dbMut.RUnlock()
 	_, err := s.insertBuilder.Values(key, value).RunWith(s.db).ExecContext(ctx)
 	if err != nil {
 		// This is difficult, ideally we need to translate any error that
@@ -225,6 +238,8 @@ func (s *sqlCache) Add(ctx context.Context, key string, value []byte, ttl *time.
 }
 
 func (s *sqlCache) Delete(ctx context.Context, key string) error {
+	s.dbMut.RLock()
+	defer s.dbMut.RUnlock()
 	_, err := s.deleteBuilder.Where(squirrel.Eq{s.keyColumn: key}).RunWith(s.db).ExecContext(ctx)
 	if err != nil && errors.Is(err, sql.ErrNoRows) {
 		err = service.ErrKeyNotFound
--- a/internal/impl/sql/conn_fields.go
+++ b/internal/impl/sql/conn_fields.go
@@ -54,7 +54,6 @@ The ` + "[`duckdb`](https://github.com/duckdb/duckdb-go)" + ` driver requires cg
 	Example("db_file.duckdb?threads=4&access_mode=READ_ONLY")
 
 func connFields() []*service.ConfigField {
-
 	connFields := []*service.ConfigField{
 		service.NewStringListField("init_files").
 			Description(`
@@ -164,7 +163,6 @@ CREATE TABLE IF NOT EXISTS some_table (
 	connFields = append(connFields, config.SessionFields()...)
 
 	return connFields
-
 }
 
 func rawQueryField() *service.ConfigField {
@@ -222,7 +220,6 @@ func (c *connSettings) apply(ctx context.Context, db *sql.DB, log *service.Logge
 				log.Debug("Successfully ran init_statement")
 			}
 		}
-
 	})
 }
 
@@ -348,7 +345,7 @@ func reworkDSN(driver, dsn string) (string, error) {
 		uq := u.Query()
 		u.Path = uq.Get("database")
 		if username, password := uq.Get("username"), uq.Get("password"); username != "" {
-			if password != "" {
+			if password == "" {
 				u.User = url.User(username)
 			} else {
 				u.User = url.UserPassword(username, password)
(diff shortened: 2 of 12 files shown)


internal/impl/sql/conn_fields.go
```diff
@@ -349,7 +345,7 @@ func reworkDSN(driver, dsn string) (string, error) {
 		uq := u.Query()
 		u.Path = uq.Get("database")
 		if username, password := uq.Get("username"), uq.Get("password"); username != "" {
-			if password != "" {
+			if password == "" {
```
gregfurman, line 348: Now that's an embarassing bug 😅

internal/impl/sql/cache_sql.go
```diff
@@ -182,20 +184,27 @@ func newSQLCacheFromConfig(conf *service.ParsedConfig, mgr *service.Resources) (
...
 		return nil, err
 	}
 	connSettings.apply(context.Background(), s.db, s.logger)
 
 	go func() {
 		<-s.shutSig.HardStopChan()
+		s.dbMut.Lock()
 		_ = s.db.Close()
+		s.db = nil
+		s.dbMut.Unlock()
```
gregfurman, line 201: Q: should we be checking if `s.db` is already `nil` here before proceeding with the close?

internal/impl/sql/processor_sql_raw.go
```diff
@@ -241,6 +203,7 @@ func newSQLRawProcessor(
 
 		s.dbMut.Lock()
 		_ = s.db.Close()
+		s.db = nil
```
gregfurman: q: same as above -- should we be doing a `nil` check here?

internal/impl/sql/cache_sql.go
erkattak: Yep, probably. I'll adjust

erkattak:
I think I've addressed feedback

gregfurman:
@erkattak OK great! I'll give this a look over the weekend or on Monday. 

I'm going to do a full SQL integration test run with these changes in the meanwhile 👍

internal/impl/sql/cache_sql.go
```diff
@@ -182,20 +184,29 @@ func newSQLCacheFromConfig(conf *service.ParsedConfig, mgr *service.Resources) (
...
+			s.db = nil
+		}
+		s.dbMut.Unlock()
 		s.shutSig.TriggerHasStopped()
 	}()
 	return s, nil
 }
 
 func (s *sqlCache) Get(ctx context.Context, key string) (value []byte, err error) {
+	s.dbMut.RLock()
```
gregfurman, line 208: Do you know what happens if the `s.db` we pass in is actually `nil` in these calls? Wonder if we should be checking that

internal/impl/sql/output_sql_insert.go
```diff
@@ -292,14 +273,19 @@ func (s *sqlInsertOutput) writeBatch(ctx context.Context, batch service.MessageB
 		if tx, err = s.db.Begin(); err != nil {
 			return err
 		}
+		defer func() {
+			_ = tx.Rollback()
```
gregfurman, line 277: Why would we want to always be rolling back? Surely we want to only do this on error?

internal/impl/sql/output_sql_insert_test.go
```diff
@@ -28,3 +36,115 @@ args_mapping: 'root = [ this.id ]'
 	require.NoError(t, err)
 	require.NoError(t, insertOutput.Close(context.Background()))
 }
+
+func TestSQLInsertOutputWriteBatchWithNilDB(t *testing.T) {
```
gregfurman, line 40: Wdyt of a test where we run `Connect()`, perform these concurrent writes, and then at some point trigger a `Close()` of the output (while write operations are busy executing / still scheduled to execute) which should close the DB and set the attribute to nil.

Then we can get a more e2e illustration of this new behaviour/test

gregfurman: Also, think if we could replicate the original scenario you were encountering into an integration test that'd be even better than these unit tests i.e

https://github.com/warpstreamlabs/bento/blob/d3d58893162d224d8f105cc91f83f52ad827499a/internal/impl/sql/integration_test.go#L1551

internal/impl/sql/output_sql_insert.go
erkattak: Calling `tx.Rollback()` doesn't always rollback. The transaction context keeps the state and if it's already been committed, this is essentially a noop that returns `sql.ErrTxDone`.

This is the idiomatic use of sql transactions in Go to ensure database connections are released and changes are undone if a function exits unexpectedly due to an error or panic

gregfurman: That's fair. Gave a read on this in https://go.dev/doc/database/execute-transactions and it seems to be idiomatic to both commit and defer a rollback call.

jem-davies:
Closing in favour of : https://github.com/warpstreamlabs/bento/pull/855

PR #783, opened 2026-03-20
fix: reject output conditions with unknown matcher keys

### LINKED ISSUE #382: Unit test output_batches will pass tests with random words as "matcher" types
Due to a typo, I tried to use `bloglang` as a matcher (instead of `bloblang`:

```
    output_batches:
      - - bloglang: 'false'
```

When running the test, I got:
`Test 'config.yaml' succeeded`.

Should bento test fail when a non-existing matcher is used? It definitely would give some "compile-time" safety to the tool.

saschabuehrle:
## Bug
Fixes #382. A typo in an  matcher key can currently pass silently. Unknown matcher keys are dropped during parsing, leaving an empty condition map that acts as a no-op and lets the test pass.

## Fix
- Treat empty parsed output condition blocks as invalid and return a parser error.
- Parse  only when it actually contains keys, so it no longer masks empty/unknown matcher blocks.
- Added regression coverage for a condition block that only contains an unknown matcher.

## Validation
- 

This keeps valid test cases unchanged while failing fast on matcher typos.

Greetings, saschabuehrle

### DIFF
--- a/internal/config/test/output.go
+++ b/internal/config/test/output.go
@@ -90,6 +90,7 @@ func (c OutputConditionsMap) CheckAll(fs fs.FS, dir string, part *message.Part) 
 
 func OutputConditionsFromParsed(pConf *docs.ParsedConfig) (m OutputConditionsMap, err error) {
 	m = OutputConditionsMap{}
+	line, _ := pConf.Line()
 	if pConf.Contains(fieldOutputBloblang) {
 		var tmpStr string
 		if tmpStr, err = pConf.FieldString(fieldOutputBloblang); err != nil {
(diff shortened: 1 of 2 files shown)


internal/config/test/output.go
```diff
@@ -119,11 +120,11 @@ func OutputConditionsFromParsed(pConf *docs.ParsedConfig) (m OutputConditionsMap
...
 
-	if pConf.Contains(fieldOutputMetadataEquals) {
-		var tmpMap map[string]*docs.ParsedConfig
-		if tmpMap, err = pConf.FieldAnyMap(fieldOutputMetadataEquals); err != nil {
-			return
-		}
+	var tmpMap map[string]*docs.ParsedConfig
+	if tmpMap, err = pConf.FieldAnyMap(fieldOutputMetadataEquals); err != nil {
+		return
+	}
```
gregfurman, line 126: This seems reasonable to me. Can you provide an example of failing (and passing) YAML? Just want to make sure I am understanding the scope of this change correctly 🙂

jem-davies: Appears that the referenced issue no longer is valid: https://github.com/warpstreamlabs/bento/issues/382#issuecomment-4199354356

gregfurman:
👋 Is this still an issue?

PR #794, opened 2026-03-30
update(go): minor bump to 1.26

gregfurman:
Bump to new minor Go version largely to address all these issues in the `govulncheck` report from the core library (see [failing run](https://github.com/warpstreamlabs/bento/actions/runs/23735490792/job/69138964737#step:4:10)).

### DIFF
--- a/.github/workflows/cross_build.yml
+++ b/.github/workflows/cross_build.yml
@@ -11,7 +11,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        go-version: [1.25.x]
+        go-version: [1.26.x]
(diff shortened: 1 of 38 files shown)


jem-davies:
Remember we will need to update the images in ./resources/docker as well

jem-davies:
https://github.com/jem-davies/bento/actions/runs/24783464817 - Checking if all vulns are fixed with this update to Go...

jem-davies:
@gregfurman - I have rebased to make my commits easier to grok ... There is still an issue re. govulncheck: https://github.com/jem-davies/bento/actions/runs/24785600362/job/72528580434 - but think we should get that on a follow up

PR #802, opened 2026-04-06
feat(aws): add DynamoDB Streams input component

danielavelez12:
## Summary

Adds a new `aws_dynamodb_streams` input that natively consumes change data capture (CDC) events from DynamoDB Streams.

- **New input:** `aws_dynamodb_streams` — consumes CDC events with balanced shard assignment across multiple consumers
- **Checkpointing:** DynamoDB-based checkpoint table (auto-created if configured) for at-least-once delivery
- **Shard balancing:** Discovers shards, claims unclaimed ones, steals from overloaded clients — modeled after the existing Kinesis input
- **Full type mapping:** Converts all DynamoDB attribute types (S, N, B, BOOL, NULL, SS, NS, BS, L, M) to JSON
- **Message metadata:** Sets `dynamodb_table`, `dynamodb_stream_arn`, `dynamodb_shard_id`, `dynamodb_event_id`, `dynamodb_event_name`, `dynamodb_sequence_number`
- **Config fields:** `table` or `stream_arn`, `checkpoint`, `start_from_oldest`, `batching`, plus standard AWS session fields

### DIFF
--- a/internal/impl/aws/input_dynamodb_streams_checkpointer.go
+++ b/internal/impl/aws/input_dynamodb_streams_checkpointer.go
@@ -1,0 +1,355 @@
+package aws
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"time"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/service/dynamodb"
+	"github.com/aws/aws-sdk-go-v2/service/dynamodb/types"
+
+	"github.com/warpstreamlabs/bento/public/service"
+)
+
+type ddbsCheckpointConfig struct {
+	Table              string
+	Create             bool
+	BillingMode        string
+	ReadCapacityUnits  int64
+	WriteCapacityUnits int64
+}
+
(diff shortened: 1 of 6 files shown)


internal/impl/aws/input_dynamodb_streams_checkpointer.go
danielavelez12, line 1: Not sure if it's better to reuse the Kinesis checkpointer and add a Dynamo-DB specific `Complete` method (as well as add TTL cleanup)

gregfurman: Think I'd prefer this. The overlap between these two components is very high so I'd be in favour of wherever we can re-use functionality

gregfurman (commented):
Hey. Apologies for the delay here. Since this is a pretty large PR, it's non-trivial to review.

Do you think there's an opportunity here to create a common stream reader between this and the kinesis component that could perhaps alleviate a lot of this duplicated code?

Although more risky (since it'll be touching an otherwise widely used component), I'd be curious whether we could generalise the pattern we used out to other stream readers.

WDYT?

PR #803, opened 2026-04-07
feat(kafka_franz): add extract_tracing_map support

danielavelez12:
## Summary

Adds `extract_tracing_map` and `new_root_span_with_link` fields to the `kafka_franz` input, matching the existing support in the `kafka` (Sarama) and NATS inputs.

## Motivation

When using `kafka_franz` with the `open_telemetry_collector` tracer, Bento creates spans for processors and outputs but they are orphan root spans — not linked to the upstream trace that produced the Kafka message. The trace context (W3C `traceparent`/`tracestate`) is available in message metadata (since `kafka_franz` maps all Kafka headers to metadata), but there's no way to wire it into the tracing system.

The `kafka` (Sarama) input already supports this via `extract_tracing_map`, but `kafka_franz` does not. This means users of `kafka_franz` (and `kafka_franz_warpstream`) cannot achieve end-to-end distributed tracing across Kafka-based pipelines.

## Changes

1. Added `NewExtractTracingSpanMappingField()` and `NewRootSpanWithLinkField()` to the `kafka_franz` config spec
2. Wrapped the input with `WrapBatchInputExtractTracingSpanMapping` in the registration function

## Usage

```yaml
input:
  kafka_franz:
    seed_brokers: ["localhost:9092"]
    topics: ["my-topic"]
    consumer_group: "my-group"
    extract_tracing_map: "root = meta()"
```

With a tracer configured, all downstream processor and output spans will now be children of the upstream trace extracted from Kafka headers.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #805, opened 2026-04-07
feat(tracer): auto-configure OTLP tracer from OTel environment variables

danielavelez12:
## Summary

When no `tracer` is explicitly configured (default `none`), bento checks for standard OTel env vars and auto-configures an OTLP trace exporter. Combined with `extract_tracing_map` (#803), this enables end-to-end distributed tracing in managed environments where the `tracer` YAML key is unavailable.

## Motivation

In managed environments like WarpStream data pipelines, the top-level `tracer` config key is [not supported](https://docs.warpstream.com/warpstream/kafka/manage-connectors/bento) (it's a runtime concern managed by the platform). However, WarpStream agents support [environment variable interpolation](https://docs.warpstream.com/warpstream/kafka/manage-connectors/bento) — operators set env vars on their agent deployment (K8s, ECS, etc.) and they're available to pipeline configs at runtime.

This means operators can set `OTEL_EXPORTER_OTLP_ENDPOINT` on their WarpStream agents, and bento will auto-configure an OTLP trace exporter without needing the blocked `tracer` YAML key.

### Setup

1. Set `OTEL_EXPORTER_OTLP_ENDPOINT` (and optionally `OTEL_SERVICE_NAME`) as env vars on your WarpStream agent deployment
2. Remove any `tracer` block from your pipeline YAML
3. Bento auto-detects the env var and configures the OTLP exporter
4. Use `extract_tracing_map` (#803) on `kafka_franz` inputs to connect upstream trace context

No changes needed from the WarpStream team — operators already control agent env vars.

## Design

- `TracerEnvFunc` registration pattern avoids circular imports between `bundle` and `impl/otlp`
- OTLP package registers `NewFromEnv()` on `AllTracers` via `init()`
- `TracerSet.Init()` tries env functions when `conf.Type == "none"` before using the noop tracer
- `Environment.Clone()` propagates env functions

### Supported env vars

Env var names and precedence follow the [OTel specification](https://opentelemetry.io/docs/specs/otel/protocol/exporter/). Signal-specific takes precedence over generic:

| Variable | Description |
|---|---|
| `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` | Traces-specific endpoint (takes precedence) |
| `OTEL_EXPORTER_OTLP_ENDPOINT` | Generic endpoint (fallback) |
| `OTEL_EXPORTER_OTLP_TRACES_PROTOCOL` | Traces-specific protocol (takes precedence) |
| `OTEL_EXPORTER_OTLP_PROTOCOL` | `http/protobuf` (default) or `grpc` |
| `OTEL_SERVICE_NAME` | `service.name` resource attribute |

This is basic OTLP auto-config. Advanced knobs (headers, certs, compression, sampler) require the explicit `open_telemetry_collector` tracer config.

## Test plan

- [x] Env var tests: no vars, generic endpoint, traces endpoint/protocol precedence, HTTP protocol
- [x] Endpoint parsing: bare host:port, k8s FQDN, http/https schemes, URLs with paths
- [x] All existing tests pass (otlp, bundle, public/service)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/internal/bundle/environment.go
+++ b/internal/bundle/environment.go
@@ -69,6 +69,7 @@ func (e *Environment) Clone() *Environment {
 	for _, v := range e.tracers.specs {
 		_ = newEnv.tracers.Add(v.constructor, v.spec)
 	}
+	newEnv.tracers.envFns = e.tracers.envFns
 	for _, v := range e.scanners.specs {
 		_ = newEnv.scanners.Add(v.constructor, v.spec)
 	}
--- a/internal/bundle/tracers.go
+++ b/internal/bundle/tracers.go
@@ -44,9 +44,15 @@ type tracerSpec struct {
 	spec        docs.ComponentSpec
 }
 
+// TracerEnvFunc is a function that attempts to create a TracerProvider from
+// environment variables. It returns (nil, nil) when no relevant env vars are
+// set.
+type TracerEnvFunc func() (trace.TracerProvider, error)
+
 // TracerSet contains an explicit set of tracers available to a Bento service.
 type TracerSet struct {
-	specs map[string]tracerSpec
+	specs  map[string]tracerSpec
+	envFns []TracerEnvFunc
 }
 
 // Add a new tracer to this set by providing a spec (name, documentation, and
@@ -66,8 +72,29 @@ func (s *TracerSet) Add(constructor TracerConstructor, spec docs.ComponentSpec) 
 	return nil
 }
 
-// Init attempts to initialise an tracer from a config.
+// RegisterEnvFn registers a function that can create a TracerProvider from
+// environment variables. Called as a fallback when no tracer is explicitly
+// configured.
+func (s *TracerSet) RegisterEnvFn(fn TracerEnvFunc) {
+	s.envFns = append(s.envFns, fn)
+}
+
+// Init attempts to initialise a tracer from a config. When the default noop
+// tracer is selected, it tries registered env functions (e.g. OTel env vars)
+// first.
 func (s *TracerSet) Init(conf tracer.Config, nm NewManagement) (trace.TracerProvider, error) {
+	if conf.Type == "none" {
+		for _, fn := range s.envFns {
+			tp, err := fn()
+			if err != nil {
+				return nil, fmt.Errorf("auto-configuring tracer from env: %w", err)
+			}
+			if tp != nil {
+				return tp, nil
+			}
+		}
+	}
+
 	spec, exists := s.specs[conf.Type]
 	if !exists {
 		return nil, component.ErrInvalidType("tracer", conf.Type)
--- a/internal/impl/otlp/tracer_otlp_env.go
+++ b/internal/impl/otlp/tracer_otlp_env.go
@@ -1,0 +1,74 @@
+package otlp
+
+import (
+	"cmp"
+	"context"
+	"os"
+
+	"go.opentelemetry.io/otel/attribute"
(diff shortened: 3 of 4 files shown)


jem-davies:
> In managed environments like WarpStream data pipelines, the top-level tracer config key is [not supported](https://docs.warpstream.com/warpstream/kafka/manage-connectors/bento) (it's a runtime concern managed by the platform).

This feels more like a constraint of managed data pipelines rather than something intrinsic to Bento. As a result, this PR's motivation reads like a workaround for that specific environment.

> Env var names and precedence follow the [OTel specification](https://opentelemetry.io/docs/specs/otel/protocol/exporter/). Signal-specific takes precedence over generic:

Ok, but then also you see on the [WithEndpoint](https://pkg.go.dev/go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracegrpc@v1.40.0#WithEndpoint) option for example, it will already utilise these env vars, providing that the endpoint isn't set explicitly. I think because of the default Bento config options it will always be set explicitly, but still, we could utilise this rather than reading these env vars ourselves?

Also this note regarding: https://opentelemetry.io/docs/languages/go/exporters/#:~:text=%7D-,Note,-The%20standard%20OTLP.

danielavelez12:
> > In managed environments like WarpStream data pipelines, the top-level tracer config key is [not supported](https://docs.warpstream.com/warpstream/kafka/manage-connectors/bento) (it's a runtime concern managed by the platform).
> 
> This feels more like a constraint of managed data pipelines rather than something intrinsic to Bento. As a result, this PR's motivation reads like a workaround for that specific environment.
> 
> > Env var names and precedence follow the [OTel specification](https://opentelemetry.io/docs/specs/otel/protocol/exporter/). Signal-specific takes precedence over generic:
> 
> Ok, but then also you see on the [WithEndpoint](https://pkg.go.dev/go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracegrpc@v1.40.0#WithEndpoint) option for example, it will already utilise these env vars, providing that the endpoint isn't set explicitly. I think because of the default Bento config options it will always be set explicitly, but still, we could utilise this rather than reading these env vars ourselves?
> 
> Also this note regarding: https://opentelemetry.io/docs/languages/go/exporters/#:~:text=%7D-,Note,-The%20standard%20OTLP.

Right, it's more of a fallback since Warpstream doesn't let us configure a tracer via YAML 😅 I think it could make sense for Bento to still export traces if there's OTel env vars present, even if there's no tracer config present

Thanks for the guidance, I simplified the implementation a bit, let me know what you think!

jem-davies:
@danielavelez12 - can you go to the warpstream community slack channel: https://console.warpstream.com/socials/slack and DM @richardartoul regarding this - he might be able to allow the setting of a tracer via the `tracer` YAML field. 

> it could make sense for Bento to still export traces if there's OTel env vars present

That is true and I agree - but maybe if we can do it explicitly in the YAML config that would be better?

jem-davies:
@danielavelez12 or you might find him named 'richie'

jem-davies:
@danielavelez12 - or use the public 'Bento' Channel there 🙏

danielavelez12:
> @richardartoul

Yep! Just messaged Richie, I'll keep you posted

jem-davies:
@danielavelez12 - hey, are we sorting this out in the warpstream agents? -> can we close this issue? Thanks, 🙏

danielavelez12:
> @danielavelez12 - hey, are we sorting this out in the warpstream agents? -> can we close this issue? Thanks, 🙏

Yeah can close this out!

PR #806, opened 2026-04-09
feat(grpc_client): add metadata support to gRPC client output

comigor:
## Summary

- Adds a new `metadata` config field to the `grpc_client` output, allowing users to attach custom key/value metadata pairs to outgoing gRPC requests
- Values support Bloblang interpolation for dynamic per-message metadata (e.g. `${! json("name") }`, `${! metadata("request_id") }`)
- Metadata is injected into all RPC types: unary (per-message), client_stream (per-batch), server_stream (per-message), and bidi (per-batch)
- Clearly documents the per-RPC-type behavior, noting that `client_stream` and `bidi` use metadata from the first message only (gRPC protocol constraint)

## Example Usage

```yaml
output:
  grpc_client:
    address: localhost:50051
    service: helloworld.Greeter
    method: SayHello
    reflection: true
    metadata:
      x-request-id: ${! uuid_v4() }
      x-source: bento
      x-correlation-id: ${! metadata("correlation_id") }
```

## Implementation Details

- New `metadata` field registered via `NewInterpolatedStringMapField` (consistent with how `headers` works in websocket, NATS, HTTP outputs)
- `contextWithMetadata` helper resolves interpolations against the current message and appends pairs to the outgoing gRPC context via `grpc/metadata.AppendToOutgoingContext`
- For stream-based RPCs (client_stream, bidi), metadata is resolved from the first message in the batch since the stream is opened once per batch — this is a gRPC protocol constraint (metadata = HTTP/2 headers = sent once at stream open)
- Added a "Metadata" section to the component description with a callout warning about the stream-level behavior, similar to the archive processor warning on the `aws_s3` output docs

## Tests

5 new test cases covering:
- Static metadata on unary, client stream, server stream, and bidi RPCs
- Bloblang interpolation producing dynamic per-message metadata values
- All existing tests continue to pass with no regressions

### DIFF
--- a/internal/impl/grpc_client/client_output.go
+++ b/internal/impl/grpc_client/client_output.go
@@ -47,7 +47,7 @@ const (
 
 const grpcClientOutputDescription = `
 
-### Expected Message Format 
+### Expected Message Format
 
 Either the field ` + "`reflection` or `proto_files`" + ` must be supplied, which will provide the protobuf schema Bento will use to marshall the Bento message into protobuf.
 
@@ -120,7 +120,7 @@ output:
 				Default([]any{}).
 				Example([]string{"./grpc_test_server/helloworld.proto"}),
 			service.NewInterpolatedStringMapField(grpcClientOutputMetadata).
-				Description("A map of metadata key/value pairs to add to gRPC requests. Values support [interpolation functions](/docs/configuration/interpolation#bloblang-queries). For `unary` and `server_stream` RPC types, metadata is evaluated per message. For `client_stream` and `bidi` RPC types, metadata is evaluated from the first message in the batch only, since gRPC stream metadata is sent once at stream creation.").
+				Description("A map of metadata key/value pairs to add to gRPC requests. For `unary` and `server_stream` RPC types, metadata is evaluated per message. For `client_stream` and `bidi` RPC types, metadata is evaluated from the first message in the batch only, since gRPC stream metadata is sent once at stream creation.").
 				Example(map[string]any{
 					"application":  "bento",
 					"x-request-id": `${!metadata("request_id")}`,
@@ -146,7 +146,7 @@ output:
 			service.NewOutputMaxInFlightField(),
 			service.NewBatchPolicyField(grpcClientOutputBatching),
 		).LintRule(
-		`root = match { 
+		`root = match {
   this.rpc_type == "bidi" && this.propagate_response == true => "cannot set propagate_response to true when rpc_type is bidi",
   this.reflection == false && (!this.exists("proto_files") || this.proto_files.length() == 0) => "reflection must be true or proto_files must be populated"
 }`,
--- a/internal/impl/grpc_client/client_output.go
+++ b/internal/impl/grpc_client/client_output.go
@@ -19,6 +19,7 @@ import (
 	"google.golang.org/grpc/credentials/oauth"
 	_ "google.golang.org/grpc/health"
 	"google.golang.org/grpc/health/grpc_health_v1"
+	grpcmd "google.golang.org/grpc/metadata"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


jem-davies:
Hey this should fix the `make docs` on the CI: https://github.com/decade-eng/bento/pull/2

internal/impl/grpc_client/client_output.go
```diff
@@ -102,6 +104,14 @@ output:
 				Description("A list of filepaths of .proto files that should contain the schemas necessary for the gRPC method.").
 				Default([]any{}).
 				Example([]string{"./grpc_test_server/helloworld.proto"}),
+			service.NewInterpolatedStringMapField(grpcClientOutputMetadata).
+				Description("A map of metadata key/value pairs to add to gRPC requests. Values support [interpolation functions](/docs/configuration/interpolation#bloblang-queries).").
```
jem-davies:
```suggestion
				Description("A map of metadata key/value pairs to add to gRPC requests.").
```

Mention of the interpolation gets added automatically - so need to remove this.

jem-davies (dismissed):
Looks good - just a fix for the CI https://github.com/decade-eng/bento/pull/2 - and a minor doc fix

internal/impl/grpc_client/client_output.go
```diff
@@ -527,8 +562,12 @@ func (gcw *grpcClientWriter) unaryHandler(ctx context.Context, msgBatch service.
 }
 
 func (gcw *grpcClientWriter) clientStreamHandler(ctx context.Context, msgBatch service.MessageBatch) error {
+	rpcCtx, err := gcw.contextWithMetadata(ctx, msgBatch[0])
```
jem-davies, line 580: Is it not possible to use all the messages in a batch to produce grpc ctx metadata? (suspect not) ... 

But we will need to update the docs to document this behaviour clearly because this will trip people up - like: https://warpstreamlabs.github.io/bento/docs/components/outputs/aws_s3 with the archive processor warning and explanation.

jem-davies (dismissed):

internal/impl/grpc_client/client_output.go
comigor: Hmm yeah, I don't it works. gRPC metadata is sent as headers when the stream is established, before any messages flow, and only at that time. This is the same constraint for bidi streams, actually. I'll add the explanation!

comigor:
@jem-davies just updated the docs, as we have no way of setting per-message meta on a stream. I'm still having some trouble with `make docs` locally, even though I'm running it with flags
```bash
$ CGO_ENABLED=1 TAGS=x_bento_extra GOTOOLCHAIN=local make docs
website/docs/components/processors/sql_raw.md(147,1) Cannot use DuckDB driver outside of CGO build. Use a binary built with tag x_bento_extra, or the CGO Docker image
website/docs/components/processors/sql_raw.md(170,1) Cannot use DuckDB driver outside of CGO build. Use a binary built with tag x_bento_extra, or the CGO Docker image
website/docs/components/outputs/sql_insert.md(145,1) Cannot use DuckDB driver outside of CGO build. Use a binary built with tag x_bento_extra, or the CGO Docker image
website/docs/components/inputs/sql_raw.md(127,1) Cannot use DuckDB driver outside of CGO build. Use a binary built with tag x_bento_extra, or the CGO Docker image
make: *** [docs] Error 1
```

It changes the docs but still exits 1, so I'm unsure if it will work on the pipeline.

jem-davies:
Ok great - thanks a lot 🙏  - looking to get more grpc components in soon - with a server similar to the http_server stuff 👍

comigor:
@jem-davies Thank you for your support! On other note, I had to use grpc_client with metadata because I'm sending data to [SpiceDB](https://github.com/authzed/api). How do you feel about creating an output component specifically for spicedb? I could work on that as well.

jem-davies:
@comigor - a spiceDB connector sounds good - difficult to say no really 😄 

I want to get the https://github.com/warpstreamlabs/bento/pull/764 (grpc_client input) done then add some tests from the [integration test suite](https://pkg.go.dev/github.com/warpstreamlabs/bento/public/service/integration#StreamTests) and take the grpc_client stuff out of experimental at that point.

PR #809, opened 2026-04-10
feat(tracing): inject message content into processor and output span …

hydrogen2:
…attributes

Adds message.content as a span attribute on every processor span (single and batched) and every output span. This enables external tracers (e.g. custom OTel exporters) to capture per-message provenance data without per-pipeline configuration.

Used by EvoIoT for unified provenance/audit logging.

### DIFF
--- a/internal/component/output/async_writer.go
+++ b/internal/component/output/async_writer.go
@@ -205,6 +205,13 @@ func (w *AsyncWriter) loop() {
 
 			w.log.Trace("Attempting to write %v messages to '%v'.\n", ts.Payload.Len(), w.typeStr)
 			_, spans := tracing.WithChildSpans(w.tracer, traceName, ts.Payload)
+			// Inject message content into output span attributes for provenance
+			_ = ts.Payload.Iter(func(i int, part *message.Part) error {
+				if i < len(spans) {
+					spans[i].SetTag("message.content", string(part.AsBytes()))
+				}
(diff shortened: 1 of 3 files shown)


internal/component/output/async_writer.go
```diff
@@ -205,6 +205,13 @@ func (w *AsyncWriter) loop() {
 
 			w.log.Trace("Attempting to write %v messages to '%v'.\n", ts.Payload.Len(), w.typeStr)
 			_, spans := tracing.WithChildSpans(w.tracer, traceName, ts.Payload)
+			// Inject message content into output span attributes for provenance
+			_ = ts.Payload.Iter(func(i int, part *message.Part) error {
+				if i < len(spans) {
+					spans[i].SetTag("message.content", string(part.AsBytes()))
```
gregfurman, line 211: I'm a bit concerned of the performance implications of an extra loop + `part.AsBytes()` on each message (which will marshal a structured payload to JSON form) -- especially if we are enabling this by default.

Also, is it best practice to include the entire payload as a tag? What if the payload is a couple of MB?

jem-davies: Agree - it doesn't seem right to include the message content as a tag

gregfurman:
Hey @hydrogen2  👋 Are you still keen to work on this or should we close it?

PR #811, opened 2026-04-11
add sanitize_namespace_names field to schema_registry_decode processor

jem-davies:


internal/impl/confluent/client.go
```diff
@@ -130,10 +214,32 @@ func (c *schemaRegistryClient) GetSchemaBySubjectAndVersion(ctx context.Context,
 		return
 	}
 
+	if c.namespaceNameSanitize {
+		var res map[string]any
+		json.Unmarshal([]byte(resBody), &res)
```
gregfurman: 👀

internal/impl/confluent/client.go
```diff
@@ -130,10 +214,32 @@ func (c *schemaRegistryClient) GetSchemaBySubjectAndVersion(ctx context.Context,
...
+
+		updateNamespaces(res, nil)
+
+		cleaned, nserr := json.Marshal(res)
+		if nserr != nil {
+			c.mgr.Logger().Errorf("failed to re-marshal schema %v", nserr)
+			return resPayload, nserr
+		}
+
+		fmt.Printf("CLEANED: \n\n%v\n\n", string(cleaned))
```
gregfurman: 🧹 
```suggestion
```

internal/impl/confluent/client.go
```diff
@@ -51,12 +109,18 @@ func newSchemaRegistryClient(
...
 		client:                hClient,
 		schemaRegistryBaseURL: u,
 		requestSigner:         reqSigner,
 		mgr:                   mgr,
-	}, nil
+	}
+
+	for _, o := range opts {
+		o(scr)
+	}
```
gregfurman: nit:


```suggestion
	for _, opt := range opts {
		opt(scr)
	}
```

internal/impl/confluent/client.go
```diff
@@ -11,15 +11,84 @@ import (
...
 )
 
+func sanitizeNamespacePart(n, k string) string {
+	parts := strings.Split(n, ".")
+	if k == "name" {
+		if len(parts) <= 1 {
+			return n
+		}
+		for i, p := range parts[:len(parts)-1] {
+			parts[i] = regexp.MustCompile(`[^A-Za-z0-9_]`).ReplaceAllString(p, "")
```
gregfurman: nit: if we're going the regex approach, this `regex.MustCompile` should probably be instantiated at build time instead of on every part.

i.e

```golang
var validAlnumRegex = regexp.MustCompile(`[^A-Za-z0-9_]`)

func sanitizeNamespacePart(n, k string) string {
   // ...
			parts[i] = validAlnumRegex.ReplaceAllString(p, "")
   // ...
}
```

internal/impl/confluent/client.go
```diff
@@ -11,15 +11,84 @@ import (
...
+		return strings.Join(parts, ".")
+	} else {
+		for i, p := range parts {
+			parts[i] = regexp.MustCompile(`[^A-Za-z0-9_]`).ReplaceAllString(p, "")
+		}
+		return strings.Join(parts, ".")
+	}
+}
+
+func updateNamespaces(res1 map[string]any, res2 []any) {
```
gregfurman: IMO this type signature is pretty confusing since we never expect `res1` and `res2` to both be non-nil. Why not we rather just do a type switch and handle each type-case individually?


```suggestion
func updateNamespaces[T interface{ map[string]any | []any }](res T) {
	switch any(res).(type) {
	case []any:
		// ...
	case map[string]any:
		// ...
	}
```

internal/impl/confluent/client.go
```diff
@@ -95,10 +171,34 @@ func (c *schemaRegistryClient) GetSchemaByID(ctx context.Context, id int) (resPa
...
+
+		updateNamespaces(res, nil)
+
+		cleaned, nserr := json.Marshal(res)
+		if nserr != nil {
+			c.mgr.Logger().Errorf("failed to re-marshal schema '%v': %v", id, nserr)
+			return resPayload, nserr
+		}
+
+		if err = json.Unmarshal(cleaned, &resPayload); err != nil {
```
gregfurman: The `SchemaInfo` doesn't have a `namespace` attribute -- are you sure that including `"namespace"` attribute sanitization even makes it to the resulting populated `SchemaInfo` struct?

https://github.com/warpstreamlabs/bento/blob/2fc25fd5313e03b94fd687b85fd5273ab4df0ad4/internal/impl/confluent/client.go#L62-L67

gregfurman (changes requested):

internal/impl/confluent/client.go
jem-davies: The schema gets stored in the `Schema` field of that struct as a JSON string - the sanitised namespace are inside that. The new tests in `processor_schema_registry_decode_test.go` wouldn't pass if not.

internal/impl/confluent/client.go
jem-davies: I think if you use generics like this you actually lose the ability to iterate over the types 🤔 - but made an alteration such that we have: 

```go
func updateNamespaces(res any) {
	switch val := res.(type) {
	case map[string]any:
	    // ...
        case []any:
        // ...
}
```

jem-davies:
@gregfurman - thanks 🙏 I have added a commit to action your feedback.

internal/impl/confluent/processor_schema_registry_decode.go
```diff
@@ -67,8 +67,8 @@ This processor decodes protobuf messages to JSON documents, you can read more ab
 			Description("Whether Avro Schemas are nested. If true bento will resolve schema references. (Up to a maximum depth of 100)").
 			Advanced().Default(false).Version("1.2.0")).
 		Field(service.NewURLField("url").Description("The base URL of the schema registry service.")).
-		Field(service.NewTransportField("transport")).Version("1.13.0")
-
+		Field(service.NewTransportField("transport").Version("1.13.0")).
+		Field(service.NewBoolField("sanitize_namespace_names").Description("When enabled, sanitizes namespace names to only contain letters, numbers, and underscores.").Default(false).Version("1.17.0"))
```
gregfurman: Can we mark this as advanced?

internal/impl/confluent/client.go
```diff
@@ -130,10 +224,34 @@ func (c *schemaRegistryClient) GetSchemaBySubjectAndVersion(ctx context.Context,
 		return
 	}
 
+	if c.namespaceNameSanitize {
+		var res map[string]any
+		nserr := json.Unmarshal(resBody, &res)
```
gregfurman: Can we not rather unmarshal the `resBody` into `resPayload`, and then do this sanitization on just the `resPayload.Schema`? This would save us having to unmarshal the same payload fields twice and instead just focus on schema

```golang
	if err = json.Unmarshal(resBody, &resPayload); err != nil {
		c.mgr.Logger().Errorf("failed to parse response for schema '%v': %v", id, err)
		return
	}

	if c.namespaceNameSanitize {
		var res map[string]any
		nserr := json.Unmarshal([]byte(resPayload.Schema), &res)
		if nserr != nil {
			c.mgr.Logger().Errorf("failed to parse response for schema '%v': %v", id, err)
			return resPayload, nserr
		}

		updateNamespaces(res)

		cleaned, nserr := json.Marshal(res)
		if nserr != nil {
			c.mgr.Logger().Errorf("failed to re-marshal schema '%v': %v", id, nserr)
			return resPayload, nserr
		}

		resPayload.Schema = string(cleaned)

		return
	}
```

internal/impl/confluent/client.go
```diff
@@ -11,15 +11,78 @@ import (
...
+			if k == "schema" {
+				if strVal, ok := v.(string); ok {
+					var schemaDef map[string]any
+					if jsonErr := json.Unmarshal([]byte(strVal), &schemaDef); jsonErr == nil {
+						updateNamespaces(schemaDef)
+						if cleaned, jsonErr := json.Marshal(schemaDef); jsonErr == nil {
+							val[k] = string(cleaned)
+						}
+					}
+				}
```
gregfurman: nit: any chance we can un-nest this? i.e

```golang
			if k == "schema" {
				strVal, ok := v.(string)
				if !ok {
					continue
				}
				var schemaDef map[string]any
				if err := json.Unmarshal([]byte(strVal), &schemaDef); err != nil {
					continue
				}
                // ...
```

jem-davies: This changes the behaviour, we want to check to see if the schema field contains a JSON encoded string that we will have to recurse into, we don't want to skip handling the "schema" field in the instance that it is encoded as a JSON object - which is what is happening with the `continue` keyword.

jem-davies: Taken it out after altering from comment https://github.com/warpstreamlabs/bento/pull/811#discussion_r3086356150

internal/impl/confluent/client.go
jem-davies: updated to do ☝️

internal/impl/confluent/client.go
```diff
@@ -131,9 +207,29 @@ func (c *schemaRegistryClient) GetSchemaBySubjectAndVersion(ctx context.Context,
 	}
 
 	if err = json.Unmarshal(resBody, &resPayload); err != nil {
-		c.mgr.Logger().Errorf("failed to parse response for schema subject '%v': %v", subject, err)
+		c.mgr.Logger().Errorf("failed to parse response for schema '%v': %v", subject, err)
```
jem-davies:
```suggestion
		c.mgr.Logger().Errorf("failed to parse response for schema subject '%v': %v", subject, err)
```

internal/impl/confluent/client.go
```diff
@@ -11,15 +11,64 @@ import (
...
+		return strings.Join(parts, ".")
+	} else {
+		for i, p := range parts {
+			parts[i] = validAvroNameCharsRegex.ReplaceAllString(p, "")
+		}
+		return strings.Join(parts, ".")
+	}
+}
+
+func updateNamespaces(res any) {
```
gregfurman, line 40: Nice refactor here! 🤩

internal/impl/confluent/client.go
```diff
@@ -11,15 +11,64 @@ import (
...
+var validAvroNameCharsRegex = regexp.MustCompile(`[^A-Za-z0-9_]`)
+
+func sanitizeNamespacePart(n, k string) string {
+	parts := strings.Split(n, ".")
+	if k == "name" {
+		if len(parts) <= 1 {
+			return n
+		}
+		for i, p := range parts[:len(parts)-1] {
+			parts[i] = validAvroNameCharsRegex.ReplaceAllString(p, "")
```
gregfurman, line 29: nit: Regex can be sluggish but this sanitize code isn't in the hotpath think it's fine to leave as-is

gregfurman (approved):

PR #812, opened 2026-04-13
feat(elasticsearch_v2): add API auth for output

alexthemayers:
This will add the option to use an elastcsearch API key to authenticate to a cluster, allowing a user to avoid the use of basic auth if they wish.

### DIFF
--- a/internal/impl/elasticsearch/output_v2.go
+++ b/internal/impl/elasticsearch/output_v2.go
@@ -30,6 +30,7 @@ const (
 	esoV2FieldTimeout               = "timeout"
 	esoV2FieldTLS                   = "tls"
 	esoV2FieldAuth                  = "basic_auth"
+	esoV2FieldAPIKey                = "api_key"
 	esoV2FieldAuthEnabled           = "enabled"
 	esoV2FieldAuthUsername          = "username"
 	esoV2FieldAuthPassword          = "password"
@@ -100,6 +101,9 @@ Both the `+"`id` and `index`"+` fields can be dynamically set using function int
 				Default("5s"),
 			service.NewTLSToggledField(esoV2FieldTLS),
 			service.NewOutputMaxInFlightField(),
+			service.NewStringField(esoV2FieldAPIKey).
+				Description("Allows you to specify a Base64-encoded token for authorization; if set, overrides basic auth.").
+				Optional(),
 			service.NewObjectField(esoV2FieldAuth,
 				service.NewBoolField(esoV2FieldAuthEnabled).
 					Description("Whether to use basic authentication in requests.").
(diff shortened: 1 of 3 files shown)


internal/impl/elasticsearch/output_v2.go
```diff
@@ -100,6 +101,8 @@ Both the `+"`id` and `index`"+` fields can be dynamically set using function int
 				Default("5s"),
 			service.NewTLSToggledField(esoV2FieldTLS),
 			service.NewOutputMaxInFlightField(),
+			service.NewStringField(esoV2FieldAPIKey).Description("Allows you to specify an API key for authentication.").
```
gregfurman: Could we include a small note in the field deacription saying that this will take precedence over `basic_auth` if specified? The `APIKey` attribute is documented as follows: 

> Base64-encoded token for authorization; if set, overrides username/password and service token.

Also, formatting nit 🙂:

```suggestion
			service.NewStringField(esoV2FieldAPIKey).
			Description("Allows you to specify an API key for authentication.").
```

internal/impl/elasticsearch/output_v2.go
```diff
@@ -211,6 +214,9 @@ func esoV2ConfigFromParsed(pConf *service.ParsedConfig) (conf esoV2Config, err e
 			TLSClientConfig: tlsConf,
 		}
 	}
+	if conf.clientConfig.APIKey, err = pConf.FieldString(esoV2FieldAPIKey); err != nil {
+		return
+	}
```
gregfurman: If a field is marked as `Optional`, this only means that parse step wont require this `api_key` field to be present.

However, this will attempt to extract the field named `api_key` from the parsed config -- despite the fact that it's optional, returning an error if not present.

The fix is to either wrap this with a `if pConf.Contains()` or to set a `Default` on the `Field` definition above i.e `Description(...).Optional().Default("")`

See example from S3:

https://github.com/warpstreamlabs/bento/blob/2fc25fd5313e03b94fd687b85fd5273ab4df0ad4/internal/impl/aws/input_s3.go#L108-L111

gregfurman (changes requested):

gregfurman:
Also, could you run `make docs` and include the changed `.md` for the `elasticsearch_v2` component?

See our `CONTRIBUTING.md` 👇 

https://github.com/warpstreamlabs/bento/blob/2fc25fd5313e03b94fd687b85fd5273ab4df0ad4/CONTRIBUTING.md?plain=1#L44

alexthemayers:
Thanks for the quick review @gregfurman!

I've updated the branch on my end. Let me know if you have any more concerns 🙏

gregfurman (approved):

PR #817, opened 2026-04-15
Feat/s3 checksum algorithm

seanbarzilay:
This adds the option to add a checksum algorithm to use for data integrity verification.
Had an issue writing to an s3 bucket with object lock turned on.

### DIFF
--- a/internal/impl/aws/output_s3.go
+++ b/internal/impl/aws/output_s3.go
@@ -36,6 +36,7 @@ const (
 	s3oFieldTimeout                 = "timeout"
 	s3oFieldKMSKeyID                = "kms_key_id"
 	s3oFieldServerSideEncryption    = "server_side_encryption"
+	s3oFieldChecksumAlgorithm       = "checksum_algorithm"
 	s3oFieldBatching                = "batching"
 )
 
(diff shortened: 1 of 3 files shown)


internal/impl/aws/output_s3.go
```diff
@@ -297,6 +307,12 @@ output:
 				Description("An optional server side encryption algorithm.").
 				Default("").
 				Advanced(),
+			service.NewStringEnumField(s3oFieldChecksumAlgorithm,
+				"CRC32", "CRC32C", "SHA1", "SHA256", "CRC64NVME",
```
gregfurman: I don't see this `CRC64NVME` as an option in the S3 `types/enums.go` 

```go
// Enum values for ChecksumAlgorithm
const (
	ChecksumAlgorithmCrc32  ChecksumAlgorithm = "CRC32"
	ChecksumAlgorithmCrc32c ChecksumAlgorithm = "CRC32C"
	ChecksumAlgorithmSha1   ChecksumAlgorithm = "SHA1"
	ChecksumAlgorithmSha256 ChecksumAlgorithm = "SHA256"
)
```

seanbarzilay: Your'e right it was added in newer versions of the sdk (v1.73.0)

gregfurman (approved):

PR #822, opened 2026-04-21
implement digest auth

### LINKED ISSUE #699: Add digest auth support for http client
https://github.com/icholy/digest

Todo for me

lublak:
fixes #699

### DIFF
--- a/internal/httpclient/client_test.go
+++ b/internal/httpclient/client_test.go
@@ -750,11 +750,10 @@ func TestHTTPClientDigestConf(t *testing.T) {
 
 	conf := clientConfig(t, `
 url: %v
-digest:
+digest_auth:
   enabled: true
-  user:
-    name: test
-    password: 123
+  username: test
+  password: 123
 `, ts.URL+"/testpost")
 
 	h, err := NewClientFromOldConfig(conf, service.MockResources())
--- a/internal/httpclient/client_test.go
+++ b/internal/httpclient/client_test.go
@@ -729,7 +729,7 @@ func TestHTTPClientDigestConf(t *testing.T) {
 		}
 
 		cred, err := digest.ParseCredentials(auth)
-		assert.NoError(t, err)
+		require.NoError(t, err)
 
 		validCred, err := digest.Digest(chal, digest.Options{
 			Method:   r.Method,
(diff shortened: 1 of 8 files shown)


internal/httpclient/config.go
```diff
@@ -101,6 +106,24 @@ func ConfigField(defaultVerb string, forOutput bool, extraChildren ...*service.C
...
+					Default(""),
+				service.NewStringField(hcFieldDigestUserPassword).
+					Description("The password of the user.").
+					Default(""),
+			).
+				Description("The user to use for the authentication.").
+				Optional(),
+		).
+			Description("Digest authentication configuration.").
+			Advanced().
```
jem-davies, line 120:
```suggestion
			Advanced().
			Version("1.18.0").
```

internal/httpclient/config.go
```diff
@@ -101,6 +106,24 @@ func ConfigField(defaultVerb string, forOutput bool, extraChildren ...*service.C
 			Description("An optional HTTP proxy URL.").
 			Advanced().
 			Optional(),
+		service.NewObjectField(hcFieldDigest,
```
jem-davies: Would just simplify this and mirror what we have for the Basic Auth fields: 

```yaml
  basic_auth:
    enabled: false
    username: ""
    password: ""
```

internal/httpclient/client_test.go
```diff
@@ -710,6 +711,66 @@ tls:
...
+
+		validCred, err := digest.Digest(chal, digest.Options{
+			Method:   r.Method,
+			URI:      r.URL.RequestURI(),
+			Cnonce:   cred.Cnonce,
+			Count:    cred.Nc,
+			Username: "test",
+			Password: "123",
+		})
+		assert.NoError(t, err)
```
jem-davies:
```suggestion
		require.NoError(t, err)
```

internal/httpclient/client_test.go
```diff
@@ -710,6 +711,66 @@ tls:
...
+		auth := r.Header.Get("Authorization")
+
+		if auth == "" {
+			w.Header().Add("WWW-Authenticate", chal.String())
+			w.WriteHeader(http.StatusUnauthorized)
+			return
+		}
+
+		cred, err := digest.ParseCredentials(auth)
+		assert.NoError(t, err)
```
jem-davies:
```suggestion
		require.NoError(t, err)
```

Am I right in thinking that a test should use `require` when Arranging & Acting but then use `assert` when  declaring the test conditions? 

Or in other words we should terminate the test run if there is an error from `digest.ParseCredentials(auth)` - because the later `asserts` will have no chance of containing useful info?

jem-davies (commented):
Looks good to me - just a few nit picky comments

jem-davies (approved):

PR #826, opened 2026-04-22
add negotiate

### LINKED ISSUE #378: support http_client sspi

lublak:
replaces #537 fixes #378


gregfurman:
Hey! Sorry for the delay here. Could you do a rebase on `main` to fix these conflicts?

Also, if you could add a brief motivation for these changes? Am not too familiar with SSPI so that'd be super helpful

lublak:
i resolved all conflicts @gregfurman 
The motivation is to support windows authentifications.
My computers are in a domain and the webserver uses https://github.com/quasoft/websspi
With sspi enabled i can use the current user of the computer (logged in) to automatically authenticate to the rest api behind websspi (without the need to add the password to a config file, etc.) For me the sspi part is important, so i can control all permissions over the active directory. But! I also contributed the rest of the implementation if sspi is not available (linux). And also fallback to the old ntlm implementation.

lublak:
Here is the graphic which i created: https://github.com/warpstreamlabs/bento/pull/537#issuecomment-3585187678

gregfurman:
Ahh I see. OK looks like @jem-davies has more info on this then since he was looking at your original.

internal/httpclient/config.go
```diff
@@ -118,7 +130,49 @@ func ConfigField(defaultVerb string, forOutput bool, extraChildren ...*service.C
...
+			service.NewObjectField(hcFieldNegotiateUser,
+				service.NewStringField(hcFieldNegotiateUserDomain).
+					Description("The domain of the user.").
+					Default(""),
+				service.NewStringField(hcFieldNegotiateUserName).
+					Description("The username of the user.").
+					Default(""),
+				service.NewStringField(hcFieldNegotiateUserPassword).
+					Description("The password of the user.").
+					Default(""),
```
jem-davies, line 155: Add a call to Secret()

jem-davies (approved):
LGTM - got a bit confused about the other fields for a second there..

PR #832, opened 2026-04-28
read bento config from env var

### LINKED ISSUE #60: Read bento config from environment variables
From [Bento Lambda Setup | Passing in config from env](https://discord.com/channels/1130882542451556482/1255548874207002706) Discord thread.

Currently, only the `bento-lambda` distribution allows for reading of bento config from the environment via the `BENTO_CONFIG` and `BENTO_CONFIG_PATH` variables.

https://github.com/warpstreamlabs/bento/blob/1d22b5d9e7f0214e3a074e191eec21ee6b7d9a35/internal/serverless/lambda/lambda.go#L31-L41

We should extend `bento` to include this functionality.

jem-davies:
fixes #60

### DIFF
--- a/internal/cli/common/reader.go
+++ b/internal/cli/common/reader.go
@@ -1,6 +1,8 @@
 package common
 
 import (
+	"os"
+
 	"github.com/warpstreamlabs/bento/internal/config"
 	"github.com/warpstreamlabs/bento/internal/filepath/ifs"
 
@@ -11,6 +13,13 @@ import (
 // config.Reader based on input CLI flags. This includes applying any config
 // overrides expressed by the --set flag.
 func ReadConfig(c *cli.Context, cliOpts *CLIOpts, streamsMode bool) (mainPath string, inferred bool, conf *config.Reader) {
+	opts := []config.OptFunc{
+		config.OptSetFullSpec(cliOpts.MainConfigSpecCtor),
+		config.OptAddOverrides(c.StringSlice("set")...),
+		config.OptTestSuffix("_bento_test"),
+		config.OptSetLintConfigWarnDeprecated(),
+	}
+
 	path := c.String("config")
 	if path == "" {
 		// Iterate default config paths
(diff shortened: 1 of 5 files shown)


internal/cli/common/reader.go
```diff
@@ -21,12 +30,10 @@ func ReadConfig(c *cli.Context, cliOpts *CLIOpts, streamsMode bool) (mainPath st
...
-	}
-	opts := []config.OptFunc{
-		config.OptSetFullSpec(cliOpts.MainConfigSpecCtor),
-		config.OptAddOverrides(c.StringSlice("set")...),
-		config.OptTestSuffix("_bento_test"),
-		config.OptSetLintConfigWarnDeprecated(),
+		// Finally check the "BENTO_CONFIG" env var
+		if os.Getenv("BENTO_CONFIG") != "" {
+			opts = append(opts, config.OptLoadFromEnvVar())
+		}
```
gregfurman, line 36: nit: we read this twice which IMO is kinda unideal. Wdyt of setting this once at startup? Or maybe using `CLIOpts` to pass it in?

internal/cli/common/reader.go
```diff
@@ -1,6 +1,8 @@
 package common
 
 import (
+	"os"
```
gregfurman, line 4: This makes me think we should be extending our `ifs` to account for environment variables, instead of directly interacting with the `os`. Not actionable for now I suppose but a good future addition.

internal/config/reader.go
```diff
@@ -183,6 +184,14 @@ func OptUseFS(fs ifs.FS) OptFunc {
 	}
 }
 
+// OptLoadFromEnvVar will configure the reader to read config from the
+// BENTO_CONFIG environment variable
+func OptLoadFromEnvVar() OptFunc {
```
gregfurman: Maybe we read this once in here?

internal/config/lint.go
```diff
@@ -109,3 +109,32 @@ func ReadFileEnvSwap(store ifs.FS, path string, lookupEnvFn func(name string) (s
 	}
 	return
 }
+
+func ReadBentoConfigEnvVarEnvSwap(lookupEnvFn func(name string) (string, bool)) (configBytes []byte, lints []docs.Lint, err error) {
+	var configString string
+	if configString = os.Getenv("BENTO_CONFIG"); configString == "" {
```
gregfurman: Can we make `BENTO_CONFIG` a constant that we import around instead? Alternatively, wdyt of a simple


```
package config

var BentoConfigEnv = os.Getenv("BENTO_CONFIG")
```

that we use instead of checking this multiple times?

gregfurman (approved):
This looks fine to go in but there are some improvements that I think would help with the future proofing of this approach.

Thanks for picking this bad-boy up though 🙂

internal/cli/common/reader.go
jem-davies: Added a commit so we will only read "BENTO_CONFIG" once - wdyt: https://github.com/warpstreamlabs/bento/pull/832/changes/125989b9e5532eae83c0b0fbeac12075635ef994 ?

internal/config/reader.go
jem-davies: Yeah good idea: https://github.com/warpstreamlabs/bento/pull/832/changes/125989b9e5532eae83c0b0fbeac12075635ef994

internal/config/lint.go
jem-davies: Have been able to refactor such we are only reading it once - I wouldn't be against another PR that would move `BENTO_CONFIG` & `BENTO_USE_NUMBER` into one specific place...

PR #836, opened 2026-05-05
move test helper code to _test.go files

jem-davies:
fixes govulncheck github action: https://github.com/jem-davies/bento/actions/runs/25389666982

govulncheck will ignore any source code for test cases - this PR moves 2 remaining CVEs to _test.go files .

### DIFF
--- a/internal/impl/elasticsearch/aws/integration_test.go
+++ b/internal/impl/elasticsearch/aws/integration_test.go
@@ -6,15 +6,20 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"os"
+	"strconv"
 	"testing"
+	"time"
 
 	"github.com/olivere/elastic/v7"
+	"github.com/ory/dockertest/v3"
+	"github.com/ory/dockertest/v3/docker"
+	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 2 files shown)


internal/impl/elasticsearch/aws/integration_test.go
```diff
@@ -189,3 +194,75 @@ output:
 		integration.StreamTestOptPort(servicePort),
 	)
 }
+
+// TODO: Add config + options pattern or use an already existing library like https://github.com/elgohr/go-localstack
+func GetLocalStack(t testing.TB, envVars []string, readyFns ...func(port string) error) (port string) {
```
gregfurman, line 199: Should we remove the other fixture in favor of this one?

jem-davies: So rather than export this function from the `aws` package: I am shamelessly coping it from there. 

We need to have this function in a file with suffix `_test.go` such that the govulncheck doesn't scan it (or it doesn't end up in a prod build). 

If there is a way to not duplicate and also not have the function in the production binary / part of govulncheck - that would be better but afaik - there isn't a way to do that 🤔

gregfurman (approved):

PR #844, opened 2026-05-12
feat(protobuf): Add emit_unpopulated option to protobuf processor

### LINKED ISSUE #496: protobuf processor: allow setting of EmitDefaults
I assume that bento is using the https://pkg.go.dev/github.com/golang/protobuf/jsonpb library to encode and decode protobuf.

I would like to suggest that, within the protobuf processor, we be able to set the marshaller's emitDefaults value to true or false. In my case I decode some protobuf files with integers set to 0. These fields are removed entirely with the default setting  that are used by bento.

For example: 

Protobuf schema:
```
syntax = "proto3";

message Environment {
    string sensorname = 1;
    uint32 temp = 2;
}
```
If sensorname is set to "sensor1" and temp is set to 0, the resulting JSON will look like this (with emitDefaults = false / default behavior):

```
{ 
"sensorname": "sensor1"
}
```
 
When emitDefaults is set to true the JSON would look like this instead:

```
{ 
"name": "jack",
"temp": 0
}
```

Both options can make sense depending on the use case, so I think it would be useful to allow configuring this value.

Greetings!

matta-dev:
Adds `emit_unpopulated` as an option to the protobuf processor which allows for protobuf fields that are not set to be present with their default values in the JSON output of the `to_json` option. `false` by default so no existing behaviour is changed.
closes #496

### DIFF
--- a/internal/impl/protobuf/processor_protobuf.go
+++ b/internal/impl/protobuf/processor_protobuf.go
@@ -32,6 +32,8 @@ const (
 	fieldBsrVersion = "version"
 )
 
+const fieldEmitUnpopulated = "emit_unpopulated"
+
 func protobufProcessorSpec() *service.ConfigSpec {
 	return service.NewConfigSpec().
 		Stable().
@@ -62,6 +64,9 @@ Attempts to create a target protobuf message from a generic JSON structure.
 			Default(false),
 		service.NewBoolField(fieldUseProtoNames).
 			Description("If `true`, the `to_json` operator deserializes fields exactly as named in schema file.").
+			Default(false),
+		service.NewBoolField(fieldEmitUnpopulated).
+			Description("If `true`, the `to_json` operator emits unpopulated fields with their default JSON values.").
 			Default(false),
 		service.NewStringListField(fieldImportPaths).
 			Description("A list of directories containing .proto files or list of file paths, including all definitions required for parsing the target message. If left empty the current directory is used. Each directory listed will be walked with all found .proto files imported. Either this field or `bsr` must be populated.").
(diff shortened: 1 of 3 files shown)


internal/impl/protobuf/processor_protobuf.go
```diff
@@ -32,6 +32,8 @@ const (
 	fieldBsrVersion = "version"
 )
 
+const fieldEmitUnpopulated = "emit_unpopulated"
```
gregfurman: nit: Can this be consolidated with the other constants?

internal/impl/protobuf/processor_protobuf.go
```diff
@@ -294,6 +299,7 @@ func newProtobufToJSONOperator(f fs.FS, msg string, importPaths []string, usePro
 			Resolver:      types,
 			UseProtoNames: useProtoNames,
 		}
+		opts.EmitUnpopulated = emitUnpopulated
```
gregfurman: Let's include this in the `MarshalOptions` struct? i.e

```golang
opts := protojson.MarshalOptions{
			Resolver:               types,
			UseProtoNames:   useProtoNames,
            EmitUnpopulated: emitUnpopulated,
		}
```

internal/impl/protobuf/processor_protobuf.go
```diff
@@ -329,6 +335,7 @@ func newProtobufToJSONBSROperator(multiModuleWatcher *MultiModuleWatcher, msg st
 			Resolver:      multiModuleWatcher,
 			UseProtoNames: useProtoNames,
 		}
+		opts.EmitUnpopulated = emitUnpopulated
```
gregfurman: same as above commen -- let's consolidate this into above struct

gregfurman (changes requested):
Very well considered & neat contribution! Once you address those consolidation comments looks good to merge 🙌 

Also I saw this was your first ever GH contribution -- so warm welcome to the Bento community 🍱

gregfurman (approved):

PR #849, opened 2026-05-12
Streambuilder with basepath

lublak:


lublak:
Notes for me:

```
https://github.com/warpstreamlabs/bento/blob/main/internal/bloblang/parser/context.go#L147
```

PR #855, opened 2026-05-19
fix(sql): goroutine leak on Connect + nil-db panic in writeBatch

adrianhaj:
## Summary

Fixes two related issues in `sql_raw` and `sql_insert` outputs that surfaced under IAM-token rotation, where the framework calls `Connect()` repeatedly to refresh credentials.

- **Goroutine leak (`sql_raw`, `sql_insert`)**: the shutdown-watcher goroutine was spawned inside `Connect()`. Each reconnect spawned a new watcher that lived for the rest of the process. Moved the watcher into the constructor (`newSQLRawOutput` / `newSQLInsertOutputFromConfig`) so it spawns exactly once.
- **Nil-pointer panic in `writeBatch` (`sql_raw`)**: `WriteBatch`'s `ErrBadConn` recovery path nils out `s.db` before reopening. A concurrent `writeBatch` taking the RLock could observe `s.db == nil` and panic on dereference. Added `if s.db == nil { return service.ErrNotConnected }` under the existing RLock — matches the pattern already used in `input_sql_raw.go` and `input_sql_select.go`.

### DIFF
--- a/internal/impl/sql/output_sql_insert.go
+++ b/internal/impl/sql/output_sql_insert.go
@@ -218,6 +218,19 @@ func newSQLInsertOutputFromConfig(conf *service.ParsedConfig, mgr *service.Resou
 		}
 	}
 
+	go func() {
+		<-s.shutSig.HardStopChan()
+
+		s.dbMut.Lock()
+		if s.db != nil {
+			_ = s.db.Close()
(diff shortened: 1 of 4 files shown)


jem-davies (approved):
lgtm

PR #857, opened 2026-05-19
Handle deprecation of optional in cuegen.

aebrahim:
Deprecation described here:
https://github.com/cue-lang/cue/commit/0265d7d62829c1002e3ea58dd6c5b992fea5f2e3

As far as I can tell, this is causing the build failures when upgrading in #843


gregfurman (approved):
LGTM!

PR #858, opened 2026-05-20
Resolve Protobuf Confluent Schema Registry references correctly for subjects containing slashes

PitchBlack07:
Protobuf schema references whose subject contains `/` must be path-escaped when requesting referenced schemas from Confluent Schema Registry.

Preserve the escaped request path so these subjects are resolved correctly instead of being double-escaped or interpreted as multiple URL segments.


internal/impl/confluent/client.go
```diff
@@ -270,9 +270,14 @@ func (c *schemaRegistryClient) walkReferencesTracked(ctx context.Context, seen m
 
 func (c *schemaRegistryClient) doRequest(ctx context.Context, verb, reqPath string) (resCode int, resBody []byte, err error) {
 	reqURL := *c.schemaRegistryBaseURL
-	if reqURL.Path, err = url.JoinPath(reqURL.Path, reqPath); err != nil {
+	var joinedPath string
+	if joinedPath, err = url.JoinPath(reqURL.EscapedPath(), reqPath); err != nil {
```
gregfurman, line 274: If the path is already escaped, wouldn't this double escape it?

internal/impl/confluent/processor_schema_registry_decode_test.go
```diff
@@ -84,7 +84,7 @@ func runSchemaRegistryServer(t testing.TB, fn func(path string) ([]byte, error))
 		reqMut.Lock()
 		defer reqMut.Unlock()
 
-		b, err := fn(r.URL.Path)
+		b, err := fn(r.URL.EscapedPath())
```
gregfurman, line 88: How come we're changing the path here now?

internal/impl/confluent/processor_schema_registry_decode_test.go
gregfurman, line 1: Could we add a test case for this? I want to make sure the logic is correct in the confluent client

gregfurman (changes requested):

internal/impl/confluent/client.go
PitchBlack07: I do not think so. `reqURL` is the URL that is loaded from the configuration and `EscapedPath()` makes sure that the path is properly escaped under Go's encoding rules.

So we are joining the escaped path segment from `reqURL` with the other escaped path segment `reqPath` yielding a consistently escaped `joinedPath`.

The actual fix is setting the `Path` _and_ `RawPath` members of the `URL` structure where `Path` is the unescaped and `RawPath` the escaped variant. This ensures that `reqURL.String()` handed over to `http.NewRequestWithContext()` uses the properly escaped version.

Here is an example how this plays out:

`reqPath` is already in escaped form. `GetSchemaBySubjectAndVersion` assembles a path for a schema reference. In my use cases, these references take the form of `<package>/<proto file>`, because I am using imports in the form of

```
import "commons/SomeEnum.proto"
```
in my Protobuf schemas that are managed via Confluent Schema Registry.

`GetSchemaBySubjectAndVersion` correctly escapes the slash and hands `subjects/commons%2FSomeEnum.proto/version/X` to `doRequest()`.

`joinedPath` makes sure that any path segment in `reqURL` is properly escaped before joining with the other already escaped `reqPath. So far so good.

If we leave it at that, `reqURL.String()` which is handed down to `http.NewRequestWithContext()` would turn my path into `subjects/commons%252FSomeEnum.proto/version/X` because it escapes an already escaped path and thus breaks the schema lookup.

We fix this, setting the `Path` and `RawPath` members of the `URL` struct
```
reqURL.Path, err = url.PathUnescape(jonedPath) // subjects/commons/SomeEnum.proto/version/X
reqURL.RawPath = joinedPath // subjects/commons%FESomeEnum.proto/version/X
```

Now that `URL` knows the already escaped version (RawPath) and the unescaped version (Path), `URL.String()` assembles the final URL correctly.

Long story short. We have to set the `RawPath` correctly, but we have to set `Path` to the unescaped variant, for `reqURL.String()` to return the correctly escaped variant.

internal/impl/confluent/processor_schema_registry_decode_test.go
PitchBlack07: The already existing test `TestProtobufReferences()` is supposed to verify exactly that use case which I have fixed (slash character in protobuf reference name). Unfortunately, the test also has a bug. 

This test runs a schema registry server which already expects the path `/subjects/stuffs%2Fthething.proto/versions/10` when resolving the main schema, i.e. it expects the _escaped_ or _encoded_ variant.

According to the documentation `r.URL.Path` stores the _unescaped_ or _decoded_ variant of the path, which  should be `/subjects/stuffs/thething.proto/versions/10`.

Before my change, the `Path` member stored `/subjects/stuffs%2Fthething.proto/versions/10`, i.e. it treated the encoded variant as decoded form of the path which is not correct. We need to pass the encoded or escaped variant to the function call. This is why we need this  change.

internal/impl/confluent/processor_schema_registry_decode_test.go
PitchBlack07: A test which verifies that schema references containing slash characters are correctly resolved already exists (c.f. `TestProtobufReferences()`).

PitchBlack07: I have added a new `TestProtobufReferencesSpecialCharacters()` test case which validates that `EscapedPath()` returns a properly escaped path which represents the actual unescaped `Path` for the Protobuf references.

gregfurman:
@PitchBlack07 Thanks for addressing my feedback! Will give this another lookover but tbh it's looking pretty good to ship 🙂

gregfurman (approved):
LGTM! Thanks for the neat fix and addressing my review.

PR #131, opened 2024-10-15
add: interpolated string fields to gcp_bigquery

gregfurman:
### Changes

Make `table` field of `gcp_bigquery` interpolatable

PR #860, opened 2026-05-20
chore: bump hugot to v0.7.4

### LINKED ISSUE #859: huggingface: processor Close panics on flow stop due to Hugot double-Destroy
## Summary

`pipelineProcessor.Close` in `internal/impl/huggingface/processor.go` calls `p.pipeline.GetModel().Destroy()` without guarding against double invocation. Bento's pipeline teardown invokes `Close` more than once on the same processor (once at flow stop, once at stream teardown), and on the second call the underlying Hugot `Destroy` closure panics with a nil pointer dereference.

Tracked upstream as [knights-analytics/hugot#130](https://github.com/knights-analytics/hugot/issues/130) (PR: [#131](https://github.com/knights-analytics/hugot/pull/131)). Even after that lands, Bento should still be defensive: stopping a flow shouldn't crash the worker.

## Reproduction

Run any flow that uses `nlp_extract_features` (or any other huggingface processor) and stop it. The worker process exits with:

```
[signal SIGSEGV: segmentation violation code=0x2 addr=0x30]

github.com/knights-analytics/hugot/backends.LoadModel.func1()
    /backends/model.go:82 +0xe8
github.com/warpstreamlabs/bento/internal/impl/huggingface.(*pipelineProcessor).Close
    /internal/impl/huggingface/processor.go:214 +0x30
```

## Affected versions

v1.17.0 (current). The Close code in question has been unchanged since the huggingface processor was added.

## Suggested fix

`internal/impl/huggingface/processor.go` around line 213, defend the Close path:

```go
func (p *pipelineProcessor) Close(context.Context) error {
    if p.pipeline == nil {
        return nil
    }
    model := p.pipeline.GetModel()
    if model == nil || model.Destroy == nil {
        return nil
    }
    // Hugot's Destroy closure does not nil-check its backend fields, so a
    // second invocation panics. Recover until upstream lands.
    defer func() { _ = recover() }()
    return model.Destroy()
}
```

Three layers:
1. Nil `pipeline` - guards against `Close` being invoked before init succeeded.
2. Nil `model`/`Destroy` - guards against partial-init where the closure was never wired up.
3. `recover` - catches the actual upstream panic on the second Close call.

sananguliyev:
Fixes #859.

## Summary

- Bumps `github.com/knights-analytics/hugot` from v0.6.4 to v0.7.4.
- v0.7.3 includes the upstream fix for the double-`Destroy` panic ([knights-analytics/hugot#130](https://github.com/knights-analytics/hugot/issues/130), [PR #131](https://github.com/knights-analytics/hugot/pull/131)) that this branch originally worked around. The local `recover()` guard in `pipelineProcessor.Close` is no longer needed and has been removed.
- Adopts v0.7.x's `context.Context`-aware API: `hugot.NewGoSession`, `hugot.DownloadModel`, `pipeline.Run`, and `pipeline.RunPipeline` all now take a ctx. Constructor sites use `context.Background()` (bento's processor constructor has no ctx to plumb); `ProcessBatch` forwards its own ctx.

## Breaking change

`nlp_classify_tokens` output: the `TokenID` field (single `uint32`) is renamed to `TokenIDs` (a `[]uint32`) to match hugot's renamed `pipelines.Entity` field. Under `WithSimpleAggregation` the slice is currently nil → marshals as `null` in JSON; under other aggregation modes it will contain the subword token IDs.

## Other changes

- Snapshot fixtures regenerated against hugot v0.7.3 + ORT v1.25.1. Token classification scores drift at ~6 decimal places; zero-shot classification scores drift more substantially due to upstream pipeline changes.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./internal/impl/huggingface/...`
- [x] `go test ./internal/impl/huggingface/...` (unit tests)
- [x] `go test ./internal/impl/huggingface/ -run '^TestIntegration_'` — all 5 integration tests pass (text classification, token classification, feature extraction, zero-shot classification, download)


sananguliyev:
Hugot team was fast, and already reviewed and merged. https://github.com/knights-analytics/hugot/issues/130 

Since they plan to release new version, I am waiting for that and then just bump the version instead of having temporary solution.

go.mod
```diff
@@ -464,4 +460,4 @@ require (
 	modernc.org/memory v1.11.0 // indirect
 )
 
-go 1.26
+go 1.26.0
```
gregfurman: I've opened a change upstream for this since it'll mean any project that imports bento will be required to pin this at `1.26.0` (which is functionally equivalent to `1.26` in anycase):

https://github.com/knights-analytics/hugot/pull/132

internal/impl/huggingface/processor.go
```diff
@@ -107,7 +107,7 @@ func newPipelineProcessor(conf *service.ParsedConfig, mgr *service.Resources) (*
 	// To prevent multiple sessions being created at once, rather store the constructor and allow it
 	// to be atomically retrieved.
 	ctorValue, _ := mgr.GetOrSetGeneric(sessionConstructorKey{}, func() (*hugot.Session, error) {
-		return hugot.NewGoSession()
+		return hugot.NewGoSession(context.Background())
```
gregfurman, line 110: nit: What in the breaking change...

gregfurman (commented):
Thanks for the change here! I've submitted a PR to the upstream since am not a big fan of that `go 1.26.0` being pinned. Otherwise, nothing blocking.

Once my pr is accepted, we can do a version bump on the hugot dep and :shipit:

go.mod
sananguliyev: yes, I am also not big fan of pinning `1.26.0`, just did not want to wait. Thanks for opening PR on upstream. It make sense to wait for clean PR here in Bento.

gregfurman: ~The above has been merged. Wdyt of pinning this dep version to the commit SHA `b5a9964775775975f412554118c560c3820836ba` then running another go mod tidy?~

OK new version is out! Let's bump to `v0.7.4`

sananguliyev: Yes, exactly. They released a new version, and I have updated it. Feel free to merge, and looking forward for the next release, and until then I will use commit sha for bento.

gregfurman:
```suggestion
go 1.26
```

sananguliyev: I bumped hugot but forget this. Thanks 👍

PR #864, opened 2026-05-24
feat(parquet): Support nested N-D lists

### LINKED ISSUE #360: [bug] Support nested list attributes in parquet encoder
The `parquet-go/parquet-go` dependency does not currently support an array of arrays, which limits our parquet encoder to 1D arrays during processing.

This is being tracked upstream at https://github.com/parquet-go/parquet-go/issues/55

gregfurman:
## Motivation

Adds support for nested N-D lists in the parquet processors.

Fixes https://github.com/warpstreamlabs/bento/issues/360

## Changes

- Updates `parquet-go` to fork branch at https://github.com/warpstreamlabs/parquet-go/tree/warpstream/dev, which includes support for N-D lists.
- Ensures nested schema structs are created with the new `parquet-element` tag.
- Unskips `TestParquetDecodeListFormatEdgeCases` and includes higher dimensional tests (3D, and 4D) as support has been added.

### DIFF
--- a/internal/impl/parquet/schema.go
+++ b/internal/impl/parquet/schema.go
@@ -28,6 +28,18 @@ func GenerateStructType(
 	return generateStructTypeFromFields(fields, schemaOpts)
 }
 
+type tags []struct {
+	name, value string
+}
+
+func (t tags) String() string {
+	res := make([]string, len(t))
+	for i, tg := range t {
+		res[i] = fmt.Sprintf(`%s:%q`, tg.name, tg.value)
+	}
+	return strings.Join(res, " ")
+}
+
 func generateStructTypeFromFields(
 	fields []*service.ParsedConfig,
 	schemaOpts schemaOpts,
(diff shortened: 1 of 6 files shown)


internal/impl/parquet/schema.go
```diff
@@ -295,6 +317,21 @@ func wrapType(
 	return baseType, nil
 }
 
+func isListOfLists(field *service.ParsedConfig) bool {
```
jem-davies, line 325: Would we need to document that we can't handle more than 2D lists? 
Would it be possible to be able to handle 3D+ lists?

gregfurman: Seems like this is only possible following [github.com/parquet-go/parquet-go/commit/a326b0cd1e5d2089ae051de957c342fc4ea72374](https://github.com/parquet-go/parquet-go/commit/a326b0cd1e5d2089ae051de957c342fc4ea72374#diff-c3b0d951288715c695f51ce2727c87a056e89ccbb4f3003be71baa8b405383e9R1083-R1091) -- tbh I don't think we should be blocking 2D support while waiting for a release containing that 3D fix.

gregfurman: OK N dimensional lists should be working! We're using a patched version at HEAD of https://github.com/warpstreamlabs/parquet-go/tree/warpstream/dev

internal/impl/parquet/schema.go
```diff
@@ -105,13 +117,23 @@ func generateStructTypeFromFields(
...
 
+		tt := tags{
+			{"parquet", parquetTag},
+			{"json", name},
+		}
+
+		if isListOfLists(field) {
+			// includes a 'parquet-element' tag, to associate a list of elements
+			// See https://pkg.go.dev/github.com/parquet-go/parquet-go#SchemaOf
+			tt = append(tt, struct{ name, value string }{"parquet-element", ",list"})
```
ryanworl, line 128: `parquet-element` is hard-coded to `",list"`, which drops the child LIST’s `optional` setting. For example, an optional inner list with input `{"matrix":[null,[1,2]]}` round-trips as `{"matrix":[[],[1,2]]}`, silently conflating null with empty. Could we propagate the child element options into this tag and add a round-trip test covering null versus empty inner lists?

gregfurman: Ahh. Good catch. Looks like this'll require extracting out the tag creation into its own function which shouldn't pollute the diff too much. 

However... extending this component has started to feel like playing whack-a-mole with edge-cases. Wondering if it's worthwhile building the nodes using the `parquet.Node` types instead of this reflection-based approach.

gregfurman:
Closing in favour of https://github.com/warpstreamlabs/bento/pull/971

PR #870, opened 2026-05-27
impl lease & cache for buf schema registry

### LINKED ISSUE #861: `protobuf` processor:  buf schema registry (bsr) should allow the use of a cache
The `protobuf` processor allows users to configure a `bsr` object field that enables the polling of a 'buf schema registry'. 

Considerable numbers of [SchemaWatchers](https://pkg.go.dev/github.com/bufbuild/prototransform#SchemaWatcher) could be created by this processor, (processing threads / streams_mode / k8s pods etc.), it is necessary to allow users to make use SchemaWatcher's functionality pertaining to Caching.

jem-davies:
Fixes #861 

The protobuf processor allows the setting of a `protobuf.bsr` field which can configure multiple clients to retrieve schemas. As reported on this [slack message](https://warpstreamlab-uwt2030.slack.com/archives/C0768UE3NEM/p1779351352943209) - there is the potential that a high volume of requests could be sent from Bento and was advised to implement caching such that only one instance will make calls  to the registry whilst others will retrieve from the cache. 

This PR adds two new config fields: 

 - `protobuf.bsr.cache` -> If set schemas retrieved from the BSR will be saved in a [cache resource](/docs/components/caches/about)
 - `protobuf.bsr.polling_period` -> The period in which to poll the Buf Schema Registry.

### DIFF
--- a/internal/impl/protobuf/multimodule_watcher.go
+++ b/internal/impl/protobuf/multimodule_watcher.go
@@ -1,16 +1,19 @@
 package protobuf
 
 import (
+	"bytes"
 	"context"
 	"errors"
 	"fmt"
 	"net/http"
 	"strings"
+	"sync"
 	"time"
 
 	"buf.build/gen/go/bufbuild/reflect/connectrpc/go/buf/reflect/v1beta1/reflectv1beta1connect"
 	connectrpc "connectrpc.com/connect"
 	"github.com/bufbuild/prototransform"
+	"github.com/bufbuild/prototransform/leaser"
 	"google.golang.org/protobuf/reflect/protoreflect"
 	"google.golang.org/protobuf/reflect/protoregistry"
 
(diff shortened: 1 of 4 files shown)


jem-davies:
I think that the issue could be being caused by long lived protobuf processors' schemaWatchers not being cancelled / cleaned up ...

jem-davies:
> I think that the issue could be being caused by long lived protobuf processors' schemaWatchers not being cancelled / cleaned up ...

https://github.com/warpstreamlabs/bento/pull/873

gregfurman:
@jem-davies Do we still need this following https://github.com/warpstreamlabs/bento/pull/873 being merged?

jem-davies:
> @jem-davies Do we still need this following #873 being merged?

I think it still has value - although perhaps less of a priority after #873

PR #873, opened 2026-06-01
Fix bsr watcher leak

jem-davies:
So the "buf schema registry" poller appears to be leaking because it's context is not cancelled on Processor Close(). 

This PR adds 3 commits: 

 1) Adds goleak to detect goroutine leaks in the ./internal/impl/protobuf tests
 1) Adds context cancellation to the schema watcher
 1) Removes an fmt.Print call 
 1) fixes the linter 
 
 ... 
 
 Previously I have been implementing the "lease" & "cache" option for this processor on #870 - however I think the issue of frequent calls to BSR services will likely be fixed with this PR.


internal/impl/protobuf/processor_protobuf.go
```diff
@@ -549,6 +549,10 @@ func (p *protobufProc) Process(ctx context.Context, msg *service.Message) (servi
 	return service.MessageBatch{msg}, nil
 }
 
-func (p *protobufProc) Close(context.Context) error {
+func (p *protobufProc) Close(ctx context.Context) error {
+	if p.multiModuleWatcher != nil {
+		fmt.Println("Shutting Down watcher")
```
jem-davies:
```suggestion
```

internal/impl/protobuf/multimodule_watcher.go
```diff
@@ -27,10 +29,18 @@ func newMultiModuleWatcher(bsrModules []*service.ParsedConfig) (*MultiModuleWatc
 	if len(bsrModules) == 0 {
 		return nil, errors.New("no modules provided")
 	}
-	multiModuleWatcher := &MultiModuleWatcher{}
+
+	ctx, cancel := context.WithCancel(context.Background())
+	httpClient := &http.Client{
+		Transport: &http.Transport{},
```
gregfurman, line 35: Seems like a good opportunity to use that `httptransport` package, and include the field options (i.e `dial_context`, `timeout`, etc.) in the proto spec.

internal/impl/protobuf/multimodule_watcher.go
```diff
@@ -27,10 +29,18 @@ func newMultiModuleWatcher(bsrModules []*service.ParsedConfig) (*MultiModuleWatc
 	if len(bsrModules) == 0 {
 		return nil, errors.New("no modules provided")
 	}
-	multiModuleWatcher := &MultiModuleWatcher{}
+
+	ctx, cancel := context.WithCancel(context.Background())
```
gregfurman: We shouldn't really be persisting `ctx` types. I think this should rather be a `shutdown.Signaller` that creates a new ctx.

e.g.

```golang
ctx, cancel := multiModuleWatcher.shutSig.SoftStopCtx(context.Background())
```

internal/impl/protobuf/multimodule_watcher.go
```diff
@@ -173,3 +183,13 @@ func (w *MultiModuleWatcher) FindEnumByName(enum protoreflect.FullName) (protore
 	}
 	return nil, fmt.Errorf("could not find %s in any loaded modules", enum)
 }
+
+func (m *MultiModuleWatcher) Close() {
+	m.cancel()
+	for _, v := range m.bsrClients {
```
gregfurman, line 191: Q: Is it a memory leak that we never actually clear these clients stored in the map? Also, what if one stops before we call close -- should it be deleted early? 🤔

internal/impl/protobuf/processor_protobuf.go
```diff
@@ -549,6 +549,9 @@ func (p *protobufProc) Process(ctx context.Context, msg *service.Message) (servi
 	return service.MessageBatch{msg}, nil
 }
 
-func (p *protobufProc) Close(context.Context) error {
+func (p *protobufProc) Close(ctx context.Context) error {
+	if p.multiModuleWatcher != nil {
+		p.multiModuleWatcher.Close()
```
gregfurman, line 554: Should we be setting `p.multiModuleWatcher = nil` here?

internal/impl/protobuf/multimodule_watcher.go
jem-davies: I am not sure - it would seem to over-burden & increase the complexity of the exposed config.

internal/impl/protobuf/multimodule_watcher.go
jem-davies: Added a `m.bsrClients = nil` - It should be ok to call Stop() multiple times.

jem-davies:
@gregfurman - thanks for review - I have responded / added commits to the code 🙏

gregfurman (approved):

PR #883, opened 2026-06-06
feat(read_until): add config for restart backoff

aramperes:
Exposes options for `restartBackoff` for the `read_until` input, for when `restart_input` is `true`.

Documentation has been updated with examples.

Use-case: some inputs can be expensive to query, such as `aws_s3` when walking a bucket. In those cases, increasing the interval to `1s` makes more sense than the default `1ms`.

Exposing `max_elapsed_time` also allows exiting a `read_until` loop after a specific time, even if the `check` condition has not passed yet.

Reference feature request from Redpanda upstream: https://github.com/redpanda-data/connect/issues/1911

### DIFF
--- a/internal/impl/pure/input_read_until.go
+++ b/internal/impl/pure/input_read_until.go
@@ -21,10 +21,11 @@ import (
 )
 
 const (
-	ruiFieldInput       = "input"
-	ruiFieldRestart     = "restart_input"
-	ruiFieldCheck       = "check"
-	ruiFieldIdleTimeout = "idle_timeout"
+	ruiFieldInput          = "input"
+	ruiFieldRestart        = "restart_input"
+	ruiFieldRestartBackoff = "restart_backoff"
+	ruiFieldCheck          = "check"
+	ruiFieldIdleTimeout    = "idle_timeout"
 )
 
 func readUntilInputSpec() *service.ConfigSpec {
@@ -88,6 +89,11 @@ input:
 		service.NewBoolField(ruiFieldRestart).
 			Description("Whether the input should be reopened if it closes itself before the condition has resolved to true.").
 			Default(false),
+		service.NewBackOffField(ruiFieldRestartBackoff, true, &backoff.ExponentialBackOff{
+			InitialInterval: time.Millisecond,
+			MaxInterval:     time.Millisecond * 100,
+			MaxElapsedTime:  0,
+		}).Description("Backoff policy for restarting the child input. Only used when `restart_input` is `true`."),
 	)
 }
 
@@ -106,7 +112,8 @@ func init() {
 }
 
 type readUntilInput struct {
-	restart bool
+	restart        bool
+	restartBackoff *backoff.ExponentialBackOff
 
 	wrappedInputLocked *atomic.Pointer[input.Streamed]
 	check              *mapping.Executor
@@ -142,6 +149,11 @@ func newReadUntilInputFromParsed(conf *service.ParsedConfig, res *service.Resour
 		return nil, err
 	}
 
+	restartBackoff, err := conf.FieldBackOff(ruiFieldRestartBackoff)
+	if err != nil {
+		return nil, err
+	}
+
 	var check *mapping.Executor
 	if checkStr, _ := conf.FieldString(ruiFieldCheck); checkStr != "" {
 		if check, err = mgr.BloblEnvironment().NewMapping(checkStr); err != nil {
@@ -163,7 +175,8 @@ func newReadUntilInputFromParsed(conf *service.ParsedConfig, res *service.Resour
 	wInputLocked := &atomic.Pointer[input.Streamed]{}
 	wInputLocked.Store(&wrapped)
 	rdr := &readUntilInput{
-		restart: restart,
+		restart:        restart,
+		restartBackoff: restartBackoff,
 
 		wrappedCtor:        wrappedCtor,
 		wrappedInputLocked: wInputLocked,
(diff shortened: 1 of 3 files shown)


internal/impl/pure/input_read_until.go
```diff
@@ -195,10 +208,8 @@ func (r *readUntilInput) loop() {
 	}()
 
 	// Prevents busy loop when an input never yields messages.
-	restartBackoff := backoff.NewExponentialBackOff()
-	restartBackoff.InitialInterval = time.Millisecond
-	restartBackoff.MaxInterval = time.Millisecond * 100
-	restartBackoff.MaxElapsedTime = 0
```
gregfurman, line 201: Now that this is configurable, should we be returning early or ensuring the `NextBackoff` call at https://github.com/aramperes/bento/blob/50394970a9b563acedddd020639518587cadcc72/internal/impl/pure/input_read_until.go#L226 is not returning the `backoff.Stop` sentinel? 

If `MaxElapsedTime > 0` and the time elapses, the backoff will just start returning `backoff.Stop` which is a duration value of `-1`. 

Since negative duration values are ignored by `time.After`, this case ends being triggered will cause all subsequent calls to `NextBackoff` to return `-1` -- causing the loop to run without _any_ delay.

internal/impl/pure/input_read_until.go
```diff
@@ -88,6 +89,11 @@ input:
 		service.NewBoolField(ruiFieldRestart).
 			Description("Whether the input should be reopened if it closes itself before the condition has resolved to true.").
 			Default(false),
+		service.NewBackOffField(ruiFieldRestartBackoff, true, &backoff.ExponentialBackOff{
+			InitialInterval: time.Millisecond,
+			MaxInterval:     time.Millisecond * 100,
+			MaxElapsedTime:  0,
+		}).Description("Backoff policy for restarting the child input. Only used when `restart_input` is `true`."),
```
gregfurman: Think we should add a `LintRule` here to ensure the `MaxElapsedTime` is always `0` e.g.


```suggestion
		}).Description("Backoff policy for restarting the child input. Only used when `restart_input` is `true`.").
			LintRule(`root = if this.max_interval != 0 { [ "max_interval can only ever be 0" ] }`),
```

internal/impl/pure/input_read_until.go
```diff
@@ -195,10 +208,8 @@ func (r *readUntilInput) loop() {
 	}()
 
 	// Prevents busy loop when an input never yields messages.
-	restartBackoff := backoff.NewExponentialBackOff()
-	restartBackoff.InitialInterval = time.Millisecond
-	restartBackoff.MaxInterval = time.Millisecond * 100
-	restartBackoff.MaxElapsedTime = 0
+	restartBackoff := *r.restartBackoff
```
gregfurman, line 224: Just want to confirm that this new backoff approach integrates with the current `idle_timeout` field.

For example, what if I set `idle_timeout: 10s` and `restart_backoff.max_interval: 30s`. This could mean that the duration of time before `idle_timeout` is actually triggered should include the backoff delay.

gregfurman (dismissed):
Apologies for the delayed review. This looks good but I'm concerned about a user setting a `max_elapsed_duration` value greater than `0` effectively disabling backoff within `loop`.

Lmk if you still have capacity to address this review!

gregfurman:
👋 @aramperes Are you still able to address my comments? If not, myself or Jem can push this over-the-line. Lemme know!

aramperes:
@gregfurman Hey, thanks for the review and sorry for the delay. I can get back to them next week when I am on holiday. If you'd like to take it over before then, though, feel free.

internal/impl/pure/input_read_until.go
aramperes: Just to confirm, I think you meant to add a LintRule to `max_elapsed_time` ?

```go
}).Description("Backoff policy for restarting the child input. Only used when `restart_input` is `true`.").
    LintRule(`root = if this.max_elapsed_time.parse_duration() > 0 { [ "max_elapsed_time must be 0 (unbounded) for restart_backoff" ] }`),
```

Personally I think having a separate field type/constructor for an "unbounded" BackOffField would be cleaner and less error-prone. Let me know if you are against it.

internal/impl/pure/input_read_until.go
aramperes: If we make `MaxElapsedTime` forced to be `0` (per your other comment; I can force that post-parsing to make sure); then `NextBackoff` will never `Stop`. That would take care of the risk of a busy loop.

internal/impl/pure/input_read_until.go
aramperes: I did not change the behaviour of that specific interaction.

`idle_timeout` timer is reset after each backoff, so a `backoff.max_interval: 30s` means that the input will not be polled for up to 30s, and _then_ the polling itself can take up to 10s (in your example).

internal/impl/pure/input_read_until.go
jem-davies: Added a commit that makes it such it is not possible to configure this with a MaxElapsedTime != 0.

internal/impl/pure/input_read_until.go
jem-davies: Yeah I think this is fine:

idle_timeout - How long to wait whilst the child input is active before shutting down (or restarting) 

restart_backoff - The backoff policy that applies to subsequent restarts of the child input

internal/impl/pure/input_read_until.go
jem-davies: Altered to use a NewObjectField - that will effectively be a BackOffField that can only give you an unbounded backoff

PR #886, opened 2026-06-10
feature(elasticsearch):add elasticsearch as processor

Omar1H1:
## Context:
Adds a new elasticsearch processor that lets us query Elasticsearch as part of a pipeline search, get a document by ID, or delete one. Works the same way [sql_raw](https://github.com/warpstreamlabs/bento/blob/main/internal/impl/sql/processor_sql_raw.go) does for databases: each message in the batch triggers a request, the result replaces the message content, and errors are set per-message so you can handle them with catch/try downstream.

### What it does
Three actions:
1. **search**, runs a Query DSL search, message becomes an array of _source objects (one per hit). Query body is built with args_mapping from the message, same idea as sql_raw's args_mapping.
2. **get**, fetches a single document by ID, message becomes the _source object.
3. **delete**, deletes a document, message passes through unchanged (exec-only).

Both index and id fields support Bloblang interpolation so we can derive them from the message or its metadata.
The processor also sets metadata keys on each output message (es_index, es_id, es_took_ms, es_result_count, es_total_hits, es_found, es_delete_result) so downstream processors can branch on them without re-parsing the body.
### Config
Same basic_auth/TLS/API key fields as elasticsearch_v2 output. Wrap in a branch processor and use result_map to merge the result into the original message rather than replacing it entirely.
### Tests
Integration tests use dockertest to spin up a real ES 8.x instance and run the 3 scenarios end-to-end, including the metadata assertions.

### flowchart 
``` mermaid
flowchart TD
    subgraph Input["Input Message Batch"]
        M1["Message 1"]
        M2["Message 2"]
        M3["Message N"]
    end

    subgraph Processor["EsProcessor.ProcessBatch"]
        LZ["Lazy Connect<br/>client.Info()"]
        CL["Clone Batch"]
        LI["Loop for each message"]
        
        subgraph Actions["Action Switch"]
            SEARCH["case 'search'"]
            GET["case 'get'"]
            DELETE["case 'delete'"]
        end
        
        subgraph SearchProcess["processSearch"]
            AM["args_mapping<br/>Query DSL"]
            ESReq["ES Search Request"]
            ESRes["Parse Response"]
            SH["extractSearchHitsAndMeta"]
        end
        
        subgraph GetProcess["processGet"]
            GR["ES Get Request"]
            GP["Parse Response"]
            GH["extractGetSourceAndMeta"]
        end
        
        subgraph DeleteProcess["processDelete"]
            DR["ES Delete Request"]
            DPR["Parse Response"]
            DM["Add Metadata"]
        end
    end

    subgraph Output["Output Message Batch"]
        OM1["Message 1"]
        OM2["Message 2"]
        OM3["Message N"]
    end

    subgraph Metadata["Metadata Set"]
        MD1["es_index"]
        MD2["es_took_ms"]
        MD3["es_result_count"]
        MD4["es_total_hits"]
        MD5["es_id"]
        MD6["es_found"]
        MD7["es_delete_result"]
    end

    M1 --> LZ
    M2 --> LZ
    M3 --> LZ
    LZ --> CL
    CL --> LI
    LI --> SEARCH
    LI --> GET
    LI --> DELETE
    
    SEARCH --> AM
    AM --> ESReq
    ESReq --> ESRes
    ESRes --> SH
    SH --> MD1
    SH --> MD2
    SH --> MD3
    SH --> MD4
    
    GET --> GR
    GR --> GP
    GP --> GH
    GH --> MD1
    GH --> MD5
    GH --> MD6
    
    DELETE --> DR
    DR --> DPR
    DPR --> DM
    DM --> MD1
    DM --> MD5
    DM --> MD7
    
    SH --> OM1
    GH --> OM2
    DM --> OM3

    classDef success fill:#000000,stroke:#000000
    classDef error fill:#000000,stroke:#000000
    classDef meta fill:#000000,stroke:#000000
    
    class MD1,MD2,MD3,MD4,MD5,MD6,MD7 meta
```


internal/impl/elasticsearch/processor.go
```diff
@@ -0,0 +1,543 @@
...
+	espFieldAuthEnabled  = "enabled"
+	espFieldAuthUsername = "username"
+	espFieldAuthPassword = "password"
+	espFieldAPIKey       = "api_key"
+)
+
+func ProcessorSpec() *service.ConfigSpec {
+	return service.NewConfigSpec().
+		Categories("Integration").
+		Summary("Queries Elasticsearch and enriches messages with the result.").
```
jem-davies:
```suggestion
		Summary("Queries Elasticsearch and replaces messages with the result.").
```

I see that in a number of places we refer to this processor potentially enriching messages - but that behaviour requires the use of a branch process, as you point out?

internal/impl/elasticsearch/processor.go
```diff
@@ -0,0 +1,543 @@
...
+	espFieldTLS          = "tls"
+	espFieldAuth         = "basic_auth"
+	espFieldAuthEnabled  = "enabled"
+	espFieldAuthUsername = "username"
+	espFieldAuthPassword = "password"
+	espFieldAPIKey       = "api_key"
+)
+
+func ProcessorSpec() *service.ConfigSpec {
+	return service.NewConfigSpec().
```
jem-davies, line 33: We could add a call to .[LintRule](https://pkg.go.dev/github.com/warpstreamlabs/bento/v4/public/service#ConfigField.LintRule) here; and catch instances of having the `action` set to 'delete' without and `id` field etc.

internal/impl/elasticsearch/processor.go
```diff
@@ -0,0 +1,543 @@
...
+`).
+		Fields(
+			service.NewStringListField(espFieldURLs).
+				Description("A list of URLs to connect to. If an item of the list contains commas it will be expanded into multiple URLs.").
+				Example([]string{"http://localhost:9200"}),
+			service.NewStringEnumField(espFieldAction, "search", "get", "delete").
+				Description("The operation to perform against Elasticsearch.").
+				Default("search"),
+			service.NewInterpolatedStringField(espFieldIndex).
+				Description("The index to query. Supports interpolation so the index can be derived from each message."),
```
jem-davies: The doc generation will append a mention that it supports interpolation, having it here will mean it will be in the description twice.

internal/impl/elasticsearch/processor.go
```diff
@@ -0,0 +1,543 @@
...
+
+	argsMapping *bloblang.Executor
+}
+
+func esProcessorConfigFromParsed(pConf *service.ParsedConfig) (conf esProcessorConfig, err error) {
+	var rawURLs []string
+	if rawURLs, err = pConf.FieldStringList(espFieldURLs); err != nil {
+		return
+	}
+	for _, u := range rawURLs {
```
jem-davies: We don't have this parsing of the URLs in elasticsearch_v2 output so, either this is redundant or the elasticsearch_v2 output needs something similar. 

I suspect we might need to fix the elasticsearch_v2 output.

internal/impl/elasticsearch/processor.go
```diff
@@ -0,0 +1,543 @@
...
+
+	p.client = client
+	p.log.Debugf("Successfully connected to Elasticsearch")
+	return nil
+}
+
+func (p *EsProcessor) ProcessBatch(ctx context.Context, batch service.MessageBatch) ([]service.MessageBatch, error) {
+
+	if p.client == nil {
+		if err := p.Connect(ctx); err != nil {
```
jem-davies: You can return a service.ErrNotConnected here - rather than calling p.Connect - service.ErrNotConnected will trigger Bento to go into a Connect loop

internal/impl/elasticsearch/processor.go
```diff
@@ -0,0 +1,543 @@
...
+	}
+
+	msg.MetaSetMut("es_index", index)
+	msg.MetaSetMut("es_id", id)
+	msg.MetaSetMut("es_delete_result", response.Result)
+
+	return nil
+}
+
+func extractSearchHitsAndMeta(msg *service.Message, index string, body []byte) error {
```
jem-davies, line 461: Would want a section in the docs detailing the metadata we are adding to the message.

internal/impl/elasticsearch/processor.go
```diff
@@ -0,0 +1,543 @@
...
+	if p.conf.argsMapping != nil {
+		argsExecutor = batch.BloblangExecutor(p.conf.argsMapping)
+	}
+
+	batch = batch.Copy()
+
+	for i, msg := range batch {
+		index, err := indexExecutor.TryString(i)
+		if err != nil {
+			p.log.Debugf("Index interpolation error: %v", err)
```
jem-davies: I don't think we should have these Debug logs - they should already be logged when we return the error - and end up as an error log.

jem-davies (changes requested):
Nice, thanks for this PR 🙏  

I have had a quick look through and made some comments that pertain to Bento idioms such as returning service.ErrNotConnected etc. - would have to take another closer look at the implementation again.

internal/impl/elasticsearch/processor.go
Omar1H1: @jem-davies first of all thank you for taking the time to review my PR and sorry if i am not that familiar with code base
something like this should be the expected output using LintRule at the spec-level !! : 
image

internal/impl/elasticsearch/processor.go
Omar1H1: If i am not wrong the bug is in output_v2, its field description says "Description("A list of URLs to connect to. If an item of the list contains commas it will be expanded into multiple URLs.")."  but the parsing code never does that split. FieldStringList just returns items as-is.
This breaks the common pattern of passing multiple URLs through a single env var like ES_URLS="http://es1:9200,http://es2:9200"
the ES client gets one malformed address and fails. I confirmed it with a docker compose test.
So I'd rather keep the split in my processor and fix output_v2 to match maybe ....
want me to include that in this PR? 
It comes down to using FieldURLList which actually do what i did here but better

``` go
func urlsFromStr(str string) (urls []*url.URL, err error) {
	for s := range strings.SplitSeq(str, ",") {
		if s = strings.TrimSpace(s); s == "" {
			continue
		}
		var u *url.URL
		if u, err = url.Parse(s); err != nil {
			return
		}
		urls = append(urls, u)
	}
	return
}
```

i almost got tricked by output V2
image

Omar1H1:
@jem-davies, finally I guess it's a better approach to patch the bug in **output_v2** _(esoV2FieldURLs)_ in a separate PR !!

internal/impl/elasticsearch/processor.go
```diff
@@ -0,0 +1,566 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return &EsProcessor{
+		log:  mgr.Logger(),
+		conf: conf,
+	}, nil
+}
+
+func (p *EsProcessor) Connect(ctx context.Context) error {
```
jem-davies: Connect isn't a part of the [BatchProcessor](https://pkg.go.dev/github.com/warpstreamlabs/bento/v4/public/service#BatchProcessor) interface like it is for the Output - we can just move the creation of the client to the NewProcessorFromConfig function?

internal/impl/elasticsearch/processor.go
```diff
@@ -0,0 +1,566 @@
...
+	p.client = client
+	p.log.Debugf("Successfully connected to Elasticsearch")
+	return nil
+}
+
+func (p *EsProcessor) ProcessBatch(ctx context.Context, batch service.MessageBatch) ([]service.MessageBatch, error) {
+
+	if p.client == nil {
+		return nil, service.ErrNotConnected
+	}
```
jem-davies:
```suggestion
```

Not relevant to check this after moving the creation of the client.

internal/impl/elasticsearch/processor.go
```diff
@@ -0,0 +1,566 @@
...
+		conf: conf,
+	}, nil
+}
+
+func (p *EsProcessor) Connect(ctx context.Context) error {
+	if p.client != nil {
+		return nil
+	}
+
+	p.log.Debugf("Connecting to Elasticsearch at %v", p.conf.clientConfig.Addresses)
```
jem-davies: I think that these log.debug calls won't really add much, we will either Connect successfully to ES and get an INFO log from Bento engine or we will fail at some point and get an Error returned.

internal/impl/elasticsearch/processor.go
```diff
@@ -0,0 +1,566 @@
...
+      "must": [
+        { "match": { "status": this.status } },
+        { "range": { "created_at": { "gte": this.since } } }
+      ]
+    }
+  },
+  "size": 10
+}`).
+				Optional(),
+			service.NewDurationField(espFieldTimeout).
```
jem-davies: We are currently only using this field for the init connection test - I think we should either add it to all the requests - or just remove it.

internal/impl/elasticsearch/processor.go
```diff
@@ -0,0 +1,566 @@
...
+	if rawURLs, err = pConf.FieldStringList(espFieldURLs); err != nil {
+		return
+	}
+	for _, u := range rawURLs {
+		for part := range strings.SplitSeq(u, ",") {
+			if part = strings.TrimSpace(part); part != "" {
+				conf.clientConfig.Addresses = append(conf.clientConfig.Addresses, part)
+			}
+		}
+	}
```
jem-davies: I think we should remove this feature of being able to have multiple urls in one entry on the list -- should just be able to add it as a separate item. 

Yeah it's a bug with the output but can just update the docs for that .

internal/impl/elasticsearch/processor.go
```diff
@@ -0,0 +1,566 @@
...
+	defer cancel()
+
+	res, err := client.Info(client.Info.WithContext(ctx))
+	if err != nil {
+		p.log.Debugf("Failed to connect to ES: %v", err)
+		return fmt.Errorf("failed to connect to elasticsearch: %w", err)
+	}
+	defer res.Body.Close()
+	if res.IsError() {
+		p.log.Debugf("ES returned error: %s", res.Status())
```
jem-davies: I don't think we should check the client's connection before because with a processor you won't get a retry mechanism like you do with an output that fails to connect.

jem-davies (dismissed):
Connect isn't a part of a processors interface - so as-is - it's not working. It's a small fix to move the client creation to processor creation 👍

Omar1H1:
> Connect isn't a part of a processors interface - so as-is - it's not working. It's a small fix to move the client creation to processor creation 👍

Thanks you for the feedback, I made the necessary changes. For now, I've opted to document the URL issue by adding a warning to the documentation.

jem-davies:
@Omar1H1  - thanks a lot! 🙇

PR #887, opened 2026-06-14
feat(gcp): add custom endpoint configuration to gcp_bigquery output

Shareed2k:
## Description
This PR introduces the `endpoint` configuration option to the `gcp_bigquery` output component. 

## Motivation and Context
Previously, the component hardcoded the default BigQuery API endpoint for production usage. Adding a configurable `endpoint` enables developers to route traffic to local BigQuery emulators during testing and development. It also unlocks support for routing through proxies or custom VPC Service Controls endpoints, aligning this component with other GCP components (like Pub/Sub and BigQuery Storage) that already support custom endpoints.

### DIFF
--- a/internal/impl/gcp/output_bigquery.go
+++ b/internal/impl/gcp/output_bigquery.go
@@ -314,7 +314,7 @@ func (g *gcpBigQueryOutput) Connect(ctx context.Context) (err error) {
 	var client *bigquery.Client
 	var opts []option.ClientOption
 	if g.conf.Endpoint != "" {
-		opts = append(opts, option.WithEndpoint(g.conf.Endpoint))
+		opts = append(opts, option.WithoutAuthentication(), option.WithEndpoint(g.conf.Endpoint))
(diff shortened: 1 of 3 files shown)


gregfurman (commented):
Thanks for opening! Is this not addressed by setting the  BIGQUERY_EMULATOR_HOST variable?

I believe this is automatically picked up by the BQ SDK client and should reroute traffic to the specified host.

Shareed2k:
> Thanks for opening! Is this not addressed by setting the BIGQUERY_EMULATOR_HOST variable?
> 
> I believe this is automatically picked up by the BQ SDK client and should reroute traffic to the specified host.

yes but what about `WithoutAuthentication` ?

gregfurman (approved):
LGTM! Seems the aforementioned environment variable is only supported in the python and typescript SDKs 🤷

gregfurman:
Also, validated this via the integration test locally.

PR #895, opened 2026-06-24
fix(aws_s3_stream): guarantee message delivery

maxtheaxe:
Due to the way input messages were being acked (immediately after writer load of content), a crash part-way through would have caused an orphaned upload, as [s3 multipart uploads require explicit completion](https://docs.aws.amazon.com/AmazonS3/latest/API/API_CompleteMultipartUpload.html). In such scenarios, on restart, as-yet unprocessed/unacked input messages would have started a new upload, recreating the file. Thus, pre-crash messages were effectively dropped.

This is addressed by the following, at a high level:

* delayed ack of the input messages (by batch) until after completion of s3 upload
* added support for resuming incomplete multipart uploads

Also re-labels `aws_s3_stream` as `experimental` (was previously `beta`), to more accurately reflect the status.

Initially reported [on slack](https://warpstreamlab-uwt2030.slack.com/archives/C0768UE3NEM/p1782242888039349).


triddell:
Thanks for this — the delayed-ack + resume work is a real improvement over what
I had in #893, and the unit coverage for the recovery paths is solid.

While testing it against a real S3 bucket I hit a shutdown deadlock for streams
that stay under the 5 MiB multipart threshold (i.e. small / low-throughput
outputs — the same case #893 targeted). The object is written correctly, but
bento never terminates on a finite/bounded input; it only completes the upload
and exits when it receives a SIGTERM/SIGINT.

**Root cause (as I read it):** messages are now acked only after their bytes are
durable in S3. For a sub-5 MiB stream nothing seals mid-stream, so the acks are
resolved only in `S3StreamingWriter.Close()`. But the output's `loop()` only
calls `Close()` when the input's transaction channel closes — and a finite input
won't close that channel until its in-flight transactions are acked. Input waits
for acks → output won't ack until `Close()` → `Close()` only runs after the input
finishes → deadlock. On `main` the output acked on write, so the input drains and
bento exits cleanly, which makes this a regression for bounded inputs.

**Reproduction** (no batching, single small object well under 5 MiB):

```yaml
# small-nobatch.yaml
input:
  generate:
    count: 3
    interval: ""
    mapping: 'root = "small nobatch line %v".format(count("nb"))'
output:
  aws_s3_stream:
    bucket: "${S3STREAMTEST_BUCKET}"
    region: "${AWS_REGION}"
    path: "s3streamtest/small-nobatch/${RUNID}.txt"
    max_buffer_bytes: 5242880
    max_buffer_period: 2s
```

```
# this PR (fix/delay-ack-s3-stream): hangs; only closes on SIGTERM
$ gtimeout 20 ./bento -c small-nobatch.yaml ; echo "exit=$?"
... INFO Received SIGTERM, the service is closing
exit=124          # object IS written, but only via the forced stop

# main: same config exits cleanly
$ gtimeout 20 ./bento-main -c small-nobatch.yaml ; echo "exit=$?"
exit=<FILL IN: expected 0>
```

A stdout output with the identical input exits cleanly (`exit=0`), so it's
specific to this output, not the generate/shutdown lifecycle or batching.

**Possible directions** (whichever fits your design):
- Resolve a small final part's acks proactively rather than only in `Close()`
  (e.g. when the input signals end-of-input), so the input can drain; or
- Have the output detect input-channel closure and run `Close()` before the
  input blocks on acks (break the ordering), or
- Fall back to `PutObject` for a single sub-5 MiB final part (what #893 did),
  which sidesteps holding the only ack until `Close()`.

Happy to test any of these in my environment — I'm running this output in prod,
so I can validate against real S3 quickly. Also worth a dedicated test for the
empty-file path: this PR completes a multipart upload with a single zero-byte
part, which I'd want to confirm against real S3 (vs. mocks).
```

**Comment by maxtheaxe** [comment]
@triddell hey, thanks—ironically I ran into the same thing last night and will work on a patch tomorrow

**Comment by triddell** [comment]
## Use case: bento embedded in AWS Lambda (bounded input), where this output's shutdown behavior is decisive

Adding a concrete real-world use case, because I think it shows why #893's ack-on-write fix has
standalone value independent of the broader delayed-ack work here — and I want to ask whether this PR
can be made to cover it.

We run bento **embedded inside an AWS Lambda**, SQS-triggered. Each invocation pushes a **finite batch**
of records through `aws_s3_stream` and then must terminate on its own. It's a bounded input, not a
long-running stream, and there's no operator to send SIGTERM — if the process doesn't self-terminate,
Lambda just hits its timeout (300–900 s), the batch is redelivered, and we pay for the wall-clock. Many
of our flushes are also well under S3's 5 MiB multipart minimum (low-volume log sources), so the
`EntityTooSmall` path is routine, not an edge case.

Testing the delayed-ack approach against **real S3** (I tested #895; I believe #897 carries the same
ack model — please correct me), we hit a **deadlock on bounded inputs**: acks resolve only in `Close()`,
but the output's run loop only calls `Close()` after the input's transaction channel closes, and a
finite input won't close that channel until its in-flight transactions are acked. Nothing finalizes
until an external SIGTERM forces `Close()`. On a long-running stream you never notice; on a
bounded/embedded input it's a hang, and it's worst precisely for the low-throughput / sub-5 MiB case.
(The deadlock only reproduces through a full input→output→shutdown loop — not unit tests or a
direct-writer CLI that calls `Close()` itself. I have a real-S3 harness and can share/run it here.)

#893's ack-on-write + `PutObject` fallback for sub-5 MiB/empty objects lets the input drain and the
process exit 0 — which is what embedded/bounded deployments need today. That's the standalone value:
it's the minimal, low-risk fix while the larger rewrite matures.

**Can this PR work for our use case?** I think yes — *if* it treats **end-of-input as a finalize
trigger**: flush and resolve the acks for the final (possibly sub-5 MiB, possibly empty) part without
waiting for a signal-driven `Close()`, with a `PutObject` fallback for a single small/zero-byte final
part (since `CompleteMultipartUpload` rejects a sole part under 5 MiB). With that, it'd be a clean
superset of #893 and we'd happily switch. Without it, it can't terminate a bounded input regardless of
its (genuinely valuable) at-least-once and crash-recovery improvements.

Happy to run this branch through the real-S3 harness and report back.

**Comment by maxtheaxe** [comment]
@triddell re: self-termination, we have a similar use case, though ours is a long-running job in our kubernetes cluster. I've been using the [**read_until**](https://warpstreamlabs.github.io/bento/docs/components/inputs/read_until/) component (with `idle_timeout` configured) to wrap my input (in this case, [**redis_streams**](https://warpstreamlabs.github.io/bento/docs/components/inputs/redis_streams/)) and achieve effectively the same thing.

I can add that to the docs as a known pattern, although it lacks support for the ability to time out files on an individual level. This means that a pipeline writing files A, B, and C dynamically via a single input (e.g. **redis_streams**) will not time out A, B, or C until _all_ inputs via **redis_streams** stop for the configured `idle_timeout`. This doesn't necessarily break anything on my end (and doesn't require that we keep the entire files in memory or anything), but does mean that it ends up doing `CompleteMultipartUpload` for every file at the same time.

I would obviously prefer integrated, output path-aware timeouts, but I'm okay with this limitation at the moment. Once I fix the sub-5MB issue, let's try to get this reviewed and think about more advanced timeouts in a future development?

**Comment by triddell** [comment]
Really like the direction here — at-least-once + interrupted-upload recovery are genuinely valuable. I want to lay out, with the bento internals, why I think #893's simpler fix is needed as the **base behavior**, and why these two can't be a single code path without an explicit opt-in flag — so we can sequence them rather than treat them as competing.

**#893 is the behavior the output needs by default.** The intuitive contract of `aws_s3_stream` — what you'd expect running bento as a CLI — is "stream a (possibly large) file to S3 without holding it all in memory, and **stop when the input is done**." That works today because acks resolve on write: at input EOF the acks are already back, the input closes its transaction channel, the output finalizes (incl. a `PutObject` for a small/final object), and the process exits. #893 only makes the sub-5 MiB / empty case valid without changing that.

**Why delayed-ack breaks it, structurally.** In `internal/component/input/async_reader.go`, the read loop registers two deferred calls:

```go
defer func() { ...; close(r.transactions); ... }()  // registered first
defer func() { pendingAcks.Wait() }()                // registered second → runs FIRST (LIFO)
```

So on input EOF the input runs `pendingAcks.Wait()` **before** closing its transaction channel — it won't close the channel until every delivered message is acked. And the only end-of-input signal an output gets (absent an external SIGINT/manager stop) is that channel closing. This PR defers acks until the tail is finalized in `Close()`, and the component only calls `Close()` when the transaction channel closes. So for any bounded, ack-gated input (`generate`, `file`, `read_until` with a `check`, `drop_on`, single-in-flight) you get a circular wait: the input won't close the channel until acks resolve → `Close()` never runs → the tail ack never fires → deadlock; it only terminates on a hard signal.

The one escape is exactly what this PR documents: `read_until` with `idle_timeout`, whose loop (`internal/impl/pure/input_read_until.go`) closes its own transaction channel **without** a `pendingAcks.Wait()` gate. That confirms the mechanism — delayed-ack (don't-ack-until-durable) and finalize-on-EOF are fundamentally incompatible for bounded inputs, and no output-side change (timer or otherwise) can reconcile them, because the EOF signal itself is ack-gated.

**One heads-up on coverage:** `TestS3StreamOutputBoundedInputSmallObjectFinalizesAndAcks` closes the transaction channel unconditionally, which doesn't model `async_reader`'s Wait-before-close ordering — so it passes even though a real bounded input deadlocks. A faithful version (spawn the ack-waiting goroutines and `pendingAcks.Wait()` before `close(txChan)`) would hang against this branch. Worth adding, since it's the exact gap.

**Suggested way forward:** land #893 first as the minimal `EntityTooSmall` fix — it preserves the ack-on-write / finalize-on-EOF behavior that CLI and bounded inputs require. Then this PR (or a follow-up) adds the durable-before-ack, at-least-once + recovery behavior as an **explicit opt-in** (a config flag / delivery-guarantee mode), documented as requiring continuous flow or `read_until.idle_timeout` for bounded inputs. Both share #893's small-file `PutObject` finalize, so neither mode regresses the other. Happy to help with the flag/plumbing — I have a real-S3 harness that exercises both the bounded-finalize and the resume paths.

gregfurman:
@maxtheaxe Just want to confirm: is this ready for review?

maxtheaxe:
@gregfurman I'm going to move this back to draft—I need to spend some more time thinking about the best way to make sure there's coverage for @triddell's workflow and others like it. I'm not sure I agree that

> The intuitive contract of aws_s3_stream — what you'd expect running bento as a CLI — is "stream a (possibly large) file to S3 without holding it all in memory, and stop when the input is done."

if that comes at the cost of nonexistent delivery guarantees, but I will think about it when I have more time.

triddell:
@gregfurman @maxtheaxe I made another commit on #893 and I have been running it in prod now with very good success. It solves the real, existing bug with the small file problem ( < 5MB initially, so use a PutObject and don't stream) and now it handles gzip compression correctly over the entire file. It also will create multiple output files to s3 based on event criteria. I think from my comments on this PR and on #893, that it's the solution needed for a "bounded" set of data... like processing a file from the CLI and having the results write to S3 in multi-part streaming fashion (keeping memory low, etc.) I think additional behavior/guarantees would need to be behind a flag (and that would work in Bento streams mode I assume).

So, since there is a current bug with small files on aws_s3_stream, I think #893 should be considered at this point.

Here is an example from one of my configs that uses it:

```
output:
  aws_s3_stream:
    bucket: your_bucket_name
    path: 'task/anthropic-compliance/activities/${ANTHROPIC_ORGANIZATION_ID}/${! env("WINDOW_START").ts_parse("2006-01-02T15:04:05Z07:00").ts_format("2006/01/02/15", "UTC") }/${! uuid_v4() }.jsonl.gz'
    region: us-west-2
    content_type: application/json
    compression: gzip
    partition_by:
      - '${ANTHROPIC_ORGANIZATION_ID}'
    max_buffer_bytes: 10485760
    batching:
      period: 10s
```

Please let me know what you think or anything else you like to see on #893.

PR #899, opened 2026-06-25
send_ack Option in NATS Core Input

ahart97:
## Overview

This PR adds support for the additional NATS input boolean parameter `send_ack`.

In [NATS core](https://docs.nats.io/nats-concepts/core-nats/reqreply), message acknowledgements are not natively supported, they are more often associated with [NATS JetStream](https://docs.nats.io/nats-concepts/jetstream/streams). This parameter (`send_ack`) allows the user to turn off acknowledgements when using the NATS core input component within Bento.

This parameter is already supported by one of the alternatives to Bento, [RedPanda Connect ](https://docs.redpanda.com/connect/components/inputs/nats/#send_ack).

Despite acknowledgements not being native to NATS core, the default of the `send_ack` parameter was set to `True` to align with previous versions of Bento if left out of the configuration.

## Contribution Details

Have ran the following commands and everything completed with no errors:

-`make fmt`
-`make docs`
-`golangci-lint run ./internal/impl/nats/...`
-`go test -timeout 3m ./internal/impl/nats/...`
-`go test -run 'TestIntegration' -timeout 5m ./internal/impl/nats/...`

### DIFF
--- a/internal/impl/nats/input.go
+++ b/internal/impl/nats/input.go
@@ -69,7 +69,8 @@ output:
 		Field(service.NewBoolField("send_ack").
 			Description("Whether to send an acknowledgement back to the input subject's reply address after consuming a message.").
 			Default(true).
-			Advanced()).
+			Advanced().
+			Version("1.19.0")).
 		Fields(connectionTailFields()...).
 		Fields(inputTracingDocs()...)
 }
--- a/internal/impl/nats/input.go
+++ b/internal/impl/nats/input.go
@@ -66,6 +66,10 @@ output:
 			Advanced().
 			Default(nats.DefaultSubPendingMsgsLimit).
 			LintRule(`root = if this < 0 { ["prefetch count must be greater than or equal to zero"] }`)).
+		Field(service.NewBoolField("send_ack").
+			Description("Whether to send an acknowledgement back to the input subject's reply address after consuming a message.").
+			Default(true).
+			Advanced()).
 		Fields(connectionTailFields()...).
 		Fields(inputTracingDocs()...)
 }
@@ -97,6 +101,7 @@ type natsReader struct {
 	queue         string
 	prefetchCount int
 	nakDelay      time.Duration
+	sendAck       bool
 
 	log *service.Logger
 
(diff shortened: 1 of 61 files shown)


internal/impl/nats/input.go
```diff
@@ -66,6 +66,10 @@ output:
 			Advanced().
 			Default(nats.DefaultSubPendingMsgsLimit).
 			LintRule(`root = if this < 0 { ["prefetch count must be greater than or equal to zero"] }`)).
+		Field(service.NewBoolField("send_ack").
+			Description("Whether to send an acknowledgement back to the input subject's reply address after consuming a message.").
+			Default(true).
+			Advanced()).
```
jem-davies:
```suggestion
			Advanced().
			Version("1.19.0")).
```

internal/impl/nats/input.go
```diff
@@ -66,6 +66,10 @@ output:
 			Advanced().
 			Default(nats.DefaultSubPendingMsgsLimit).
 			LintRule(`root = if this < 0 { ["prefetch count must be greater than or equal to zero"] }`)).
+		Field(service.NewBoolField("send_ack").
+			Description("Whether to send an acknowledgement back to the input subject's reply address after consuming a message.").
```
jem-davies: Isn't this a bit inaccurate? - The acknowledgement will only be sent after it has been succesfully delivered not when it is consumed?

jem-davies (dismissed):

internal/impl/nats/input.go
ahart97: Thanks for the feedback on this!

Wrote the wording here from the server/listener side (i.e., Bento service receiving the incoming NATS message) instead of the client/publisher side (i.e., the users service triggering the input).

I see the wording "delivered" and "consumed" as two sides of the same coin, with "delivered" being the event from the client/publisher side and "consumed" being that same event from the server/listener side.

Very much open to changing this if the overall documentation is from the perspective of the client using Bento rather than Bento itself.

jem-davies: The difference which I am referring between 'delivered' & 'consumed' is that: 

 - consumed: bento reads a message from the source 
 - delivered: bento successfully sends a message to the sink

not a difference of the perspective between client & server. 

This sendAck func will only trigger if the message has been delivered by Bento - that is why it's inaccurate as-is.

PR #901, opened 2026-06-26
fix(nats): improve resilience of nats_kv components against JetStream…

### LINKED ISSUE #900: fix(nats): improve resilience of nats_kv components against JetStream stalls
The nats_kv components (cache_kv, output_kv) can stall or hang indefinitely when NATS JetStream is slow to respond or the underlying TCP connection is lost:

 1. cache_kv deadlock on JetStream stall – connect() holds connMut.Lock() while calling js.KeyValue(bucket) from the deprecated nats.JetStreamContext API, which has no context/timeout parameter. Any delay in JetStream causes all callers to block indefinitely.
 2. Stale connection not detected – When the TCP connection drops, cache_kv has no mechanism to proactively clear cached handles, so subsequent operations fail on a stale connection before reconnecting.
 3. output_kv leaks stale JetStream handle – On a Put failure, output_kv does not call disconnect(), leaving the stale handle in place and preventing a clean reconnect on the next retry.
 4. Slow dead-connection detection – The default PingInterval of 2 minutes means dead connections aren't detected for up to 4 minutes. MaxReconnects is also not set to unlimited, so components can give up reconnecting during extended outages.

henrikschristensen:
… stalls

- cache_kv: migrate from deprecated nats.JetStreamContext to the new jetstream package so all KV operations pass a context. Previously connect() held connMut.Lock() while calling js.KeyValue(bucket) which has no context parameter, causing all callers to block indefinitely when JetStream was slow to respond. Adds getKV() reconnect helper and disconnect() calls on operation errors, consistent with processor_kv and output_kv.

- cache_kv: add a DisconnectErrHandler that proactively clears the cached natsConn and kv handles when the underlying TCP connection is lost, so the next operation calls connect() with fresh handles rather than failing on a stale connection first.

- output_kv: call disconnect() when Put fails so the stale JetStream handle is released and Connect() is called on the next retry, matching the behaviour of cache_kv and processor_kv.

- connection: set PingInterval=30s (was 2m) so dead connections are detected within ~60s rather than up to 4 minutes, and MaxReconnects=-1 for unlimited reconnect attempts so components recover automatically from extended outages.

Fixes #900


internal/impl/nats/integration_kv_test.go
```diff
@@ -392,3 +392,98 @@ cache_resources:
...
+	require.Eventually(t, func() bool {
+		cache.connMut.RLock()
+		defer cache.connMut.RUnlock()
+		return cache.natsConn == nil
+	}, 10*time.Second, 50*time.Millisecond,
+		"DisconnectErrHandler should have cleared natsConn after server stop")
+
+	// Restart the container — Docker reuses the original port bindings when
+	// no HostConfig is specified.
+	require.NoError(t, pool.Client.StartContainer(resource.Container.ID, nil))
```
jem-davies: I noticed that when I was running this - the ephemeral port on my local machine changed - and the subsequent wait for NATS to accept connections again always failed.

jem-davies (changes requested):

jem-davies (dismissed):

jem-davies:
bit baffled why the CI is failing

henrikschristensen:
@jem-davies I pinned the docker port and merged with main for a green CI

PR #905, opened 2026-06-29
chore(sentry): Update sentry to v0.47.0

gregfurman:
## Motivation

Updates sentry to the latest minor version `v0.47.0`, which introduces the following:

- Closing the entire `Transport` via a `Close` method
- Flushing a `Transport` with context using `FlushWithContext`

This should also unblock dependabot since this sentry upgrading is causing build + lint failures https://github.com/warpstreamlabs/bento/pull/892/

## Changes

- The processor now closes a `Transport` if `FlushWithContext` occurs within the ctx deadline -- in accordance with the `flush_timeout` duration value.
- Updates the mocks to now include `Close` and `FlushWithContext` instead of just `Flush`

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -70,7 +70,7 @@ require (
 	github.com/fatih/color v1.18.0
 	github.com/fsnotify/fsnotify v1.9.0
 	github.com/generikvault/gvalstrings v0.0.0-20180926130504-471f38f0112a
-	github.com/getsentry/sentry-go v0.27.0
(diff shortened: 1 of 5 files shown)


gregfurman:
@jem-davies OK should be ready now. No idea how I missed that `make deps` part 😅 This'll unblock our dependabot pipeline 🙏

jem-davies (approved):

PR #909, opened 2026-07-04
fix(aws_dynamodb): store JSON numbers as Number (N), not String (S)

### LINKED ISSUE #908: aws_dynamodb output stores JSON numbers as String (S) instead of Number (N)
# `aws_dynamodb` output stores JSON numbers as String (S) instead of Number (N)

### Version
main (and 1.18.x)

### Summary
The `aws_dynamodb` output writes every numeric field of a structured message to DynamoDB as a String (`S`) attribute instead of a Number (`N`).

`anyToAttributeValue` (`internal/impl/aws/output_dynamodb.go`) has correct `float64`/`int`/`int64` branches that produce `AttributeValueMemberN` — but Bento decodes structured JSON bodies with `UseNumber()`, so numeric leaves arrive as `json.Number`, and the `case json.Number` branch returned `AttributeValueMemberS`. Since `json.Number` matches first, the numeric branches are dead for any JSON-sourced message: every number, at every depth, is written as a string.

### Steps to reproduce
```sh
echo '{"id":"sku-1","price":42.8,"inventory":98}' | bento -c ./config.yaml
```
```yaml
# config.yaml
input:
  stdin: {}
output:
  aws_dynamodb:
    region: us-east-1
    json_map_columns:
      "": "."
```
(No processors — a bloblang `mapping` that assigns numeric literals produces native `float64`/`int64` and would mask the bug.)

### Expected
`price` and `inventory` stored as `N`.

### Actual
```json
{ "id": {"S":"sku-1"}, "price": {"S":"42.8"}, "inventory": {"S":"98"} }
```
Numeric attributes stored as `S`. This silently breaks numeric key conditions / range comparisons (lexicographic on strings) and any reader unmarshalling into a numeric field.

### Fix
`case json.Number` should return `AttributeValueMemberN{Value: v.String()}` (precision-preserving; DynamoDB `N` is string-on-the-wire). PR attached.

cthorner:
# fix(aws_dynamodb): store JSON numbers as Number (N), not String (S)

Fixes #908

## Summary

The `aws_dynamodb` output writes every numeric field in a structured message to DynamoDB as a String (`S`) attribute instead of a Number (`N`). The fix is a one-line change in `anyToAttributeValue`, plus a table-driven test covering the conversion.

## The bug

`anyToAttributeValue` in `internal/impl/aws/output_dynamodb.go` converts a decoded message value into a `types.AttributeValue`. It has correct branches for `float64`, `int`, and `int64` that all produce `AttributeValueMemberN` — but for messages sourced from JSON, those branches are never reached.

When Bento parses a structured JSON body it decodes numbers into `json.Number` (the decoder uses `UseNumber()` semantics), not `float64`. So `json_map_columns` values, and any nested numeric leaves, arrive at `anyToAttributeValue` as `json.Number`. The `case json.Number` branch returned an `S`:

```go
case json.Number:
    return &types.AttributeValueMemberS{   // wrong: this is a number
        Value: v.String(),
    }
```

Because `json.Number` is matched first, the `float64` / `int` / `int64` cases below it are effectively dead code for any JSON-derived message — every number, at every depth, was silently written as a string.

## Impact

Every numeric field — prices, quantities, counts, inventory levels, timestamps — lands in DynamoDB as `S` instead of `N`. Concretely this breaks:

- **KeyConditionExpression / range keys** on numeric attributes — `price > :p` and `BETWEEN` comparisons compare lexicographically against a string, so `"98" < "9"` and `"42.8"` sorts wrong.
- **Comparisons and filters** in queries/scans against those attributes.
- **Downstream readers** that unmarshal into a numeric Go/typed field (`attributevalue.UnmarshalMap` into an `int`/`float64`) — they fail or need per-field workarounds.

It's also silent: the write succeeds, the value round-trips as text, and the wrong type only surfaces later at query time or in a consumer.

## The fix

```go
case json.Number:
    return &types.AttributeValueMemberN{
        Value: v.String(),
    }
```

`v.String()` returns the original literal exactly as it appeared in the source JSON. This is precision-preserving: DynamoDB's `N` type is transmitted as a decimal string on the wire (the SDK's `AttributeValueMemberN.Value` is a `string`), so `"42.8"` and large integers are sent verbatim with no float64 round-trip or `strconv` reformatting. This is strictly better than routing through the `float64` branch, which would risk representational drift. `S` remains correct for genuine `string` values, so no existing string behavior changes.

## Test

Added `TestAnyToAttributeValue`, a parallel table-driven test asserting the conversion for each relevant input type:

- `json.Number("98")` → `N{"98"}` (integer)
- `json.Number("42.8")` → `N{"42.8"}` (float, precision preserved)
- `"hello"` → `S{"hello"}` (strings still map to `S`)
- `true` → `BOOL{true}`
- `map[string]any{"qty": json.Number("5")}` → `M{ qty: N{"5"} }` (numeric leaves inside nested maps also map to `N`)

## Minimal repro

The message must arrive as **raw JSON** so that it is parsed with `UseNumber()` and its numeric leaves become `json.Number` (a bloblang `mapping` that assigns numeric literals would produce native `float64`/`int64`, which already hit the correct `N` branches — so this bug only shows on JSON-sourced messages that are written through unchanged):

```sh
echo '{"id":"sku-1","price":42.8,"inventory":98}' | bento -c ./config.yaml
```

```yaml
# config.yaml
input:
  stdin: {}

# no processors — keep the parsed json.Number leaves intact

output:
  aws_dynamodb:
    region: us-east-1
    json_map_columns:
      "": "."            # write every top-level field as an item attribute
```

**Before** (numbers stored as `S`):

```json
{
  "id":        { "S": "sku-1" },
  "price":     { "S": "42.8" },
  "inventory": { "S": "98" }
}
```

**After** (numbers stored as `N`):

```json
{
  "id":        { "S": "sku-1" },
  "price":     { "N": "42.8" },
  "inventory": { "N": "98" }
}
```

Strings still map to `S`; only JSON numbers change from `S` to `N`.

### DIFF
--- a/internal/impl/aws/output_dynamodb.go
+++ b/internal/impl/aws/output_dynamodb.go
@@ -327,7 +327,7 @@ func anyToAttributeValue(root any) types.AttributeValue {
 			Value: v,
 		}
 	case json.Number:
-		return &types.AttributeValueMemberS{
+		return &types.AttributeValueMemberN{
 			Value: v.String(),
 		}
 	case float64:
--- a/internal/impl/aws/output_dynamodb.go
+++ b/internal/impl/aws/output_dynamodb.go
@@ -35,6 +35,10 @@ const (
 	ddboFieldDeletePartitionKey = "partition_key"
 	ddboFieldDeleteSortKey      = "sort_key"
 	ddboFieldBatching           = "batching"
+	ddboFieldJSONNumberType     = "json_number_type"
+
+	ddboJSONNumberTypeString = "string"
+	ddboJSONNumberTypeNumber = "number"
 )
 
 type ddboConfig struct {
@@ -46,6 +50,7 @@ type ddboConfig struct {
 	DeleteConditionExec     *bloblang.Executor
 	PartitionKeyDeleteField string
 	SortKeyDeleteField      string
+	NumbersAsN              bool
 
 	aconf       aws.Config
 	backoffCtor func() backoff.BackOff
@@ -67,6 +72,11 @@ func ddboConfigFromParsed(pConf *service.ParsedConfig) (conf ddboConfig, err err
 	if conf.TTLKey, err = pConf.FieldString(ddboFieldTTLKey); err != nil {
 		return
 	}
+	var jsonNumberType string
+	if jsonNumberType, err = pConf.FieldString(ddboFieldJSONNumberType); err != nil {
+		return
+	}
+	conf.NumbersAsN = jsonNumberType == ddboJSONNumberTypeNumber
 	if conf.aconf, err = GetSession(context.TODO(), pConf); err != nil {
 		return
 	}
@@ -164,6 +174,10 @@ This output benefits from sending messages as a batch for improved performance. 
 			service.NewStringField(ddboFieldTTLKey).
 				Description("The column key to place the TTL value within.").
 				Default("").
+				Advanced(),
+			service.NewStringEnumField(ddboFieldJSONNumberType, ddboJSONNumberTypeString, ddboJSONNumberTypeNumber).
+				Description("Controls how JSON numbers extracted via `"+ddboFieldJSONMapColumns+"` are stored. When set to `"+ddboJSONNumberTypeNumber+"` JSON numbers are stored as DynamoDB `N` (number) attributes, which is recommended for numeric data. Note that DynamoDB `N` accepts at most 38 digits of precision and rejects values exceeding it. When set to `"+ddboJSONNumberTypeString+"` JSON numbers are stored as `S` (string) attributes, preserving the legacy behaviour.").
+				Default(ddboJSONNumberTypeString).
 				Advanced(),
 			service.NewObjectField(ddboFieldDelete,
 				service.NewBloblangField(ddboFieldDeleteCondition).
@@ -304,12 +318,12 @@ func (d *dynamoDBWriter) Connect(ctx context.Context) error {
 	return nil
 }
 
-func anyToAttributeValue(root any) types.AttributeValue {
+func anyToAttributeValue(root any, numbersAsN bool) types.AttributeValue {
 	switch v := root.(type) {
 	case map[string]any:
 		m := make(map[string]types.AttributeValue, len(v))
 		for k, v2 := range v {
-			m[k] = anyToAttributeValue(v2)
+			m[k] = anyToAttributeValue(v2, numbersAsN)
 		}
 		return &types.AttributeValueMemberM{
 			Value: m,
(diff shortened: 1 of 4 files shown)


internal/impl/aws/output_dynamodb.go
```diff
@@ -327,7 +327,7 @@ func anyToAttributeValue(root any) types.AttributeValue {
 			Value: v,
 		}
 	case json.Number:
-		return &types.AttributeValueMemberS{
+		return &types.AttributeValueMemberN{
```
jem-davies: Switching this will potentially cause data that could be delivered currently to now fail. 

If you consider the valid json number:  1234567890123456789012345678901234567890

With the DynamoDB API the PutItem action will fail with: Attempting to store more than 38 significant digits in a Number. 

but currently this would be stored as a String in DynamoDB. (Which isn't what you want, I understand)

----

There is an env var option BENTO_USE_NUMBER=false that will alter Bento's json decoder such it will unmarshal numbers into a float64. 

If you use the option BENTO_USE_NUMBER=false - it will handle both numbers with precision > 38 and < 38 BUT WILL SILENTLY LOSE A NUMBER OF SIGNIFICANT DIGITS. 

But with this PR as-is, we cannot handle valid json numbers with precision > 38: we get rejections for the dynamodb api. 

----

I think that the appropriate thing to do, is to add a new config option that, allows the user to decide between AttributeValueMemberN or AttributeValueMemberS for json.Numbers. And that would provide a place to describe the trade-offs with each approach?

jem-davies (dismissed):

internal/impl/aws/output_dynamodb.go
cthorner: Is that a realistic scenario for anyone using this? 38 digit number? Seems like a better default would be to store numbers as numbers and outliers with a env var.

cthorner: Should we do something similar to what  @gregfurman proposed for the analogous flag on #913. Avoid breaking anything now with a flag and change the default on next minor?

jem-davies: > Is that a realistic scenario for anyone using this? 38 digit number? 

I agree that a 38 digit number is obviously an unrealistic size for any real-world quantity. 

This has come up before though in the context that productIds - were losing precision. (In that instance the user perhaps shouldn't have been storing productIds as numbers though imo)

----

Changing it to `AttributeValueMemberN` would alter any existing streams, so it better if we altered this such that the users can opt-in via a config spec field?

cthorner: Sounds good. This is done with the string value as default. The flag is json_number_type, so to opt in to numbers:

```yaml
output:
  aws_dynamodb:
    json_number_type: number
```

PR #911, opened 2026-07-04
fix(aws_dynamodb): don't drop built messages when a batch peer fails construction

### LINKED ISSUE #910: aws_dynamodb output silently drops messages when a batch peer fails request construction
# `aws_dynamodb` output silently drops messages when a batch peer fails request construction

### Version
main (and 1.18.x)

### Summary
When one message in a multi-message batch fails **request construction** in the `aws_dynamodb` output — a non-JSON body with `json_map_columns` configured, or a `delete.condition` mapping that errors or doesn't resolve to a bool — every *other* message in that batch is acked as delivered but never written to DynamoDB.

Mechanism (`internal/impl/aws/output_dynamodb.go`, `WriteBatch`):

1. The per-message request builder runs inside `MessageBatch.WalkWithBatchedErrors`, which by design does **not** stop on per-message errors: it records them into a `*service.BatchError` (indexing only the failed messages) and keeps walking, so `writeReqs` ends up containing valid requests for all the successful messages.
2. `WriteBatch` then does `if err := b.WalkWithBatchedErrors(...); err != nil { return err }` — returning the `BatchError` **before `BatchWriteItem` is ever called**.
3. Per the `BatchError` contract, messages without an indexed error are treated as processed successfully — so the framework acks all the successfully-constructed messages, whose writes were never attempted.

The net effect: one malformed message poisons the batch, and the well-formed messages are the ones that get lost — silently. (The write-failure path below it handles per-message errors correctly via individual retries; only the construction-failure path has this hole.)

### Steps to reproduce
```yaml
# config.yaml
input:
  stdin: {}
  # batch two messages together
  processors: []

output:
  aws_dynamodb:
    region: us-east-1
    json_map_columns:
      "": "."
    batching:
      count: 2
```
```sh
printf '%s\n%s\n' '{"id":"good-1"}' 'not-json' | bento -c ./config.yaml
```

### Expected
`good-1` is written to DynamoDB; the malformed message is nacked/handled by the configured error strategy.

### Actual
`BatchWriteItem` is never called. The malformed message is nacked — and `good-1` is acked as delivered without ever being written. No error is surfaced for it.

### Fix
Proceed to `BatchWriteItem` with the successfully-constructed requests and merge the construction failures into the returned error so they remain nacked:

- hold the walk error instead of returning early (preserving the `ErrNotConnected` fast path, and skipping the write only when *every* message failed construction),
- send only the successfully-built requests in the initial `BatchWriteItem` (the individual-retry path already skips empty entries),
- after the existing write/retry logic resolves, fold the construction failures into the result: into the `BatchError` if there is one, into a fresh `BatchError` if the write succeeded, and unchanged for a plain (whole-batch) error since that already nacks everything.

PR to follow.

cthorner:
# fix(aws_dynamodb): don't drop built messages when a batch peer fails construction

Fixes #910

## Summary

When one message in a multi-message batch failed **request construction** (a non-JSON body with `json_map_columns` configured, or a `delete.condition` that errors or doesn't resolve to a bool), `WriteBatch` returned before `BatchWriteItem` was ever called — and because the returned `*service.BatchError` indexes only the *failed* message, every other message in the batch was acked as delivered without ever being written. Silent data loss for the well-formed messages.

## The bug

`WriteBatch` builds requests per message inside `MessageBatch.WalkWithBatchedErrors`, which by design does not stop on per-message errors: it records them in a `*BatchError` and keeps walking, leaving `writeReqs` populated for all the successful messages. The code then did:

```go
if err := b.WalkWithBatchedErrors(...); err != nil {
    return err   // BatchWriteItem never called
}
```

Per the `BatchError` contract, messages without an indexed error are treated as processed successfully — so the framework acked all successfully-constructed messages whose writes were never attempted. (The write-failure path below handles per-message errors correctly via individual retries; only the construction-failure path had this hole.)

## The fix

Proceed with the successfully-built requests and keep the construction failures nacked:

- Construction failures are recorded by original batch index during the walk instead of aborting the batch. The `ErrNotConnected` fast path is preserved, and if *every* message failed construction (including the single-message case) the walk error is returned as before, with no write call.
- The initial `BatchWriteItem` sends only the successfully-built requests (failed indices hold zero-value `WriteRequest{}` entries that DynamoDB would reject). `writeReqs` itself is not reindexed: the individual-retry path indexes it by original message index and already skips nil entries, and the unprocessed-items re-drive operates on the response's `UnprocessedItems`.
- After the existing write/retry logic resolves, a `finalize` step folds the construction failures into the result: attached as indexed errors to the `*BatchError` if there is one, into a fresh `*BatchError` if the write succeeded, and left untouched for a plain error (which already nacks the entire batch). Construction failures are deliberately **not** merged inside the individual-retry loop, whose convergence check (`IndexedErrors() == 0`) they could never satisfy.

Net behavior: successfully-built messages are written and acked; construction-failed messages are nacked exactly as before.

## Tests

- `TestDynamoDBWriteBatchPartialConstructionFailure`: batch of three (valid, non-JSON, valid) with `json_map_columns` — asserts `BatchWriteItem` receives exactly the two good puts and the returned error is a `*BatchError` with exactly one indexed failure. This test fails against the previous code with "batchFn was not called — successfully built messages were dropped".
- `TestDynamoDBWriteBatchAllConstructionFailure`: batch of two invalid messages — asserts an error is returned and `BatchWriteItem` is never called.
- All existing unit tests pass unchanged. (Integration tests require localstack; unit tests, `go vet`, `gofmt`, `golangci-lint`, and `make docs` all run clean locally.)

## Minimal repro

```yaml
# config.yaml
input:
  stdin: {}

output:
  aws_dynamodb:
    region: us-east-1
    json_map_columns:
      "": "."
    batching:
      count: 2
```
```sh
printf '%s\n%s\n' '{"id":"good-1"}' 'not-json' | bento -c ./config.yaml
```

**Before:** `BatchWriteItem` is never called; `good-1` is acked as delivered but never written.

**After:** `good-1` is written; the malformed message is nacked and handled by the configured error strategy.


gregfurman:
Thanks for opening this PR! 

While a bug has definitely been identified, the real issue here is that it's currently non-trivial to attribute a `WriteRequest` from `UnprocessedItems` to a corresponding Bento `service.Message`. 

So assuming we are able to filter out these invalid messages, there's still the problem of those items returned by `BatchWriteItem` are out-of-order and (currently) not attributable back to the original batch.

I think the simplest approach here would be to just demote the returned error from a `service.BatchError` to instead just be formatted as a standard output error -- thereby failing the whole batch.

While this doesn't fix the aforementioned attribution problem, it means that we just fail the entire batch if we cannot construct a batched request correctly.

WDYT?

cthorner:
Agreed that mapping a WriteRequest from UnprocessedItems back to a service.Message is non-trivial, but this PR doesn't do that. The construction failures are indexed during the batch walk, before any WriteRequest exists, so the attribution there is exact. The UnprocessedItems path is unchanged: when the re-drive runs out of retries it still returns a plain error and fails the whole batch, since at that point we really can't attribute.

The reason I want to keep the construction errors indexed instead of demoting them to a plain error is the fallback broker. nextBatchFromErr in internal/impl/pure/output_fallback.go routes only the indexed failed messages to the next tier, but a plain error sends the entire batch through. We run this output in production as fallback: [aws_dynamodb, dlq]. With an indexed error the good messages get written to DynamoDB and only the malformed one goes to the dlq. With a plain error nothing gets written and all of the messages end up in the dlq, so the well formed messages still never make it to the table, which is the same problem this PR is trying to fix.

That said, if you still prefer the plain error version for maintainability it does fix the silent ack, and I'm happy to cut the PR down to that. Your call.

internal/impl/aws/output_dynamodb.go
```diff
@@ -377,41 +377,119 @@ func (d *dynamoDBWriter) WriteBatch(ctx context.Context, b service.MessageBatch)
...
-		if d.conf.DeleteConditionExec == nil {
-			return d.addPutRequest(i, &b, writeReqs, p)
-		}
+	// Record per-message request construction failures by their original batch
+	// index rather than early-returning. Previously a single construction
+	// failure (e.g. a non-JSON body or a failing delete condition) caused an
+	// early return before BatchWriteItem was ever called, which meant every
+	// other message in the batch was acked as delivered but never written -
+	// silent data loss.
+	constructionErrs := map[int]error{}
```
gregfurman: nit: Why use a map here? Think a slice of `error` would be better

internal/impl/aws/output_dynamodb.go
```diff
@@ -377,41 +377,119 @@ func (d *dynamoDBWriter) WriteBatch(ctx context.Context, b service.MessageBatch)
...
+	var firstConstructionErr error
+
+	walkErr := b.WalkWithBatchedErrors(func(i int, p *service.Message) error {
```
gregfurman, line 416: `WalkWithBatchErrors` returns an `error` of type `*service.BatchError` or `service.ErrNotConnected`. Think we can leverage this e.g.

```golang
	batchError, ok := err.(*service.BatchError)
	if !ok {
		// if a BatchError is not returned, assume error is fatal/terminal
		return err
	}

	if batchError.IndexedErrors() == len(b) {
		// all messages failed, so return early
		return batchError
	}
```

internal/impl/aws/output_dynamodb.go
```diff
@@ -377,41 +377,119 @@ func (d *dynamoDBWriter) WriteBatch(ctx context.Context, b service.MessageBatch)
...
+	// finalize merges per-message construction failures into the final write
+	// result. Semantics: construction failures could not be sent to DynamoDB
+	// and so must be surfaced to the framework, otherwise the successfully
+	// built messages would be considered the whole batch and the failed ones
+	// would be falsely acked as delivered. A plain (non-*BatchError) error
+	// already nacks the entire batch, so construction failures are covered by
+	// it; but a nil or *BatchError result acks every message not explicitly
+	// marked failed, so we must attach the construction failures as indexed
+	// errors on it.
+	finalize := func(err error) error {
```
gregfurman: Why not just return the `walkErr` value returned by `WalkWithBatchedErrors`? Doesn't that contain all of this information in anycase?

internal/impl/aws/output_dynamodb.go
```diff
@@ -377,41 +377,119 @@ func (d *dynamoDBWriter) WriteBatch(ctx context.Context, b service.MessageBatch)
...
+	reqItems := writeReqs
+	if len(constructionErrs) > 0 {
+		reqItems = make([]types.WriteRequest, 0, len(writeReqs))
+		for _, req := range writeReqs {
+			if req.PutRequest == nil && req.DeleteRequest == nil {
+				continue
+			}
+			reqItems = append(reqItems, req)
+		}
+	}
```
gregfurman, line 481: 👍

gregfurman (dismissed):
I think the general idea + approach here is good 👍 I'm wondering whether the implementation could be simplified a bit, as the additional closures + `BatchError` types make this (already) confusing function more difficult to follow.

cthorner:
Done. Dropped the finalize closure and just return the walk error like you suggested, plus merged main in for #909 and #913.

PR #913, opened 2026-07-04
fix(aws_dynamodb): don't write NULL for a missing json_map_columns path

### LINKED ISSUE #912: `aws_dynamodb` output writes a NULL attribute for a missing `json_map_columns` path
# `aws_dynamodb` output writes a NULL attribute for a missing `json_map_columns` path

### Version
main (and 1.18.x)

### Summary
The `aws_dynamodb` output documents that, for `json_map_columns`, "if a path is not found within a document the column will not be populated." In practice a missing path is written as a `NULL` attribute rather than being skipped.

`jsonToMap` (`internal/impl/aws/output_dynamodb.go`) resolves an absent gabs dot-path to a `nil` value, and `anyToAttributeValue(nil)` returns `AttributeValueMemberNULL`; the put loop then does `items[k] = attr`, populating the column with `NULL`.

### Steps to reproduce
```sh
echo '{"id":"sku-1"}' | bento -c ./config.yaml
```
```yaml
# config.yaml
input:
  stdin: {}
output:
  aws_dynamodb:
    region: us-east-1
    json_map_columns:
      id: id
      category: meta.category   # absent from the document
```

### Expected
`category` is not populated (per the documented behaviour).

### Actual
```json
{ "id": {"S":"sku-1"}, "category": {"NULL": true} }
```

### Fix
Skip a column whose (non-empty) path does not exist in the document (gabs `ExistsP`), while still writing `NULL` for a path that is present with an explicit `null` value. PR to follow.

cthorner:
# fix(aws_dynamodb): don't write NULL for a missing json_map_columns path

Fixes #912

## Summary

The `aws_dynamodb` output documents that, for `json_map_columns`, "if a path is not found within a document the column will not be populated." In practice a missing path was written as a `NULL` attribute instead of being skipped, so the item ended up with a populated `NULL` column.

## The bug

`jsonToMap` resolves a `json_map_columns` dot-path with gabs; a path that is absent from the document resolves to a `nil` value, and `anyToAttributeValue(nil)` returns an `AttributeValueMemberNULL`. The put loop then unconditionally does `items[k] = attr`, so the column is written as `NULL` rather than left out.

## The fix

Skip a column whose (non-empty) path does not exist in the document, using gabs's `ExistsP`. A path that *is* present but holds an explicit `null` is still written as `NULL` — the documented contract is specifically about paths that are *not found*, and this preserves that distinction. The whole-document mapping (`"": "."`, normalized to an empty path) is unaffected and always populated.

## Test

Adds `TestDynamoDBPutJSONMapColumnsMissingPath`: a config mapping one present path (`id`) and one absent path (`meta.category`) against `{"id":"foo"}`, asserting the produced `PutRequest` item contains only `id` — no `NULL` `category` attribute.

## Minimal repro

```sh
echo '{"id":"sku-1"}' | bento -c ./config.yaml
```

```yaml
# config.yaml
input:
  stdin: {}
output:
  aws_dynamodb:
    region: us-east-1
    json_map_columns:
      id: id
      category: meta.category   # absent from the document
```

**Before** (`category` written as `NULL`):

```json
{ "id": {"S":"sku-1"}, "category": {"NULL": true} }
```

**After** (`category` left unpopulated, as documented):

```json
{ "id": {"S":"sku-1"} }
```


gregfurman:
Thanks for opening this! Can confirm that this is a bug [given the docs](https://warpstreamlabs.github.io/bento/docs/components/outputs/aws_dynamodb/#json_map_columns:~:text=If%20a%20path%20is%20not%20found%20within%20a%20document%20the%20column%20will%20not%20be%20populated):

My concern, however, is that omitting non-existent values from the payload, instead of including them as a `NULL` is _technically_ a breaking change. To preserve the original behaviour following your proposed change, users would need to include an `.or(null)` on their fields that explicitly short-circuits to `NULL` e.g

```yaml
    json_map_columns:
      id: id
      category: meta.category.or(null)   # if absent, write `NULL`
```

Since the component should be mapping a bloblang `null` to dynamodb `NULL`:

Here's what I think we should do:

1. Feature gate this deletion behaviour behind a flag e.g `omit_if_empty`/`drop_empty_column` that initially defaults to `False` 
2. Log a warning on startup saying this behaviour will change in the next minor release.
3. In the next minor release, we default the flag to `True`, remove the warning, and mark this field as `Advanced()`. Can add an inline TODO as well.

WDYT?

cthorner:
I agree. I think changing it to the more expected behavior on next minor makes sense with a flag for now. I'll get that going.

cthorner:
Done, gated behind omit_if_empty with false as the default so nothing changes unless you opt in.

jem-davies:
Also, update the relevant part where it says "[If a path is not found within a document the column will not be populated.](https://warpstreamlabs.github.io/bento/docs/components/outputs/aws_dynamodb#:~:text=If%20a%20path%20is%20not%20found%20within%20a%20document%20the%20column%20will%20not%20be%20populated.)" 

🙏

cthorner:
Done, updated that doc line and merged main to resolve the conflict with #909.

jem-davies:
@cthorner - thank you for contribution & fix 🙏

cthorner:
No problem. Thanks for the review!

PR #917, opened 2026-07-06
fix(aws_kinesis): prevent busy-spin in shard consumer select loop

richardartoul:
When a shard consumer had a message ready for dispatch but the pipeline wasn't immediately ready to accept it, nextPullChan could be set to the permanently unblocked (closed) channel, in which case the consumer's select statement never blocked and the loop spun hot without making progress. With high shard counts (e.g. 1000 shards funneling into the single msgChan) this burned enormous amounts of CPU in runtime.selectgo and channel lock contention.

Also replace the per-iteration time.After calls for pull backoff and timed batches with reusable timers to avoid allocating a new runtime timer on every empty poll, which is significant when running thousands of shard consumers.


gregfurman (approved):
LGTM. Ran kinesis int. test on my local and all looks good.

PR #919, opened 2026-07-07
feat(gcp_bigtable): Support cell timestamps

gregfurman:
Adds support for setting a BigTable cell's timestamp dynamically via a bloblang expression.

Here, we expect the timestamp expression to resolve to a valid value a UNIX timestamp in seconds (fractional values for sub-second precision), or a string value as `RFC3339Nano`.

Otherwise, if omitted or the expression is `""` at runtime, defaults to `time.Now()`.

### DIFF
--- a/website/docs/components/outputs/gcp_bigtable.md
+++ b/website/docs/components/outputs/gcp_bigtable.md
@@ -39,7 +39,7 @@ output:
     row_key: ${!metadata("kafka_key")} # No default (required)
     column: payload # No default (required)
     family: cf1 # No default (required)
-    timestamp: ""
+    timestamp: metadata("timestamp") # No default (optional)
     batching:
       count: 0
       byte_size: 0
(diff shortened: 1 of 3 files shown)


website/docs/components/outputs/gcp_bigtable.md
```diff
@@ -173,6 +175,24 @@ family: cf1
...
+
+Type: `string`  
+Default: `""`  
+
+```yml
+# Examples
+
+timestamp: metadata("timestamp")
+
+timestamp: this.event_ts_ms * 1000 + stable_hash(this.event_id) % 1000
```
di-cai: i think your code would truncate

gregfurman: Yeah good catch. This was a leftover from when I was initially writing these docs. Should be removed now.

jem-davies (approved):
LGTM (some lint issue atm)

PR #933, opened 2026-07-14
feat(kafka): support transaction isolation levels

### LINKED ISSUE #932: Support Kafka transaction isolation on kafka inputs
### Problem

The `kafka` and `kafka_franz` inputs cannot currently select Kafka consumer transaction isolation. Both underlying clients therefore use their `read_uncommitted` defaults.

This is unsafe for consumers of transactional producers such as Kafka Streams with exactly-once processing: aborted records can be observed, and records after an open transaction are not governed by the expected last-stable-offset behavior.

### Proposed change

Add the same advanced field to both inputs:

```yaml
transaction_isolation_level: read_committed
```

Allowed values would be `read_uncommitted` and `read_committed`, with `read_uncommitted` retained as the backward-compatible default.

The field maps directly to Sarama's `Consumer.IsolationLevel` and franz-go's `FetchIsolationLevel`. Redpanda Connect already uses the same field name for its franz-go Kafka input.

### Validation

The proposed tests cover both clients and prove:

- the existing default remains unchanged;
- explicit committed/uncommitted configuration maps to the client;
- invalid values fail configuration;
- an earlier open transaction holds later records behind the last stable offset;
- committed and non-transactional records remain visible;
- aborted records remain invisible.

mbuckley2000:
## What

Add an advanced `transaction_isolation_level` field to both Kafka inputs:

- `read_uncommitted` (the backward-compatible default)
- `read_committed`

The legacy `kafka` input maps it to Sarama's consumer isolation level. `kafka_franz` maps it to franz-go's fetch isolation level. Both inputs share one field definition and defensive parser, and their generated documentation describes the behavior.

Closes #932.

## Why

Both clients support Kafka transactions, but Bento previously offered no way to hide aborted records from consumers of transactional producers such as Kafka Streams exactly-once topologies.

The field name and values follow the existing Redpanda Connect franz-go input convention.

## Tests

- Unit/config tests for both inputs:
  - default remains uncommitted;
  - explicit uncommitted;
  - explicit committed;
  - invalid values fail closed.
- Broker-backed integration test for both inputs:
  - open transaction enforces the last-stable-offset barrier;
  - abort releases the following non-transactional record without leaking the aborted one;
  - committed transaction is visible;
  - later aborted transaction is hidden;
  - no extra records leak.
- `go vet ./internal/impl/kafka`
- `golangci-lint -j 1 run -c .golangci.yml ./internal/impl/kafka` (0 issues)
- `go test ./internal/impl/kafka -run TestIntegrationKafkaTransactionIsolation -count=1 -v -timeout 2m`
  - Sarama pass
  - franz-go pass
- Focused generated-document lint passes for both changed pages.

`make docs` generated the changed Kafka pages successfully; its repository-wide lint subsequently failed on pre-existing DuckDB pages because the normal non-CGO documentation build cannot load the DuckDB driver.


internal/impl/kafka/transaction_isolation.go
```diff
@@ -0,0 +1,41 @@
...
+)
+
+func newTransactionIsolationLevelField() *service.ConfigField {
+	return service.NewStringAnnotatedEnumField(transactionIsolationLevelFieldName, map[string]string{
+		string(transactionIsolationLevelReadUncommitted): "Consume all records, including records from aborted or open transactions.",
+		string(transactionIsolationLevelReadCommitted):   "Consume only non-transactional records and records from committed transactions.",
+	}).
+		Description("Controls the isolation level used for Kafka fetch requests.").
+		Default(string(transactionIsolationLevelReadUncommitted)).
+		Advanced()
```
jem-davies:
```suggestion
		Advanced().
		Version("1.20.0")
```

jem-davies (commented):
Thanks for the PR 🙏  - yeah seems we aren't exposing this config option at the moment. 

Only thing I would say is that we could get rid of a lot of stuff:

 - The `NewStringAnnotatedEnumField` will throw lint errors if the value isn't one of the options, anyway. (I do realise it doesn't do this using the .ParseYAML() like in your test) Perhaps this could be simplified such we only have one switch in the components constructor i.e. `newFranzKafkaReaderFromConfig` - we add a default here that throws an error - like with `kafka`'s `sasl.mechanism`
- The new unit tests are somewhat redundant - we could keep them - but they largely seem to just be checking configField spec & parsing - not much else.
- I would move this field down such it appears later in the config for doc readability.
- I would just duplicate the field across the two components personally 

This review is nit-picky but then we would want to be consistent with the existing patterns in the code imo.

PR #934, opened 2026-07-15
feat(aws): add aws_bedrock_invoke processor

louisdvs:
## Context

Adds an `aws_bedrock_invoke` processor, in a similar style to the `aws_lambda` processor. This processors should be used in conjunction with payloads that are dynamically correspond to a model's required invocation request shape.


website/docs/components/processors/aws_bedrock_embeddings.md
```diff
@@ -0,0 +1,270 @@
...
+input_type: search_query
+
+input_type: classification
+
+input_type: clustering
+```
+
+### `rate_limit`
+
+An optional [`rate_limit`](/docs/components/rate_limits/about) resource to throttle Bedrock `InvokeModel` requests by. The limit is applied per request, and each Cohere batch (up to 96 texts) counts as one request. Note that Bento rate limits are enforced per running instance, so this does not bound the aggregate rate across horizontally scaled deployments.
```
gregfurman:
```suggestion
An optional [`rate_limit`](/docs/components/rate_limits/about) resource to throttle Bedrock `InvokeModel` requests by. The limit is applied per request, and each Cohere batch (up to 96 texts) counts as one request. 
```

This is not strictly true since you could use the [redis rate limit component](https://warpstreamlabs.github.io/bento/docs/components/rate_limits/redis) which would be shareable across multiple instances. From the linked docs:

> The rate limit is shared across all instances of Bento that use the same Redis instance, which must all have a consistent count and interval.

internal/impl/aws/processor_bedrock_embeddings.go
```diff
@@ -0,0 +1,384 @@
...
+				"amazon.titan-embed-text-v2:0",
+				"amazon.titan-embed-text-v1",
+				"cohere.embed-english-v3",
+				"cohere.embed-multilingual-v3",
+			)).
+		Field(service.NewStringField(bedFieldProvider).
+			Description("The model provider family, which determines the request and response encoding. When set to `auto` it is inferred from the `model` ID, but this is not possible when `model` is an inference profile ARN, so it must be set explicitly in that case.").
+			Default(bedProviderAuto).
+			LintRule(`root = if ![ "auto", "amazon_titan", "cohere" ].contains(this) { [ "provider must be one of: auto, amazon_titan, cohere" ] }`).
+			Examples(bedProviderAuto, bedProviderTitan, bedProviderCohere)).
```
gregfurman: IMO we should remove this. Think it's fair to expect models be specified in full by their ID. Also means that for every new model type that gets supported, we have to add some more handling for the auto type.

gregfurman (dismissed):
@louisdvs Thanks for opening! Do you mind if I commandeer this PR? have realised that this component is going to require a bit more in-depth work if we're wanting to support multiple model families -- since each has a differing input structure.

Think the codec approach is a good start but I'm worried that we'll run into too many edge cases as we start extending support to other model types.

Tbh I think the best approach here would be if we could use some form of request/payload mapping via bloblang to convert a message/batch of messages into the correct format -- that way we could just use a mapping to return some JSON bytes in the correct format for bedrock without needing to explicitly support each model input type.

gregfurman:
I've simplified the approach to instead allow for any input to be passed, instead of constraining this via codecs. 

Basically, we expect the input to be determined via a `request_map` -- similar to how we do things with the lambda processor.

@jem-davies Could you give this a look?

jem-davies (approved):
LGTM

jem-davies:
@gregfurman - will need you to dismiss review, before we can merge it

louisdvs:
Thanks for the work team ❤️  Appreciate it.

PR #940, opened 2026-07-20
feat(cache): add ListableCache interface for enumerating cache keys

ecordell:
## Description

Adds an optional `ListableCache` interface to the `service` package for caches that are able to enumerate the keys they hold:

```go
type ListableCache interface {
	Cache

	// ListKeys returns a slice of all keys currently held by the cache.
	ListKeys(ctx context.Context) ([]string, error)
}
```

Not all caches can support key listing, so support is detected via type assertion on the cache obtained from `*Resources.AccessCache`. The air gap, metrics and reverse air gap wrappers propagate the capability only when the underlying cache implements it, so the assertion stays truthful through the whole chain. The metrics wrapper emits the standard cache metrics with `operation="list_keys"`.

Implementations included: `memory` (walks shards, omits TTL-expired items to match `Get` semantics), `file` (recursive directory walk), `aws_s3` (paginated `ListObjectsV2` with the cache's existing backoff pattern) and `gcp_cloud_storage` (bucket object iterator). The mock cache behind `MockResourcesOptAddCache` also implements it so plugin authors can test listable-cache consumers.

Motivation: components consuming a cache resource currently have no way to enumerate its contents, so patterns like "replay everything stored in this cache" require maintaining a separate index entry alongside the cached items. A follow-up PR adds a `cache` input built on this interface.

## Testing

- Unit tests for the type-assertion passthrough at each wrapper layer, including a round trip through all of them
- `ListKeys` unit tests for `memory` (including shards and TTL expiry) and `file` (including nested directories)
- New `CacheTestListKeys` integration test definition, wired into the `aws_s3` localstack suite and verified locally

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ All notable changes to this project will be documented in this file.
 
 ## Unreleased
 
+### Added
+
+- Optional `ListKeys` method for caches enabling consumers of cache resources to enumerate keys, caches that do not support key listing yield `ErrKeyListingNotSupported` @ecordell
+- `memory`, `file`, `aws_s3` and `gcp_cloud_storage` caches support `ListKeys` @ecordell
+
 ### Fixed
 
 - `parse_big_decimal` rejects scales above 16383 to avoid unbounded big-integer work @aratz-lasa
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,8 +7,8 @@ All notable changes to this project will be documented in this file.
 
 ### Added
 
-- Optional `ListKeys` method for caches enabling consumers of cache resources to enumerate keys, caches that do not support key listing yield `ErrKeyListingNotSupported` @ecordell
-- `memory`, `file`, `aws_s3` and `gcp_cloud_storage` caches support `ListKeys` @ecordell
+- Optional `Keys` method for caches enabling consumers of cache resources to enumerate keys, caches that do not support key listing yield `ErrKeyListingNotSupported` @ecordell
+- `memory`, `file`, `aws_s3` and `gcp_cloud_storage` caches support `Keys` @ecordell
 
 ### Fixed
 
--- a/internal/component/cache/interface.go
+++ b/internal/component/cache/interface.go
@@ -2,6 +2,7 @@ package cache
 
 import (
 	"context"
+	"iter"
 	"time"
 )
 
@@ -36,6 +37,11 @@ type V1 interface {
 
 	// Delete attempts to remove a key. Returns an error if a failure occurs.
 	Delete(ctx context.Context, key string) error
+
+	// ListKeys returns an iterator over all keys currently held by the cache.
+	// Iteration stops after the first non-nil error, caches that are unable
+	// to enumerate their keys yield component.ErrKeyListingNotSupported.
+	ListKeys(ctx context.Context) iter.Seq2[string, error]
 
 	// Close the component, blocks until either the underlying resources are
 	// cleaned up or the context is cancelled. Returns an error if the context
--- a/internal/component/cache/interface.go
+++ b/internal/component/cache/interface.go
@@ -42,13 +42,17 @@ type V1 interface {
 	// Delete attempts to remove a key. Returns an error if a failure occurs.
 	Delete(ctx context.Context, key string) error
 
-	// Keys returns an iterator over all keys currently held by the cache.
-	// Iteration stops after the first non-nil error, caches that are unable
-	// to enumerate their keys yield component.ErrKeyListingNotSupported.
-	Keys(ctx context.Context) KeyIterator
-
 	// Close the component, blocks until either the underlying resources are
 	// cleaned up or the context is cancelled. Returns an error if the context
 	// is cancelled.
 	Close(ctx context.Context) error
+}
+
+type Listable interface {
+	V1
+
+	// Keys returns an iterator over all keys currently held by the cache.
+	// Iteration stops after the first non-nil error, caches that are unable
+	// to enumerate their keys yield component.ErrKeyListingNotSupported.
+	Keys(ctx context.Context) KeyIterator
 }
--- a/internal/component/cache/interface.go
+++ b/internal/component/cache/interface.go
@@ -12,6 +12,10 @@ type TTLItem struct {
 	TTL   *time.Duration
 }
 
+// KeyIterator is an iterator over the keys held by a cache. Iteration stops
+// after the first non-nil error.
+type KeyIterator = iter.Seq2[string, error]
+
 // V1 Defines a common interface of cache implementations.
 type V1 interface {
 	// Get attempts to locate and return a cached value by its key, returns an
@@ -38,10 +42,10 @@ type V1 interface {
 	// Delete attempts to remove a key. Returns an error if a failure occurs.
 	Delete(ctx context.Context, key string) error
 
-	// ListKeys returns an iterator over all keys currently held by the cache.
+	// Keys returns an iterator over all keys currently held by the cache.
 	// Iteration stops after the first non-nil error, caches that are unable
 	// to enumerate their keys yield component.ErrKeyListingNotSupported.
-	ListKeys(ctx context.Context) iter.Seq2[string, error]
+	Keys(ctx context.Context) KeyIterator
 
 	// Close the component, blocks until either the underlying resources are
 	// cleaned up or the context is cancelled. Returns an error if the context
--- a/internal/component/errors.go
+++ b/internal/component/errors.go
@@ -87,15 +87,21 @@ func (e *ErrBackOff) Error() string {
 
 // Manager errors.
 var (
-	ErrInputNotFound          = errors.New("input not found")
-	ErrCacheNotFound          = errors.New("cache not found")
-	ErrProcessorNotFound      = errors.New("processor not found")
-	ErrRateLimitNotFound      = errors.New("rate limit not found")
-	ErrOutputNotFound         = errors.New("output not found")
+	ErrInputNotFound     = errors.New("input not found")
+	ErrCacheNotFound     = errors.New("cache not found")
+	ErrProcessorNotFound = errors.New("processor not found")
+	ErrRateLimitNotFound = errors.New("rate limit not found")
+	ErrOutputNotFound    = errors.New("output not found")
+	ErrPipeNotFound      = errors.New("pipe was not found")
+)
+
+//------------------------------------------------------------------------------
+
+// Cache errors.
+var (
 	ErrKeyAlreadyExists       = errors.New("key already exists")
 	ErrKeyNotFound            = errors.New("key does not exist")
 	ErrKeyListingNotSupported = errors.New("cache does not support listing keys")
-	ErrPipeNotFound           = errors.New("pipe was not found")
 )
 
 //------------------------------------------------------------------------------
--- a/internal/component/errors.go
+++ b/internal/component/errors.go
@@ -87,14 +87,15 @@ func (e *ErrBackOff) Error() string {
 
 // Manager errors.
 var (
-	ErrInputNotFound     = errors.New("input not found")
-	ErrCacheNotFound     = errors.New("cache not found")
-	ErrProcessorNotFound = errors.New("processor not found")
-	ErrRateLimitNotFound = errors.New("rate limit not found")
-	ErrOutputNotFound    = errors.New("output not found")
-	ErrKeyAlreadyExists  = errors.New("key already exists")
-	ErrKeyNotFound       = errors.New("key does not exist")
-	ErrPipeNotFound      = errors.New("pipe was not found")
+	ErrInputNotFound          = errors.New("input not found")
+	ErrCacheNotFound          = errors.New("cache not found")
+	ErrProcessorNotFound      = errors.New("processor not found")
+	ErrRateLimitNotFound      = errors.New("rate limit not found")
+	ErrOutputNotFound         = errors.New("output not found")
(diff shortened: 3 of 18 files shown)


ecordell:
👋 I realize this was a little out of the blue, any objections to the feature? @jem-davies @gregfurman 

This is in service of an improved kube-watch component I'm working on. Not being able to cache list has meant some rough workarounds.

This PR is the only one I need since the other component consumes the cache directly. I made the other one (#941) just for symmetry with other components.

gregfurman:
Hey @ecordell. Apologies for the delay here! Will give this a look within the week.

internal/component/cache/interface.go
```diff
@@ -11,6 +11,13 @@ type TTLItem struct {
 	TTL   *time.Duration
 }
 
+// KeyLister is an optional interface implemented by caches that are capable
+// of enumerating the keys they hold.
+type KeyLister interface {
+	// ListKeys returns a slice of all keys currently held by the cache.
+	ListKeys(ctx context.Context) ([]string, error)
```
gregfurman: Why not we rather add this to the `V1` interface?

public/service/cache.go
```diff
@@ -66,7 +79,11 @@ type airGapCache struct {
 func newAirGapCache(c Cache, stats metrics.Type) cache.V1 {
 	ag := &airGapCache{c: c, cm: nil}
 	ag.cm, _ = c.(batchedCache)
-	return cache.MetricsForCache(ag, stats)
+	var v1 cache.V1 = ag
+	if lc, ok := c.(ListableCache); ok {
+		v1 = &listableAirGapCache{airGapCache: ag, l: lc}
+	}
```
gregfurman: IMO the attribute approach is much cleaner. That is, include an attribute for the listenable cache, a private interface `listableCache`, etc.

gregfurman (changes requested):
Hey @ecordell. I've given this a look and think better support for cache item enumeration is a neat addition.

However, I think we could improve the approach here to closer resemble how we support multiple adds via `SetMulti`.

Lmk if you have capacity to address this!

gregfurman:
Hey @ecordell. Are you still able to address my review? If not, myself or Jem are happy to commandeer this PR since I think this'll be a great add to the component library! Lmk 🙂

ecordell:
Just got to it, sorry for the delay!

The only thing I'd call out is that I found it useful to have the interface public so that consuming components (i.e. those that embed a `cache:` subconfig) can assert on it. If there's a better pattern for that happy to change that too.

public/service/cache.go
```diff
@@ -14,6 +14,10 @@ import (
 var (
 	ErrKeyAlreadyExists = errors.New("key already exists")
 	ErrKeyNotFound      = errors.New("key does not exist")
+
+	// ErrKeyListingNotSupported is returned by the ListKeys method of caches
+	// that are unable to enumerate the keys they hold.
```
gregfurman: 🧹 Think we can remove the extra comments in favour of a smaller diff

```suggestion
```

internal/impl/gcp/cache_cloud_storage.go
```diff
@@ -140,6 +141,22 @@ func (c *gcpCloudStorageCache) Delete(ctx context.Context, key string) error {
...
 }
 
+func (c *gcpCloudStorageCache) ListKeys(ctx context.Context) ([]string, error) {
+	var keys []string
+	it := c.bucketHandle.Objects(ctx, nil)
+	for {
+		attrs, err := it.Next()
+		if errors.Is(err, iterator.Done) {
+			break
+		}
```
gregfurman: I actually really like how this bucket iteration is handled. Makes me wonder whether we should actually be returning an iterator (perhaps like `iter.Seq`) that a caller can use to iterate over the keys themselves instead of the `ListKeys` approach. Similar to how this iterator works with a `Next` and an `iterator.Done` sentinel. Could also allow us to add a`Values` enumator method as well e.g

```go
type EnumerableCache interface {
   Keys(ctx context.Context) iter.Seq[string]
   Values(ctx context.Context) iter.Seq[[]byte]
}
```

Although it would require quite some re-work on this PR I'm wondering if this could be more ergonomic for callers.

Wdyt?

gregfurman (commented):
> The only thing I'd call out is that I found it useful to have the interface public so that consuming components (i.e. those that embed a cache: subconfig) can assert on it. 

Assuming you're still planning on using the exposed `AccessCache` method, I think for now we can just rely on manually asserting via a separately defined interface e.g

```go
var keys []string
if err := r.resources.AccessCache(ctx, r.cache, func(c service.Cache) {
        lc, ok := c.(interface { ListKeys(ctx context.Context) ([]string, error) })
        if !ok {
          return
        }
	keys, _ = lc.ListKeys(ctx)
}); err != nil {
	return err
}

for _, key := range keys {
   // do stuff with keys if set
}
```

Once Go 1.27 is out, I think we could extend some of these `Access*` methods to easier accept generic interfaces, which IMO could then justify exporting the `ListenableCache` interface:

```go
func (r *Resources) AccessCache[C Cache](ctx context.Context, name string, fn func(c C)) error {
	return r.mgr.AccessCache(ctx, name, func(c cache.V1) {
		fn(newReverseAirGapCache(c))
	})
}
```

internal/impl/gcp/cache_cloud_storage.go
ecordell: I like it, I prefer native iterators and I don't recall why I didn't do that in the first place.

I went with:

```go
ListKeys(ctx context.Context) iter.Seq2[string, error]
```

since for many of the implementations, fetching can fail mid-iteration.

I held off on `Values` for now; to me that requires a little more thought and some other cache implementations to look at. For example a lsm-tree cache would give you keys for free when iterating values, so `ListEntries() iter.Seq[Entry]` might make more sense. And for blob stores (afaik) getting the value always requires a separate fetch so there's not much benefit to iterating values vs. iterating keys and doing a `get` per key.

But if you have a strong opinion on the direction to go I don't mind implementing it.

gregfurman: Ahh nice this is looking great! Will leave another review shortly.

internal/component/cache/interface.go
```diff
@@ -37,6 +38,11 @@ type V1 interface {
 	// Delete attempts to remove a key. Returns an error if a failure occurs.
 	Delete(ctx context.Context, key string) error
 
+	// ListKeys returns an iterator over all keys currently held by the cache.
+	// Iteration stops after the first non-nil error, caches that are unable
+	// to enumerate their keys yield component.ErrKeyListingNotSupported.
+	ListKeys(ctx context.Context) iter.Seq2[string, error]
```
gregfurman: Two suggestions:

1. `ListKeys` can be renamed to `Keys`? IMO the `ListKeys` verb could be misleading since it sounds like it'll return all keys in a single call.
2. Can we create a separate internal type alias for this returned `iter.Seq2[string, error]`? Can just be called `Iterator` (or similar) but I think it'd provide an easier interface for assertions that doesn't require manually defining `iter.Seq2[string, error]` on an interface for type inference. This can be defined both internally in `internal/component/cache` and then re-exported in `public/service` similar to how we define `TTLItem` internally and `CacheItem` in `public` e.g


https://github.com/warpstreamlabs/bento/blob/7cc0a077515347e66b7a944e517fe8d8fd8d0f94/internal/component/cache/interface.go#L9-L12


https://github.com/warpstreamlabs/bento/blob/7cc0a077515347e66b7a944e517fe8d8fd8d0f94/public/service/cache.go#L43-L47

internal/impl/aws/cache_s3.go
```diff
@@ -256,6 +257,42 @@ func (s *s3Cache) Delete(ctx context.Context, key string) (err error) {
...
+		defer func() {
+			boff.Reset()
+			s.boffPool.Put(boff)
+		}()
+
+		pager := s3.NewListObjectsV2Paginator(s.s3, &s3.ListObjectsV2Input{
+			Bucket: &s.bucket,
+		})
+		for pager.HasMorePages() {
+			page, err := pager.NextPage(ctx)
```
ecordell, line 275: this is basically the naive approach and I'm mulling whether it's a premature optimization to try and fetch the next batch while yielding the current batches results

I have the prefetching ready if that's of interest

public/service/cache.go
```diff
@@ -137,6 +151,23 @@ func (a *airGapCache) Delete(ctx context.Context, key string) error {
 	return a.c.Delete(ctx, key)
 }
 
+func (a *airGapCache) ListKeys(ctx context.Context) iter.Seq2[string, error] {
+	return func(yield func(string, error) bool) {
+		if a.cl == nil {
+			yield("", component.ErrKeyListingNotSupported)
+			return
+		}
+		for key, err := range a.cl.ListKeys(ctx) {
```
gregfurman: suggestion: wdyt about using a sentinel error in a similar approach to how GCP cloud storage does it i.e `iterator.Done`

internal/component/errors.go
```diff
@@ -87,14 +87,15 @@ func (e *ErrBackOff) Error() string {
...
-	ErrKeyNotFound       = errors.New("key does not exist")
-	ErrPipeNotFound      = errors.New("pipe was not found")
+	ErrInputNotFound          = errors.New("input not found")
+	ErrCacheNotFound          = errors.New("cache not found")
+	ErrProcessorNotFound      = errors.New("processor not found")
+	ErrRateLimitNotFound      = errors.New("rate limit not found")
+	ErrOutputNotFound         = errors.New("output not found")
+	ErrKeyAlreadyExists       = errors.New("key already exists")
+	ErrKeyNotFound            = errors.New("key does not exist")
+	ErrKeyListingNotSupported = errors.New("cache does not support listing keys")
```
gregfurman, line 104: nit: perhaps these should all be moved into their own `Cache errors.` section like with the buffer errors below.

gregfurman (commented):
OK I've gone through this quite thoroughly and just realised that all caches accessed from the public `service.Resources` will be wrapped with a `reverseAirGapCache`, meaning that we wont be able to assert on `service.Cache` like I initially thought as the reverse airgap struct will ensure this _always_ fits the interface -- meaning your initial flagging of this was correct, my bad 🤦 

With all of that said, since this approach will unblock the input component you're working on, I think the best way forward is to keep this approach as-is but rely on the `ErrKeyListingNotSupported` being returned to signal whether a component has listable support. 


While the interface approach would be ideal, perhaps we do it in a follow-up, since I'm probably pushing the envelope of your generosity here 😅

public/service/cache.go
```diff
@@ -63,19 +68,93 @@ type existsCache interface {
...
+// once the caller has stopped consuming (an early break, or a downstream
+// error), after which produce should stop and return - typically nil.
+// Returning a non-nil error from produce terminates iteration after yielding
+// that error to the caller. If ctx is cancelled iteration stops without
+// yielding a further error; produce should return promptly when ctx is done.
+//
+// This is a convenience for implementing the optional key-listing behaviour of
+// a paginated cache; it manages the goroutine lifecycle and cancellation on
+// early termination so implementations only need to express their paging loop.
+func PrefetchKeys(ctx context.Context, readAhead int, produce func(ctx context.Context, emit func(key string) bool) error) KeyIterator {
```
gregfurman: While I think this is a neat optimisation, this kinda feels to me like it should be done at the cache implementation level instead of being a standalone utility we expose.

Not that I don't think it's useful, but keep in mind anything we expose in `public` is intended to be supported indefinitely, so I'm hesitant to provide this in the standard SDK.

Perhaps this can be moved into the internal cache package instead and we only allow internal bento components to use it.

gregfurman (commented):
I've added the interfaces back to the implementation. Lastly, just a comment on the now included prefetch approach.

public/service/cache.go
ecordell: Since the native iterators "end" the for loop when they're exhausted, I personally think a sentinel error is a little less idiomatic

public/service/cache.go
ecordell: yeah I just realized that the naive implementation would cause pauses for each page for no reason, and the pattern could be generalized to any paginated source. I had it exported with the idea that if you were writing an external component you could use it, but I moved it back internal (easy enough to replicate if you need it).

Which reminds me, we wrote (and have been using  in prod) an azure blob store cache impl for over a year now, I forgot to move that upstream. Will open a PR soon if it would be accepted? It gives us parity between aws/gcp/azure for blob-store-as-cache.

ecordell:
@gregfurman sorry for the delay here, I think I've addressed everything. Thanks for your patience!

CHANGELOG.md
```diff
@@ -5,13 +5,14 @@ All notable changes to this project will be documented in this file.
 
 ## Unreleased
 
-### Fixed
+### Added
 
-- `parse_big_decimal` rejects scales above 16383 to avoid unbounded big-integer work @aratz-lasa
+- Optional `ListKeys` method for caches enabling consumers of cache resources to enumerate keys, caches that do not support key listing yield `ErrKeyListingNotSupported` @ecordell
+- `memory`, `file`, `aws_s3` and `gcp_cloud_storage` caches support `ListKeys` @ecordell
```
gregfurman:
```suggestion
- Optional `Keys` method for caches enabling consumers of cache resources to enumerate keys, caches that do not support key listing yield `ErrKeyListingNotSupported` @ecordell
- `memory`, `file`, `aws_s3` and `gcp_cloud_storage` caches support `Keys` @ecordell
```

gregfurman (approved):
LGTM! Think this is gonna be a neat addition to the cache components. Feels like I put you through the ringer here with this PR, so appreciate you humoring me 😅 

Let's :shipit:

PR #941, opened 2026-07-20
feat(input): add cache input for consuming items from a cache resource

ecordell:
## Description

**Stacked on #940** — only the last commit (`feat(input): add cache input...`) is new to this PR; the diff will reduce to it once #940 merges.

Adds a `cache` input that reads all items stored within a cache resource supporting the optional `ListableCache` interface from #940, emitting one message per item with the item's key in the `cache_key` metadata field, and shutting down once the contents of the cache have been fully consumed:

```yaml
input:
  cache:
    resource: mycache
```

Keys are captured once on connect and values are then read individually, so only the key list is held in memory rather than the full cache contents. Items deleted after the keys were captured are skipped. Attempting to consume from a cache that doesn't support key listing surfaces as a clear connection error.

Because the input terminates once exhausted, it composes with `sequence` for seed-then-follow patterns (replay stored state, then consume a live source).

## Testing

Unit tests covering read-all, skip-deleted and missing-resource behaviour, plus a stream-level test through `StreamBuilder` backed by a `memory` cache. Component docs generated with `bento_docs_gen` and lint-checked.


internal/impl/pure/input_cache.go
```diff
@@ -0,0 +1,155 @@
...
+				Description("The [`cache` resource](/docs/components/caches/about) to read from."),
+			service.NewAutoRetryNacksToggleField(),
+		).
+		Example(
+			"Migrate cache items",
+			"Reads all items from a `file` cache and writes them into a `redis` cache under the same keys.",
+			`
+input:
+  cache:
+    resource: source
```
gregfurman, line 42: Should this be mirroring the output destination with the `target` field being used to specify the cache resource? e.g

```yml
input:
  cache:
    target: source
```

internal/impl/pure/input_cache.go
```diff
@@ -0,0 +1,155 @@
...
+			// The item was deleted after the keys were listed.
+			continue
+		}
+		if getErr != nil {
+			return nil, nil, getErr
+		}
+
+		msg := service.NewMessage(value)
+		msg.MetaSetMut("cache_key", key)
+		return msg, func(context.Context, error) error { return nil }, nil
```
gregfurman, line 148: Q: should we not be deleting the key from the cache on successful delivery? Or perhaps an option that allows for this would be a good idea like we do with the `aws_s3` input/processor's `delete_objects` field.

internal/impl/pure/input_cache.go
```diff
@@ -0,0 +1,155 @@
...
+	var listErr error
+	if err := c.res.AccessCache(ctx, c.cacheName, func(cache service.Cache) {
+		lister, ok := cache.(interface {
+			Keys(ctx context.Context) service.KeyIterator
+		})
+		if !ok {
+			listErr = fmt.Errorf("cache resource '%v' does not support listing keys", c.cacheName)
+			return
+		}
+		for key, err := range lister.Keys(ctx) {
```
gregfurman, line 109: suggestion: Wondering if this should be converted into a pull-style iterator (via `iter.Pull`) that gets pulled from each time `Read` is called instead of this approach that'll only iterate over what we have on startup.

Otherwise, I think we should consider using a channel here that gets populated and read off-of in the read call. That way, we could potentially even run this listing logic within a goroutine that loops around constantly giving us new keys for the `Read` to use.

internal/impl/pure/input_cache.go
```diff
@@ -0,0 +1,155 @@
...
+
+import (
+	"context"
+	"errors"
+	"fmt"
+
+	"github.com/warpstreamlabs/bento/public/service"
+)
+
+const ciFieldResource = "resource"
```
gregfurman, line 11: I think some filtering here to include/exclude certain keys from the iteration could be a good idea e.g

```yml
cache:
  resource: source
  exclude_prefixes:
    - "system_"
    - "foo_"
```

That way, we can save ourselves a `processor` call to filter out unwanted data from within the component instead of only after a `Get` call has been made.

internal/impl/pure/input_cache.go
```diff
@@ -0,0 +1,155 @@
...
+
+func cacheInputSpec() *service.ConfigSpec {
+	return service.NewConfigSpec().
+		Beta().
+		Categories("Utility").
+		Version("1.21.0").
+		Summary("Reads items stored within a [cache resource](/docs/components/caches/about), emitting each item as a message and shutting down once the contents of the cache have been fully consumed.").
```
gregfurman, line 18: > shutting down once the contents of the cache have been fully consumed.

We could the https://warpstreamlabs.github.io/bento/docs/components/inputs/read_until to restart the input when exhausted. Think an example of this would also be great for those wanting to constantly poll from the input.

internal/impl/pure/input_cache.go
```diff
@@ -0,0 +1,155 @@
...
+
+- cache_key
+
+You can access these metadata fields using [function interpolation](/docs/configuration/interpolation#bloblang-queries).`).
+		Fields(
+			service.NewStringField(ciFieldResource).
+				Description("The [`cache` resource](/docs/components/caches/about) to read from."),
+			service.NewAutoRetryNacksToggleField(),
+		).
+		Example(
```
gregfurman, line 36: Can we add another example here using the https://warpstreamlabs.github.io/bento/docs/components/inputs/sequence/ ?

Think it's a really good use-case for this component to drain a cache prior to the next one starting!

gregfurman (changes requested):
I've rebase this to include the latest changes. It's looking great!

gregfurman:
@ecordell OK I've merged the prefix field! Lmk when you have a chance to give this another look. Appreciate your patience here 🙂

PR #946, opened 2026-07-24
move mutex lock to batchErrFailed

jem-davies:
Mutex was only being locked in the callback func onFailureHandler - should be anywhere that we check the nil-ness of batchErr. 

Had claude do a review of this `elasticsearch_v2` component and it pointed it out.


gregfurman (approved):
LGTM. Just wanna confirm that we _actually_ need a mutex here though 🤔

jem-davies:
> LGTM

I have ran the race detector on the tests, without the mutex and it flags. I understand that the call onFailureHandler callback can happen concurrently by the elasticsearch lib.

Issue #939, opened 2026-07-20
Feature: Support HTTPS URLs in the `-c`/`--config` flag


Omar1H1:
## Summary

Allow Bento to load its configuration from an HTTPS URL in addition to a local file path.

Example:

```bash
bento -c https://config.internal/bento.yaml
```
Optionally, remote configuration requests could support custom HTTP headers:
```bash
bento \
  -c https://config.internal/bento.yaml \
  --config-header "Authorization: Bearer $TOKEN"
```
## Motivation

Currently, Bento expects configuration to be provided as a local file.

In some scenarios, we migh want to provide this via https instaid of local files. Supporting remote configuration sources would allow users to point Bento directly at a remote configuration location instead of adding an additional step to download and store the file locally.

This is also consistent with the approach used by other tooling, such as `kubectl apply -f`, where configuration resources can be loaded from either local files or HTTP(S) URLs.
```bash
kubectl apply -f ./my-manifest.yaml                 # create resource(s)
kubectl apply -f ./my1.yaml -f ./my2.yaml           # create from multiple files
kubectl apply -f ./dir                              # create resource(s) in all manifest files in dir
kubectl apply -f https://example.com/manifest.yaml  # create resource(s) from url (Note: this is an example domain and does not contain a valid manifest)
```
## Proposed behavior

- `-c`/`--config` accepts either:
    - a local filesystem path;
    - an HTTPS URL.
- When an HTTPS URL is provided, Bento fetches the configuration during startup.
- Optional HTTP headers can be provided for authenticated remote configuration endpoints.
- File watching remains supported only for local configuration files, since remote URLs do not provide the same filesystem semantics, if the flag is set with https it exit and return ""error: --watcher is not supported with a remote config URL"

## Implementation notes

The implementation would mainly involve:
- CLI handling:
  - `internal/cli/run.go`
  - `internal/cli/common/reader.go`
    
  Add support for accepting remote configuration sources and handling optional HTTP headers.

- Configuration loading:
  - `internal/config/reader.go`
  - `internal/config/lint.go`
    
  Extend the configuration reader to fetch remote configurations and pass the fetched content through the existing parsing/linting flow.

The goal is to keep the existing local file behavior unchanged while adding HTTPS as an additional configuration source.

## Open question: implicit vs explicit configuration sources

One design question is whether remote configuration should extend the existing `-c` flag or use a dedicated flag.

### Option 1: Extend `-c` (implicit)
```bash
[...]

PR #950, opened 2026-07-24
Feat: support HTTPS URLs in the -c/--config flag

Omar1H1:
## Summary

Implements #939, allows `-c`/`--config` to accept an `https://` URL 
in addition to a local file path.

## Changes

- `internal/config/reader.go` — URL detection, fetch logic, watcher guard
- `internal/config/lint.go` — URL support for `ReadYAMLFileLinted`
- `internal/cli/run.go` — `--config-header` flag definition
- `internal/cli/common/reader.go` — watcher fatal error, headers wired into `NewReader`
- `internal/config/reader_test.go` — added unit tests for `fetchRemoteConfig` and remote reader behaviour
- `internal/config/lint_test.go` — new file, unit tests for `ReadYAMLFileLinted` with remote URLs

## Behaviour

- `https://` prefix → remote fetch before config parsing
- Everything else → unchanged file path behaviour
- Non-2xx response → exits with status code in error message
- `--watcher` + URL → exits with fatal error immediately
- env var interpolation (`${FOO:default}`) still works after fetch
- `--config-header "Name: Value"` (repeatable) for auth headers

## Testing

- bento -c _url_ → stream starts correctly
- bento -c _url_ echo → prints normalised config
- bento -c _url_ streams → streams mode works
- bento lint _url_ → lints remote config
- bento -c _url_ --config-header "K: V" → auth header forwarded
- bento -c _url_ --config-header "bad" → clear format error
- bento -c _url-returning-404_ → clear 404 error
- bento -c _url-returning-401_ → clear 401 error
- bento -c _url_ --watcher → fatal error on startup
- bento -c _path/to/file.yaml_ → unchanged behaviour

## Known limitations

- Only `https://` URLs are supported. plain `http://` falls back to 
  file path behaviour.
- `bento lint <url>` works for public endpoints but does not support 
  `--config-header`. `ReadYAMLFileLinted` has no access to CLI flags 
  at that call site. Can be addressed in a follow-up.
  
  ## Open question

As noted in #939. happy to move to a dedicated `--config-url` flag if that is preferred over extending `-c`.

### DIFF
--- a/internal/config/lint.go
+++ b/internal/config/lint.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"io/fs"
 	"os"
+	"strings"
 	"time"
 	"unicode/utf8"
 
@@ -16,9 +17,28 @@ import (
 // ReadYAMLFileLinted will attempt to read a configuration file path into a
 // structure. Returns an array of lint messages or an error.
 func ReadYAMLFileLinted(fs ifs.FS, spec docs.FieldSpecs, path string, skipEnvVarCheck bool, lConf docs.LintConfig) (Type, []docs.Lint, error) {
-	configBytes, lints, _, err := ReadFileEnvSwap(fs, path, os.LookupEnv)
-	if err != nil {
-		return Type{}, nil, err
+	var configBytes []byte
+	var lints []docs.Lint
+	var err error
+
+	if strings.HasPrefix(path, "https://") {
+		if configBytes, err = fetchRemoteConfig(path, nil); err != nil {
+			return Type{}, nil, err
+		}
+		if configBytes, err = ReplaceEnvVariables(configBytes, os.LookupEnv); err != nil {
+			var errEnvMissing *ErrMissingEnvVars
+			if errors.As(err, &errEnvMissing) {
+				configBytes = errEnvMissing.BestAttempt
+				lints = append(lints, docs.NewLintError(1, docs.LintMissingEnvVar, err))
+			} else {
+				return Type{}, nil, err
+			}
+		}
+	} else {
+		configBytes, lints, _, err = ReadFileEnvSwap(fs, path, os.LookupEnv)
+		if err != nil {
+			return Type{}, nil, err
+		}
 	}
 
 	if skipEnvVarCheck {
--- a/internal/config/reader.go
+++ b/internal/config/reader.go
@@ -5,7 +5,9 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"io"
 	"io/fs"
+	"net/http"
 	"os"
 	"path/filepath"
 	"strings"
@@ -84,11 +86,12 @@ type Reader struct {
 	changeFlushPeriod  time.Duration
 	changeDelayPeriod  time.Duration
 	filesRefreshPeriod time.Duration
+	configHeaders      []string
 }
 
 // NewReader creates a new config reader.
 func NewReader(mainPath string, resourcePaths []string, opts ...OptFunc) *Reader {
-	if mainPath != "" {
+	if mainPath != "" && !strings.HasPrefix(mainPath, "https://") {
 		mainPath = filepath.Clean(mainPath)
 	}
 	r := &Reader{
@@ -120,6 +123,14 @@ func NewReader(mainPath string, resourcePaths []string, opts ...OptFunc) *Reader
 
 // OptFunc is an opt function that changes the behaviour of a config reader.
 type OptFunc func(*Reader)
+
+// OptSetConfigHeaders sets HTTP headers to send when fetching a remote config
+// via HTTPS URL. Each entry should be in "Name: Value" format.
+func OptSetConfigHeaders(headers []string) OptFunc {
+	return func(r *Reader) {
+		r.configHeaders = headers
+	}
+}
 
 // OptSetFullSpec overrides the default general config spec with the provided
 // one.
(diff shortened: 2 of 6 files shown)


internal/config/lint.go
```diff
@@ -16,9 +17,28 @@ import (
...
 // structure. Returns an array of lint messages or an error.
 func ReadYAMLFileLinted(fs ifs.FS, spec docs.FieldSpecs, path string, skipEnvVarCheck bool, lConf docs.LintConfig) (Type, []docs.Lint, error) {
-	configBytes, lints, _, err := ReadFileEnvSwap(fs, path, os.LookupEnv)
-	if err != nil {
-		return Type{}, nil, err
+	var configBytes []byte
+	var lints []docs.Lint
+	var err error
+
+	if strings.HasPrefix(path, "https://") {
```
jem-davies, line 24:
```suggestion
    parsedPath, err := url.Parse(path)
    if err != nil {
        ... 
    }
    
    if parsedPath.Scheme == "https" || parsedPath.Scheme == "http" {
        ...
    }
```

Could use url.Parse here - if the provided path is a relative filepath `./config.yaml` scheme should be empty string, you could also have a scheme `file` too - `file:///Users/.../config.yaml`

internal/config/reader.go
```diff
@@ -448,3 +453,32 @@ func (r *Reader) TriggerMainUpdate(mgr bundle.NewManagement, strict bool, newPat
 	}
 	return nil
 }
+
+func fetchRemoteConfig(url string, headers []string) ([]byte, error) {
+	req, err := http.NewRequest(http.MethodGet, url, nil)
```
jem-davies, line 458: could use a `http.NewRequestWithContext()` and pass a context with a short deadline.

internal/config/reader.go
```diff
@@ -448,3 +453,32 @@ func (r *Reader) TriggerMainUpdate(mgr bundle.NewManagement, strict bool, newPat
...
+	}
+	for _, h := range headers {
+		name, value, ok := strings.Cut(h, ":")
+		if !ok {
+			return nil, fmt.Errorf("invalid config header %q: expected Name: Value format", h)
+		}
+		req.Header.Add(strings.TrimSpace(name), strings.TrimSpace(value))
+	}
+
+	client := &http.Client{
```
jem-davies, line 470: I think we might want to expose some more options for the client - such it could work with TLS certificates that are signed with private CA's etc.

jem-davies (changes requested):
Would need to document the new feature by changing the documentation strings for the `-c` / `--config` flag as well

internal/config/reader.go
Omar1H1: Hello @jem-davies 
Thanks for the feedback, i'm just wondering should this reuse existing CLI flags or 
env vars, or would you prefer dedicated flags like --config-tls-ca-file ?

jem-davies:
Hi @Omar1H1 

Me and @gregfurman have just been discussing this PR.

I am in the position that the config options for this new http client could be handled similarly to the `http` [root level config](https://warpstreamlabs.github.io/bento/docs/configuration/about) - perhaps with a field `http_client`. 

That it would be possible to provide this to Bento in a ./resources.yaml - such it could then subsequently fetch a stream config. 

Added multiple CLI flags I think wouldn't be the way to go, there is precedent for the majority of config be provided via YAML files. 

@gregfurman - perhaps was more critical that there is other ways to handle this situation i.e. sidecar / init containers, posting into bento running streams_mode, that we discussed on discord before. That adding this feature is moving the  complexity from already existing solutions into Bento. 

Also Greg mentioned that we could perhaps have the concept of 'hooks', such there would be an 'on_init' hook that would evoke some user-defined logic at start-up and there we could fetch the config.

Omar1H1:
Hi @jem-davies thanks for the detailed direction.

I think what you have suggested is the right way to go. If I may add, maybe naming it `remote_config` or `config_source` instead of `http_client` to make the intent clearer.

I totally understand @gregfurman concerns, the alternatives he mentioned are valid. That said, I do think there's value in making this a first-class feature rather than pushing it to the infrastructure layer. Happy to implement this direction if there's alignment. If not, that's completely fine and I can close the PR for now. maybe a clearer need will come up in the future.

Omar1H1:
Hello @jem-davies,
I have been looking at how to implement the `resources.yaml` approach and hit a sequencing problem.

Looking at `internal/config/reader.go` in the `Read()` function (line 214-222), the flow is strictly:

1. `readMain` runs first, it fetches and parses the main `-c` config
2. `readResources` runs second, it reads the `-r` resource files

```go
func (r *Reader) Read() ... {
    if conf, pConf, lints, lintWarns, err = r.readMain(r.mainPath); err != nil {
        return
    }
    // resources are read after main config is already parsed
    if rLints, rLintWarns, err = r.readResources(&conf.ResourceConfig); err != nil {
        return
    }
}
```

So a `remote_config` block in `resources.yaml` would be parsed too late, the main config has already been fetched before resources are even opened.

I see three possible directions:
**1. Two-pass read** : load resources before main config. High risk, touches the existing read flow, file watcher, and resource tracking.

**2. Local stub** : `remote_config` lives in the root `-c` config itself. A small local YAML file acts as a pointer:

```yaml
# local-stub.yaml
remote_config:
  url: https://config.internal/bento.yaml
  tls:
    root_cas_file: /certs/ca.pem
  headers:
    Authorization: "Bearer ${TOKEN}"
```

`readMain` detects the `remote_config` block, fetches the real config using the TLS and headers defined there, and parses the fetched bytes as the actual pipeline config. No flow changes needed, TLS and headers live in YAML exactly as you have suggested, env var interpolation still works on the stub before the fetch. Another advantage of option 2 is that linting works naturally `bento lint ./local-stub.yaml` would lint the fetched remote config since linting runs inside readMain after the fetch, giving the same validation guarantees as linting a local file. I'm leaning toward this approach.

**3. Something else entirely** : happy to follow whatever direction makes sense architecturally.

One open question for option 2: when the remote config is fetched, does it completely replace the stub, or do the stub's other fields (like `logger`, `http`) get merged into the fetched config? On the merge question; if the stub defines a `logger` block alongside `remote_config`, and the fetched config also defines `logger`, which one wins? Or is the stub treated purely as a pointer and discarded entirely after the fetch, with the remote config being the only source of truth?

PR #962, opened 2026-08-08
fix(fsevent): preserve file create events when watching new subdirs

### LINKED ISSUE #942: fsevent: watch_new_subdirs drops file CREATE events
## Description

When `watch_new_subdirs` is enabled, the `fsevent` input suppresses non-directory `CREATE` events while checking whether the created path should be added to the watcher.

This can silently miss newly created files inside dynamically created subdirectories. The behavior reproduces consistently on macOS, where creating the file may not produce a later `WRITE` event that would otherwise mask the dropped `CREATE`.

## Version and environment

- Bento: `main@eba3ed4192ea65bbe7bfc6b1d9fb124bea1483cb`
- Also present in the latest release, `v1.19.0`
- Go: `go1.26.5 darwin/arm64`
- OS: macOS / Darwin 25.5.0 arm64

## Minimal configuration

```yaml
input:
  fsevent:
    paths:
      - /path/to/watched-directory
    watch_new_subdirs: true
```

After starting the input:

1. Create a subdirectory below the watched directory.
2. Create a file inside that new subdirectory.
3. Observe that Bento emits the subdirectory `CREATE` event but no event for the file.

The existing repository test reproduces this twice:

```shell
go test ./internal/impl/io \
  -run 'TestFSEventWatchNewSubdirs$|TestFSEventWatchNewSubdirsDisabled$' \
  -count=2 -v
```

Result:

- `TestFSEventWatchNewSubdirs` fails twice with `Should have received event for file in subdirectory`.
- The nearby `TestFSEventWatchNewSubdirsDisabled` control passes twice.

## Expected behavior

Enabling `watch_new_subdirs` should add newly created directories to the watcher without suppressing filesystem events for ordinary files. The file creation should produce an `fsevent` message.

## Actual behavior

The event loop checks every `CREATE` event with `os.Stat`. When the path is not a directory, it executes `continue`, bypassing extension filtering and message emission:

https://github.com/warpstreamlabs/bento/blob/eba3ed4192ea65bbe7bfc6b1d9fb124bea1483cb/internal/impl/io/input_fsevent.go#L180-L202

## Suggested fix

Only call `watcher.Add` when the created path is a directory, but continue emitting the original event for ordinary files. A focused regression test can create an empty file with `watch_new_subdirs: true` and require its `CREATE` event.

I am happy to open a small PR with the fix and regression test if this approach looks right.

mattfaltyn:
## Summary

- preserve ordinary file CREATE events when watch_new_subdirs is enabled
- add new paths to the watcher only when they are confirmed directories
- keep emitting the original event when directory inspection fails
- strengthen the regression test with an empty file and an exact CREATE assertion

Closes #942

## Testing

- macOS fsevent enabled/disabled regression: 5 repeated passes plus a final post-lint pass
- Linux Go 1.26.5 container: relevant fsevent controls and regression passed twice from the exact commit
- full io package passed on macOS, excluding two upstream event-variant assertions validated by the Linux run
- full Cypher package passed against Colima; transient socket-server test passed 3 repeated runs
- all remaining Go packages passed in the repository-wide sweep
- go vet passed; golangci-lint v2.11.3 reported 0 issues
- component template lint and config tests passed

## Documentation

No documentation changes are required. The existing fsevent contract already states that CREATE events are emitted and that watch_new_subdirs watches events from newly created subdirectories.

### DIFF
--- a/internal/impl/io/input_fsevent.go
+++ b/internal/impl/io/input_fsevent.go
@@ -182,18 +182,12 @@ func (f *fsEventWatcher) Connect(ctx context.Context) error {
 					st, err := os.Stat(event.Name)
 					if err != nil {
 						f.log.Warnf("Cannot check for new subpath: %s", err)
-						continue
-					}
-
-					// if it is a file dont add it.
-					if !st.IsDir() {
-						continue
-					}
-
-					// adding the same path more than once is a noop,
-					// so safe even though it is already in the watchlist
-					if err := f.watcher.Add(event.Name); err != nil {
-						f.log.Warnf("Failed to add path %v: %s", event.Name, err)
+					} else if st.IsDir() {
+						// adding the same path more than once is a noop,
+						// so safe even though it is already in the watchlist
+						if err := f.watcher.Add(event.Name); err != nil {
+							f.log.Warnf("Failed to add path %v: %s", event.Name, err)
+						}
 					}
 				}
 
--- a/internal/impl/io/input_fsevent_test.go
+++ b/internal/impl/io/input_fsevent_test.go
@@ -234,7 +234,7 @@ fsevent:
 	time.Sleep(time.Second)
 
 	fileInSubdir := filepath.Join(subdir, "file.txt")
-	require.NoError(t, os.WriteFile(fileInSubdir, []byte("content"), 0o644))
+	require.NoError(t, os.WriteFile(fileInSubdir, nil, 0o644))
 
 	// We should receive events for both the subdir creation and the file creation
 	var subdirCreated, fileCreated bool
@@ -251,7 +251,7 @@ fsevent:
 
 			if eventPath == subdir && operation == "CREATE" {
 				subdirCreated = true
-			} else if eventPath == fileInSubdir && (operation == "WRITE" || operation == "CREATE") {
+			} else if eventPath == fileInSubdir && operation == "CREATE" {
 				fileCreated = true
 			}
 


mattfaltyn:
Final local review is green across macOS and Linux fsevent coverage. GitHub Actions is still waiting for maintainer approval before jobs can start: https://github.com/warpstreamlabs/bento/actions/runs/31248055706

internal/impl/io/input_fsevent.go
```diff
@@ -182,18 +182,12 @@ func (f *fsEventWatcher) Connect(ctx context.Context) error {
 					st, err := os.Stat(event.Name)
 					if err != nil {
 						f.log.Warnf("Cannot check for new subpath: %s", err)
-						continue
-					}
-
-					// if it is a file dont add it.
-					if !st.IsDir() {
-						continue
-					}
```
gregfurman, line 191: Ahh this was a good catch. Nice!

internal/impl/io/input_fsevent.go
```diff
@@ -182,18 +182,12 @@ func (f *fsEventWatcher) Connect(ctx context.Context) error {
 					st, err := os.Stat(event.Name)
 					if err != nil {
 						f.log.Warnf("Cannot check for new subpath: %s", err)
-						continue
```
gregfurman, line 185: Should we be falling through on an error? Feels like the event should probably be dropped.

internal/impl/io/input_fsevent.go
```diff
@@ -182,18 +182,12 @@ func (f *fsEventWatcher) Connect(ctx context.Context) error {
...
-					// if it is a file dont add it.
-					if !st.IsDir() {
-						continue
-					}
-
-					// adding the same path more than once is a noop,
-					// so safe even though it is already in the watchlist
-					if err := f.watcher.Add(event.Name); err != nil {
-						f.log.Warnf("Failed to add path %v: %s", event.Name, err)
+					} else if st.IsDir() {
```
gregfurman, line 185: nit(non-blocking): I personally find `else if` a bit tough to reason about at first glance. Think this is a sign that this flow could be refactored to maybe include an anonymous function, or extracted wholesale into a separate method/function.

gregfurman (approved):
LGTM! Reproduced the failing test suite and the minimal example in your original issue, and seems your changes fix both.

Thanks for the neat addition!

internal/impl/io/input_fsevent.go
mattfaltyn: Thanks!

internal/impl/io/input_fsevent.go
mattfaltyn: `Stat` only detects new directories here, so I’d preserve the fsnotify event when inspection fails.

internal/impl/io/input_fsevent.go
mattfaltyn: Thanks - I’d keep this single-use branch inline for now to avoid extra indirection.

PR #964, opened 2026-08-10
fix: support JSON Schema drafts 2019-09 and 2020-12

### LINKED ISSUE #963: JSON Schema validation silently ignores 2019-09/2020-12 keywords and `$ref` siblings
## Summary

Bento's JSON Schema validation runs on `github.com/xeipuuv/gojsonschema`, which implements drafts 4, 6 and 7. Schemas declaring draft 2019-09 or 2020-12 are neither rejected nor validated correctly — they load without complaint and are validated with the newer keywords skipped. The result is wrong verdicts with no error and no log line.

This affects the `json_schema` processor, the `json_schema` Bloblang method, and `schema_registry_decode`/`schema_registry_encode` for JSON schemas.

## Reproduction

Each schema below declares `$schema: https://json-schema.org/draft/2020-12/schema`. Behaviour under `gojsonschema v1.2.0` (current) and `santhosh-tekuri/jsonschema v6.0.3`:

| Schema feature | Instance | Correct verdict | gojsonschema | v6 |
|---|---|---|---|---|
| `{"$ref": "#/$defs/positive", "maximum": 10}` | `{"n": 50}` | invalid | **valid** | invalid |
| `prefixItems: [string, integer]` | `{"pair": [1, "x"]}` | invalid | **valid** | invalid |
| `unevaluatedProperties: false` | `{"a":"ok","extra":1}` | invalid | **valid** | invalid |
| `$anchor` + `$ref: "#myAnchor"` | `{"s": "hello"}` | valid | **invalid** | valid |

The first three are silent false-accepts: invalid data passes a validation gate. The fourth is a false-reject — `#myAnchor` fails to resolve and falls back to the root schema, so *every* value fails with the misleading message `s Invalid type. Expected: object, given: string`.

The `$ref`-with-siblings case needs no new keyword at all. Since 2019-09, keywords alongside `$ref` are applied; under draft-07 semantics `$ref` replaces them. So a schema using only long-standing keywords still gets a silently weaker constraint set.

<details>
<summary>Self-contained repro</summary>

```go
package main

import (
	"encoding/json"
	"fmt"
	old "github.com/xeipuuv/gojsonschema"
)

const schema = `{
  "$schema": "https://json-schema.org/draft/2020-12/schema",
  "$defs": { "positive": { "type": "integer", "minimum": 0 } },
  "type": "object",
  "properties": { "n": { "$ref": "#/$defs/positive", "maximum": 10 } }
}`

func main() {
	s, err := old.NewSchema(old.NewStringLoader(schema))
	if err != nil {
		panic(err)
	}
	var doc any
	_ = json.Unmarshal([]byte(`{"n": 50}`), &doc)
	res, _ := s.Validate(old.NewGoLoader(doc))
	fmt.Println("valid:", res.Valid()) // prints: valid: true — should be false
}
```

</details>

## Why it is silent

`GetDraftVersion` matches only three exact draft-4/6/7 `$schema` URLs (`draft.go:73-80`). Any other value — including both 2019-09 and 2020-12 — leaves the compiler in `Hybrid` mode (`draft.go:33`), which accepts keywords from every draft at once and ignores ones it does not recognise. Metaschema validation is off by default, so nothing rejects the schema at load time either.

A schema that happens to avoid the newer keywords therefore validates correctly, but coincidentally rather than by design, and stops doing so the moment one is introduced.

## No upgrade path

`xeipuuv/gojsonschema` has published exactly two versions — `v1.1.0` and `v1.2.0`. `v1.2.0` is from 2019-10-15 and upstream's last commit is 2020-10-27. There is no newer release to move to, and no 2019-09/2020-12 support to gain by bumping.

## Proposed fix

Replace it with `github.com/santhosh-tekuri/jsonschema/v6` across all four call sites in one change.

- Drafts 4/6/7/2019-09/2020-12, selected from `$schema`. It is the de facto standard Go implementation, which matters more for a validation gate than for an ordinary dependency.
- Apache-2.0, same licence as the outgoing library. Bento stays MIT.
- Latest release v6.0.3 (2026-06-28).
- **Net −2 modules.** Drops `gojsonschema` plus `xeipuuv/gojsonpointer` and `xeipuuv/gojsonreference`, which reach Bento only through it (`go mod why -m`) and are pinned to 2018/2019 pseudo-versions. Adds one module whose two requirements, `golang.org/x/text` and `dlclark/regexp2`, are already in `go.mod` — no new transitive dependencies.
- Remote `$ref`s stop being fetched by default. Today any non-`file://` reference goes through a bare `http.Get` on the default client with no timeout (`jsonLoader.go:146-204`), so a schema can reach the network at pipeline startup. v6 defaults to a file-only loader; `schema_path: http://` keeps working through an explicitly installed loader with a timeout.

## Breaking changes

Both are user-visible, so they need agreement before a PR:

**Schemas relying on `Hybrid` leniency stop compiling.** No `DefaultDraft` setting reproduces Hybrid — it accepts draft-4 `id` beside `$id`, and boolean and numeric `exclusiveMinimum` in the same document. A schema without `$schema` using draft-4 spellings compiles today and will not afterwards. This fails at pipeline startup rather than at runtime.

**Schemas already declaring 2019-09/2020-12 get stricter.** That is the point of the change, but messages passing validation today may start failing, so it belongs in the changelog rather than only in the PR.

Error message text also changes. No compatibility shim is proposed: gojsonschema's messages come from a ~50-template locale table, and reproducing it would mean tracking v6 releases indefinitely. v6's output is tree-structured and attributes each failure to its `oneOf`/`anyOf` branch, which the current flat list does not.

## Implementation note

gojsonschema's `NewGoLoader` round-trips the instance through `json.Marshal` and re-decodes with `UseNumber` (`jsonLoader.go:302-313`), normalising `[]byte` to base64 and `time.Time` to RFC3339. v6's `Validate(any)` does not, and rejects such values outright. Since `[]byte` is a first-class Bloblang value, a naive port silently breaks validation of messages carrying bytes or timestamps. The migration needs an explicit normalisation step at both `Validate` call sites.

## Questions

1. Is the breaking-change framing acceptable, and does it want a `**Breaking:**` changelog entry alongside the existing one in `## Unreleased`?
2. Should `schema_registry` validation reach the network for `$ref`s at all, or should registry references resolve only from the registry? Today an unresolvable absolute `$ref` there falls through to a silent HTTP GET.
3. `schema_path` accepts `file://` and `http://` but rejects `https://` (`processor_jsonschema.go:121`). Widen as part of this, or leave it?

Happy to open the PR — the work is scoped and I have the four sites mapped. Wanted the direction and the breaking-change framing agreed first, per CONTRIBUTING.

slachiewicz:
Fixes #963.

Draft while the breaking-change framing in the issue is still open — the three questions there decide whether this lands as-is.

`xeipuuv/gojsonschema` implements drafts 4/6/7 and has no newer release, so this swaps it for `santhosh-tekuri/jsonschema/v6` across the `json_schema` processor, the Bloblang method, and the schema registry path. The shared wrapper in `internal/jsonschema` exists so compilation, instance handling and error rendering stay identical between the three.

## Things a reviewer would otherwise read as mistakes

Instances are round-tripped through `encoding/json` before validation. `NewGoLoader` did this implicitly, and it is what renders `[]byte` as base64 and `time.Time` as RFC3339 — both still reachable as Bloblang values, so removing the round trip would silently break validation of messages carrying bytes or timestamps.

The file loader trims the `file://` prefix instead of parsing the URL, because `url.Parse` reads `..` in a scheme-relative path as the host and drops segments. `TestJSONSchemaExternalSchemaRelativePath` depends on that not happening.

Registry schemas compile under a synthetic `registry:///` base, so a relative `$ref` resolves to a sibling reference name rather than a path on the host running Bento.

## Behaviour changes

Schemas declaring 2019-09 or 2020-12 are validated with those drafts rather than having the newer keywords silently skipped, so messages that pass today may start failing. Schemas without a `$schema` that mix draft-4 spellings no longer compile — that fails at pipeline startup, not at runtime.

Error text changes and is not shimmed: the old wording came from a locale table with no upstream equivalent, and reproducing it would mean tracking v6 releases indefinitely. Updated in the two tests and the Bloblang example; `website/docs/guides/bloblang/methods.md` is regenerated via `make docs`, not hand-edited.

A second commit fixes an asymmetry found after the first: `CompilePath` registered file and http loaders but `CompileString` used the bare compiler, so an absolute `$ref` inside an inline `schema:` failed to compile where the previous implementation fetched it. Inline schemas now get the same loaders rather than `schema_path` losing them, so the swap removes no capability. Both paths gained a request timeout, which the previous default client did not have.

Not included: a field mapping schema URLs to local files, so a pipeline need not reach the network at startup at all. It belongs on top of this change rather than inside it, and can only be built once the new compiler is in. Happy to add it here if you would rather review them together.

With #966 merged, all three `xeipuuv/*` modules leave `go.mod`; the production binary no longer contains any `xeipuuv/*` package (`go list -deps ./cmd/bento`). They stay in `go.sum` only through `trino-go-client`'s test dependencies until #1053 lands.

Verified: `golangci-lint run -c .golangci.yml` over the changed packages, 0 findings in changed files. Note `make lint` runs with `--fix` and rewrites ~24 unrelated files carrying pre-existing `modernize` findings; those are deliberately not included.

Full `make test` was not run green locally — `internal/impl/cypher` needs Docker and `TestFSEvent*` fails on macOS, both confirmed failing identically on an unmodified `upstream/main` worktree. CI is the authority on the suite.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ All notable changes to this project will be documented in this file.
 ### Fixed
 
 - `parse_big_decimal` rejects scales above 16383 to avoid unbounded big-integer work @aratz-lasa
+- **Breaking:** JSON Schema validation understands drafts 2019-09 and 2020-12, replacing `github.com/xeipuuv/gojsonschema` with `github.com/santhosh-tekuri/jsonschema/v6`. Schemas declaring either draft previously loaded with the newer keywords silently ignored, so `json_schema` checks that pass today may start failing. Schemas without a `$schema` that mix draft-4 spellings no longer compile, and validation error messages have changed shape @slachiewicz
 
 ### Security
 
--- a/go.mod
+++ b/go.mod
@@ -130,6 +130,7 @@ require (
 	github.com/redis/go-redis/v9 v9.18.0
 	github.com/rickb777/period v1.0.7
 	github.com/robfig/cron/v3 v3.0.1
+	github.com/santhosh-tekuri/jsonschema/v6 v6.0.3
 	github.com/segmentio/ksuid v1.0.4
 	github.com/sijms/go-ora/v2 v2.8.22
 	github.com/sirupsen/logrus v1.9.4
@@ -146,7 +147,6 @@ require (
 	github.com/urfave/cli/v2 v2.27.7
 	github.com/vmihailenco/msgpack/v5 v5.4.1
 	github.com/xdg-go/scram v1.1.2
-	github.com/xeipuuv/gojsonschema v1.2.0
 	github.com/youmark/pkcs8 v0.0.0-20240726163527-a2c0da244d78
 	go.etcd.io/etcd/api/v3 v3.6.5
 	go.etcd.io/etcd/client/v3 v3.6.5
@@ -238,6 +238,7 @@ require (
 	github.com/spiffe/go-spiffe/v2 v2.6.0 // indirect
 	github.com/theparanoids/crypki v1.20.9 // indirect
 	github.com/twpayne/go-geom v1.6.1 // indirect
+	github.com/xeipuuv/gojsonschema v1.2.0 // indirect
 	github.com/yalue/onnxruntime_go v1.30.1 // indirect
 	go.opentelemetry.io/auto/sdk v1.2.1 // indirect
 	go.opentelemetry.io/otel/exporters/otlp/otlpmetric/otlpmetricgrpc v1.43.0 // indirect
--- a/go.sum
+++ b/go.sum
@@ -1777,6 +1777,8 @@ github.com/ruudk/golang-pdf417 v0.0.0-20181029194003-1af4ab5afa58/go.mod h1:6lfF
 github.com/ruudk/golang-pdf417 v0.0.0-20201230142125-a7e3863a1245/go.mod h1:pQAZKsJ8yyVxGRWYNEm9oFB8ieLgKFnamEyDmSA0BRk=
 github.com/s2-streamstore/s2-sdk-go v0.13.3 h1:JBZFchhA45XMBdQxpy7a/BbAeS1ZJloVtUbtU6hwllM=
 github.com/s2-streamstore/s2-sdk-go v0.13.3/go.mod h1:1a+v2sGqU+s5neI8XwqRJz78ktStkR+mZH/JEi9HNSo=
+github.com/santhosh-tekuri/jsonschema/v6 v6.0.3 h1:1EYB5IzjZawrrnELUi78f9fPu57HuXjmddZPjrls/28=
+github.com/santhosh-tekuri/jsonschema/v6 v6.0.3/go.mod h1:JXeL+ps8p7/KNMjDQk3TCwPpBy0wYklyWTfbkIzdIFU=
 github.com/schollz/progressbar/v3 v3.19.0 h1:Ea18xuIRQXLAUidVDox3AbwfUhD0/1IvohyTutOIFoc=
 github.com/schollz/progressbar/v3 v3.19.0/go.mod h1:IsO3lpbaGuzh8zIMzgY3+J8l4C8GjO0Y9S69eFvNsec=
 github.com/segmentio/asm v1.2.1 h1:DTNbBqs57ioxAD4PrArqftgypG4/qNpXoJx8TVXxPR0=
--- a/internal/bloblang/query/methods_structured.go
+++ b/internal/bloblang/query/methods_structured.go
@@ -9,7 +9,7 @@ import (
 	"strings"
 
 	"github.com/Jeffail/gabs/v2"
-	jsonschema "github.com/xeipuuv/gojsonschema"
+	"github.com/warpstreamlabs/bento/internal/jsonschema"
 
 	"github.com/warpstreamlabs/bento/internal/value"
 )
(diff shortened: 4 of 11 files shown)


slachiewicz:
Two things, one fixed here and one deliberately left out.

**Inline schemas could not resolve absolute `$ref`s.** `CompilePath` registered file and http loaders but `CompileString` used the bare compiler, whose default resolves `file` URLs only, so an absolute `$ref` inside an inline `schema:` failed to compile where the previous implementation fetched it. Verified with a local server before and after: inline returned `invalid file url: http://...` while `schema_path` resolved. Fixed in 3ca71a7 by giving inline schemas the remote loaders rather than removing them from `schema_path`, so the swap takes away no capability anyone has today. Both paths now carry a request timeout, which the previous implementation's default client did not. Regression test covers both entry points so they cannot drift apart again.

**Not included: a field to avoid the network entirely.** The fix above restores parity, but it leaves a schema with absolute `$ref`s making outbound HTTP when a pipeline starts, with no way to opt out — an availability and supply-chain concern for something whose job is to be a validation gate. The concrete case is OpenLineage, whose 37 facet schemas each reference the root by absolute `https://openlineage.io/...` URL, so validating those events either depends on that host being reachable at startup or requires pre-bundling all 37 into one document.

The fix is a config field mapping URL prefixes to local paths, backed by the compiler's `AddResource`; the library's own CLI models it as `jv --map url_prefix=/path/to/dir`. I have left it out of this PR because it adds user-facing config surface to a change whose breaking-change framing is still open in #963, and because it can only be built on the new compiler — it cannot land before this does. Happy to add it here instead if you would rather review them together.

Worth flagging for the same reason: if you would prefer this whole swap be opt-in — a new field or a separate processor rather than a change in place — say so and I will rework it. My reservation is that the defect in #963 is silent under-enforcement, so an opt-in leaves the wrong behaviour as the default, but that is your call rather than mine.

Issue #965, opened 2026-08-10
Upgrade ory/dockertest to v4 to drop docker/cli and the xeipuuv modules


slachiewicz:
## Summary

`ory/dockertest/v3` is the only thing pulling `docker/cli` into the module graph, and `docker/cli`'s compose-file loader is in turn the only remaining consumer of `xeipuuv/gojsonschema` and its two helper modules. Upgrading to `dockertest/v4` drops all four.

```
$ go mod why -m github.com/docker/cli
github.com/warpstreamlabs/bento/internal/impl/amqp09.test
github.com/ory/dockertest/v3
github.com/ory/dockertest/v3/docker/opts
github.com/docker/cli/cli/compose/loader

$ go mod why -m github.com/xeipuuv/gojsonschema
... github.com/docker/cli/cli/compose/schema → github.com/xeipuuv/gojsonschema
```

`v4.0.0` (2026-04-10) replaces the vendored `go-dockerclient` fork and `docker/cli` with `moby/moby/api` and `moby/moby/client`. Its `go.mod` requires neither `docker/cli` nor any `xeipuuv` module.

`docker/docker` does **not** leave with them. It has two other sources: `apache/pulsar-client-go`'s tests, which moved to `moby/moby/api` in v0.21.0 (apache/pulsar-client-go#1513) but which Bento pins at v0.17.0, and `trinodb/trino-go-client`, which requires `dockertest/v3` in its own `go.mod`.

Current pins are `docker/cli v25.0.3+incompatible` and `dockertest v3.10.0`; the `+incompatible` Docker libraries are the ones with a CVE history, which is the underlying motivation.

## Scope

This is not an import-path bump. v4 is an API redesign. Measured across the tree:

| Call site | Occurrences |
| --- | --- |
| `dockertest.NewPool` | 82 |
| `pool.Purge` | 84 |
| `pool.Retry` | 81 |
| `MaxWait` | 91 |
| `pool.RunWithOptions` | 58 |
| `dockertest.RunOptions` | 58 |
| `pool.Client` | 6 |

53 files import dockertest; 13 of those also import `dockertest/v3/docker`, the vendored client v4 deletes. That subpackage is used narrowly — `docker.Port` and `docker.PortBinding` (24 each), `docker.HostConfig` (2), `docker.RestartPolicy` and `docker.CreateNetworkOptions` (1 each) — so those map onto `moby/moby/api/types/container` equivalents.

The API changes, per the library's own `UPGRADE.md`:

[...]

PR #966, opened 2026-08-10
test: bump dockertest to v4, drop the old vendored docker/cli

slachiewicz:
Implements #965. **No test imports `ory/dockertest/v3` any more, and it leaves `go.mod`.** 57 files across 26 packages.

## Correct the expected benefit first

#965 claimed dockertest v3 was the only thing pulling `docker/cli` into the graph. That was wrong, and finishing the migration proved it: `trinodb/trino-go-client` requires `ory/dockertest/v3` in its own `go.mod`, so `docker/cli` and the three `xeipuuv` modules remain regardless. Details in [this comment](https://github.com/warpstreamlabs/bento/issues/965#issuecomment-5243819599).

What this actually delivers: Bento's integration tests stop running on a vendored Docker client last released in 2019, carrying the CVEs v4 exists to escape; `docker/cli` becomes a purely third-party transitive dependency rather than a consequence of our own; and with #964, no `xeipuuv` package executes in the shipped binary.

Draft until #965 has a direction on the container-reuse default.


slachiewicz:
@gregfurman would you be able also to check this PR?

gregfurman:
@slachiewicz Thanks for the ping! Am giving this a look now

internal/impl/azure/integration_test.go
```diff
@@ -264,7 +261,7 @@ func TestIntegrationCosmosDB(t *testing.T) {
 	_, servicePort, err := net.SplitHostPort(listener.Addr().String())
 	require.NoError(t, err)
 
-	err = pool.Retry(func() error {
+	err = pool.Retry(t.Context(), 0, func() error {
```
gregfurman, line 264: super-nit: A comment saying that `0` means we default to the `pool.MaxWait` would be useful

internal/impl/elasticsearch/aws/integration_test.go
```diff
@@ -218,28 +217,26 @@ func GetLocalStack(t testing.TB, envVars []string, readyFns ...func(port string)
...
-
-	resource, err := pool.RunWithOptions(&dockertest.RunOptions{
-		Repository:   lsImageName,
-		Tag:          "4.9.2", // pinning version: latest needs a license.
-		ExposedPorts: []string{"4566/tcp"},
-		PortBindings: map[docker.Port][]docker.PortBinding{
-			docker.Port(port + "/tcp"): {
-				docker.PortBinding{HostIP: "", HostPort: port},
+	resource := pool.RunT(t, lsImageName,
+		dockertest.WithTag("4.9.2"), // pinning version: latest needs a license.
```
gregfurman, line 221: nit: Unrelated to this PR but the last version to require a license was `4.14.0`


```suggestion
		dockertest.WithTag("4.14.0"), // pinning version: latest needs a license.
```

internal/impl/gcp/integration_bigquery_test.go
```diff
@@ -51,36 +52,35 @@ func TestIntegrationBigQuery(t *testing.T) {
...
-		ExposedPorts: []string{"9050/tcp", "9060/tcp"},
-		Cmd:          []string{"--project", "test-project"},
-	}
+	pool := dockertest.NewPoolT(t, "", dockertest.WithMaxWait(maxWait))
 
 	if runtime.GOOS == "darwin" {
-		runOpts.Platform = "linux/amd64"
+		// v4 has no Platform option; pre-pulling picks the architecture because
+		// Run skips its own pull once the image inspects cleanly.
+		pullAMD64(t, pool, "ghcr.io/goccy/bigquery-emulator:latest")
```
gregfurman, line 65: praise: neat workaround for this lack of a platform option 👍

internal/impl/kafka/input_kafka_franz_test.go
```diff
@@ -174,41 +174,38 @@ func TestIntegrationFranzInputDetectUnknownTopicError(t *testing.T) {
...
+	pool, err := dockertest.NewPool(t.Context(), "", dockertest.WithMaxWait(time.Minute))
 	if err != nil {
 		t.Skipf("Could not connect to docker: %s", err)
 	}
+	t.Cleanup(func() {
+		// Mirror NewPoolT: a cleanup failure is logged, not fatal.
+		if err := pool.Close(context.WithoutCancel(t.Context())); err != nil {
+			t.Logf("pool.Close() error: %v", err)
+		}
+	})
```
gregfurman, line 186: Is there a reason why we aren't just using `NewPoolT`? Like, if we're always going to be ignoring context cancellation of `t.Context`, I think it'd be simpler to just use the single fixture that'll automatically handle all of this for us.

gregfurman (approved):
LGTM! Big kudos for taking on this whole migration 🥳 

Have left a couple of comments. Nothing blocking, but I am wondering whether we should just be using that `NewPoolT` fixture instead of current approach of mirroring this with an uncancellable context (derived from `t.Context()`).

Perhaps there's something I'm missing here though so once clarified I'm happy to merge :shipit:

internal/impl/azure/integration_test.go
slachiewicz: Done in #1052.

internal/impl/elasticsearch/aws/integration_test.go
slachiewicz: Done in #1052.

internal/impl/kafka/input_kafka_franz_test.go
slachiewicz: Done in #1052.

PR #967, opened 2026-08-10
query: resolve all-map Bloblang field paths without gabs

slachiewicz:
Field reads in a Bloblang mapping go through `gabs.Wrap(target).S(path...).Data()`, which builds and discards a `*gabs.Container` per path segment. Allocations therefore scale with the number of field reads in a mapping rather than staying flat. Walking `map[string]any` directly removes that.

Arrays are deliberately left on gabs. `Container.S` calls `searchStrict` with `allowWildcard=true`, so a `*` segment over an array is real behaviour; `resolveFieldPath` hands the whole path back to gabs the moment it meets anything that is not a `map[string]any`, rather than reimplementing it. Equivalence against gabs is asserted directly in `TestResolveFieldPathMatchesGabs` over missing keys, nil intermediates, scalars mid-path, negative and out-of-range indices, wildcards on both arrays and maps, and numeric-looking map keys.

The first commit adds benchmarks because there were none for this path — only the parser and iterator had any, so `Executor.Query` and `Overlay` were unmeasured.

`benchstat`, 10 runs per side, darwin/arm64 (M1). Allocation counts are exact and stable (±0% on every case):

```
                        │  allocs/op before │  allocs/op after   vs base            │
ExecutorQuery/field_read         5.000              4.000        -20.00% (p=0.000)
ExecutorQuery/predicate          5.000              4.000        -20.00% (p=0.000)
ExecutorQuery/and2               6.000              4.000        -33.33% (p=0.000)
ExecutorQuery/and3               8.000              4.000        -50.00% (p=0.000)
ExecutorQuery/contains          13.00              12.00          -7.69% (p=0.000)
ExecutorQuery/exists             6.000              5.000        -16.67% (p=0.000)
ExecutorQuery/multi_assign      12.000              9.000        -25.00% (p=0.000)
ExecutorOverlay                  8.000              7.000        -12.50% (p=0.000)
ExecutorQueryWithVars            6.000              5.000        -16.67% (p=0.000)
geomean                          7.229              5.530        -23.50%

B/op geomean                     231.2              198.6        -14.10%
```

`and3` is the shape that shows the point: 8 allocations to 4, and constant from there regardless of how many fields the mapping reads.

On wall-clock I am claiming less. My timing run was noisy — several cases came back with ±65–85% variance and no significant delta, while `and3` showed -16.82% and the geomean -7.39%. The allocation reduction is the solid result here; treat the latency figures as indicative and worth re-running on a quiet machine.

Two allocation sites remain on this path and are untouched here: the `Vars: map[string]any{}` built per call in `public/bloblang/executor.go` even when a mapping declares no variables, and `FunctionContext.WithValue` taking `any` by value and storing `&value`, which forces a heap escape per call. Both look addressable; I left them out to keep this reviewable.

Verified: `golangci-lint run internal/bloblang/query/... public/bloblang/...` reports 0 issues, and `go test ./internal/... ./public/...` passes apart from the `TestFSEvent*` cases in `internal/impl/io`, which fail identically on an unmodified `main` on this machine (Darwin fsnotify reporting CHMOD where the tests expect WRITE) — pre-existing and unrelated.

The same change is proposed upstream at redpanda-data/benthos#472.

---

**Disclosure:** this change was developed with AI assistance (Claude Code) and reviewed by me before submission.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ All notable changes to this project will be documented in this file.
 
 ## Unreleased
 
+### Changed
+
+- Bloblang field reads over maps no longer allocate per path segment, reducing allocations in mappings that read several fields @slachiewicz
+
 ### Fixed
 
 - `parse_big_decimal` rejects scales above 16383 to avoid unbounded big-integer work @aratz-lasa
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ All notable changes to this project will be documented in this file.
 ### Changed
 
 - Bloblang field reads over maps no longer allocate per path segment, reducing allocations in mappings that read several fields @slachiewicz
+- Bloblang mappings that assign no variables no longer allocate a variables map per execution @slachiewicz
 
 ### Fixed
 
--- a/internal/bloblang/mapping/executor.go
+++ b/internal/bloblang/mapping/executor.go
@@ -222,6 +222,22 @@ func (e *Executor) QueryTargets(ctx query.TargetsContext) (query.TargetsContext,
 	return ctx, paths
 }
 
+// AssignsVariables returns true if any statement within the mapping assigns to
+// a variable, and therefore whether executing it requires a writable variables
+// map.
+//
+// Maps applied with the `apply` method are deliberately not consulted: that
+// method isolates variables by replacing the map before invoking the target, so
+// a named map cannot write to its caller's variables.
+func (e *Executor) AssignsVariables() bool {
+	for _, t := range e.AssignmentTargets() {
+		if t.Type == TargetVariable {
+			return true
+		}
+	}
+	return false
+}
+
 // AssignmentTargets returns a slice of all targets assigned to by statements
 // within the mapping.
 func (e *Executor) AssignmentTargets() []TargetPath {
--- a/internal/bloblang/query/functions.go
+++ b/internal/bloblang/query/functions.go
@@ -74,7 +74,33 @@ func (f *fieldFunction) Exec(ctx FunctionContext) (any, error) {
 	if len(f.path) == 0 {
 		return target, nil
 	}
-	return gabs.Wrap(target).S(f.path...).Data(), nil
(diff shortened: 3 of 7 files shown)


slachiewicz:
Added the variables-map commit mentioned above, and closing out the third idea as a dead end.

**Variables map.** `AssignsVariables` walks the statements' existing `AssignmentTargets` for a `TargetVariable`; mappings without one share a single empty map. Two things make this narrower than it first looked: `apply` already replaces `ctx.Vars` before invoking a named map (the `// ISOLATED VARIABLES` line in `methods.go`), so named maps cannot write to their caller's variables and need no analysis; and the map has to be empty rather than nil, because a nil `Vars` reports `variables were undefined` where an empty one reports `variable 'x' undefined`. Tests cover both, plus a race-detector run and a case asserting a var-assigning mapping still gets a private map.

Cumulative against the branch point, allocations per `Query`:

| mapping | before | after |
|---|---|---|
| `this.a."b"."c" != "x"` | 5 | **3** |
| 3-term `&&` over 4 field reads | 8 | **3** |
| several `root.x =` assignments | 12 | 8 |
| `Overlay` | 8 | 6 |

B/op on the simple predicate goes 120 → 56.

**The dead end**, for anyone who tries it next: `FunctionContext.WithValue` takes `any` by value and stores `&value`, which looked like a free heap escape to remove. It is not. Adding a `WithValuePtr(*any)` and passing `&val` from `Query` moves the escape rather than removing it — `-gcflags=-m` then reports `moved to heap: val`, and the benchmark is unchanged to the nanosecond (3 allocs, 56 B/op both ways). Removing that one needs `FunctionContext` to hold the value directly instead of behind `*any`, which changes `Value()` and every caller — not worth it for 16 bytes, in my view, but flagging it so the option is on the record.

On wall-clock I am claiming nothing for this commit: this machine sat at load average 8 on 8 cores while I measured, and an interleaved A/B put the latency delta inside the noise (p=0.33–0.65 at ±22–40% variance). The allocation counts are exact and reproducible; the timing needs a quiet machine.

PR #969, opened 2026-08-11
Resolve YAML merge keys when converting a config node to a map

### LINKED ISSUE #872: lint accepts YAML merge keys (<<) but config init ignores them — config lints clean, fails to start
`bento lint` accepts a config that uses YAML merge keys (`<<: *anchor`), but at startup the merge is **not** applied — the merged-in fields are silently dropped, and init fails with a `field '…' is required` error for a field that *was* present via the merge. The result is a config that passes `lint` (e.g. in CI) but fails to boot on deploy.

### Minimal repro (no external services needed)

```yaml
input:
  generate: { count: 1, interval: "", mapping: 'root.id = "x"' }
output:
  broker:
    outputs:
      - sql_raw: &base
          driver: postgres
          dsn: "postgres://localhost/x"
          query: "SELECT 1"
      - sql_raw:
          <<: *base          # merge key — intended to inherit driver/dsn
          query: "SELECT 2"
```

```
$ bento lint mergebug.yaml      # exit 0, no errors
$ bento -c mergebug.yaml
... Service closing due to: output 1: failed to init output <no label> path root.output.broker.outputs.1: field 'driver' is required and was not present in the config
```

### Expected

Either merge keys are honored at init (the 2nd output inherits `driver`/`dsn`), **or** `lint` rejects/warns on `<<` merge keys so the failure surfaces at lint time rather than on deploy.

### Actual

`lint` passes; init drops the merged fields and fails with a misleading "field is required" error pointing at a field that was supplied via the merge.

### Version

1.18.0

### Likely cause

go-yaml v3 dropped YAML-1.1 merge-key (`<<`) support. The lint path and the config-init path appear to diverge in how they handle the (unresolved) `<<` key.

youdie006:
### Summary

Fixes #872. A config that uses YAML merge keys (`<<: *anchor`) passes `bento lint` but fails at startup with the merged fields reported missing (e.g. `field 'driver' is required and was not present`).

### Root cause

Linting resolves merge keys: `FieldSpecs.LintYAML` checks `Content[i].Tag == "!!merge"` and recurses into the alias, so merged fields are seen. But `FieldSpecs.YAMLToMap` (used to build the `ParsedConfig` every component reads at init) did not: it iterated the raw key/value pairs, and the `<<` key matched no field spec, so it stored the anchor's contents under a literal `"<<"` key. The merged fields never reached the top level and, being required-with-no-default, init errored.

### Fix

Resolve merge keys before mapping fields (`resolveMergeKeys`), mirroring the lint behaviour. Explicit keys take precedence over merged ones (YAML merge semantics), and both the single-alias (`<<: *a`) and sequence (`<<: [*a, *b]`) forms are handled. The existing non-merge path is unchanged.

### Test

Added `TestFieldsNodeToMapMergeKeys` to `internal/docs`: a config whose fields come via `<<: *defaults` with one explicit override. Before the fix `YAMLToMap` errors (`field 'driver' is required`); after, the merged and overriding fields resolve correctly. `go test ./internal/docs/` passes with no regressions.

---

Disclosure: prepared with AI assistance (Claude); I reviewed it and verified the red-green test and package suite.

### DIFF
--- a/internal/docs/format_yaml.go
+++ b/internal/docs/format_yaml.go
@@ -977,9 +977,7 @@ func (f FieldSpecs) YAMLToMap(node *yaml.Node, conf ToValueConfig) (map[string]a
 	resultMap := map[string]any{}
 
 	// Resolve YAML merge keys ("<<") so that fields supplied via an anchor are
-	// recognised, matching how linting resolves them (LintYAML). Without this a
-	// config that lints clean fails at init with the merged fields reported as
-	// missing. Explicit keys take precedence over merged ones.
+	// recognised. Explicit keys take precedence over merged ones.
 	seen := map[string]struct{}{}
 	for _, pair := range resolveMergeKeys(node) {
 		fieldName := pair.key.Value
--- a/internal/docs/format_yaml.go
+++ b/internal/docs/format_yaml.go
@@ -920,6 +920,49 @@ func (f FieldSpec) YAMLToValue(node *yaml.Node, conf ToValueConfig) (any, error)
 	return node, nil
 }
 
+type yamlNodePair struct {
+	key   *yaml.Node
+	value *yaml.Node
+}
+
+// mergeAliases returns the mapping nodes referenced by a "<<" merge-key value,
+// which may be a single alias or a sequence of aliases.
+func mergeAliases(node *yaml.Node) []*yaml.Node {
+	switch node.Kind {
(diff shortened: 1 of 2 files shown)


jem-davies (approved):
LGTM - just expanded the number of tests related to YAML merge keys

PR #978, opened 2026-08-14
feat(cache): add prefix field to aws_s3 and gcp_cloud_storage caches

ecordell:
Adds a `prefix` field to the `aws_s3` and `gcp_cloud_storage` caches, following the `redis` cache precedent.

The prefix is prepended to keys on all operations, pushed down natively when listing (`ListObjectsV2` `Prefix` / GCS `Query.Prefix`), and stripped from yielded keys so they round-trip through Get/Delete. This lets multiple caches share one bucket, since in the clouds the bucket is the heavy IAM/provisioning boundary.

The S3 integration suite now also runs against a prefixed cache, which exercises the round-trip and the listing strip.


gregfurman (commented):
I'm not convinced on these changes since (IMO) they pre-suppose that a cache resource contains irrelevant data that you wouldn't want to iterate over -- catering more for the case where the cache resource does not have exclusive read/write access to the underlying data source.

Perhaps this logic can instead be defined within the relevant components instead, similar to how the `redis` cache has a `prefix` attribute... but this also feels kinda off. e.g

```yml
gcp_cloud_storage:
  bucket: "bar"
  iter_keys:
    prefix: "/bar"
```

Question, if this were to be exposed to the user via config, how would you suppose its done?

ecordell:
IMO the prefix case isn't about non-exclusive access, it's more that in the clouds the bucket is the IAM/provisioning boundary, and it's heavy. It made sense to me to have one bucket for a whole Bento program with several caches sharing it under different key prefixes. 

Prior to _adding_ List support, this was completely do-able with caches just by manually adding key prefixes on Get/Set. So I see this as closing that gap.

As for config I could see this being used in the input introduced in https://github.com/warpstreamlabs/bento/pull/941, i.e.:

```yaml
input:
  cache:
    resource: mycache
    prefix: foo
```

wdyt?

ecordell:
heh, just got back to looking at:
https://github.com/warpstreamlabs/bento/pull/941#discussion_r3768237430

gregfurman:
> It made sense to me to have one bucket for a whole Bento program with several caches sharing it under different key prefixes.

Ahh I see what you're saying. Alright so reflecting on my original response then:

> Perhaps this logic can instead be defined within the relevant components instead, similar to how the redis cache has a prefix attribute... but this also feels kinda off. [...]

I think that cache component specific fields (e.g `aws_s3.prefix` + `aws_s3.start_at`) being used to determine filtering is appealing to me a lot more than exposing this behaviour in the public SDK -- being more declarative and inline with Bento's design philosophy.

So (in the S3 or Cloud Storage cases) multiple caches could be defined per bucket with `prefix` being used as the grouping mechanism to prevent overlap. Similar to the redis cache, we could also use automatically prepend `prefix` all incoming operations in addition to being used when listing + filtering keys.

e.g defining a cache with `prefix` and using it in your (proposed) cache input:

```yaml
input:
  cache:
    resource: aws_s3_foo

cache_resources:
  - label: aws_s3_foo
    aws_s3:
      bucket: "super_important_bucket"
      prefix: /foo
```

Then if you want to combine the streams of multiple caches, you could use a `broker` https://warpstreamlabs.github.io/bento/docs/components/inputs/broker e.g

```yaml
input:
  broker:
    inputs:
     - cache:
        resource: aws_s3_foo
     - cache:
        resource: aws_s3_bar

cache_resources:
  - label: aws_s3_foo
    aws_s3:
      bucket: "super_important_bucket"
      prefix: /foo

  - label: aws_s3_bar
    aws_s3:
      bucket: "super_important_bucket"
      prefix: /bar
```

So my suggestions in this PR would be to remove the filtering options from the public interface and instead extend relevant cache components to include a `prefix` field.

Would be curious to hear your thoughts!

internal/impl/gcp/cache_cloud_storage.go
```diff
@@ -21,7 +22,10 @@ func gcpCloudStorageCacheConfig() *service.ConfigSpec {
 		Field(service.NewStringField("bucket").
 			Description("The Google Cloud Storage bucket to store items in.")).
 		Field(service.NewStringField("content_type").
-			Description("Optional field to explicitly set the Content-Type.").Optional())
+			Description("Optional field to explicitly set the Content-Type.").Optional()).
+		Field(service.NewStringField("prefix").
```
gregfurman, line 26: Could we mark this as `Advanced()`? Same with the `aws_s3` one.

gregfurman (changes requested):
Looking great! Seems like there's a merge conflict. Once unblocked (and last feedback addressed) think this looks good to ship.

Have run integration tests locally and can confirm all works fine.

ecordell:
Thanks for the feedback! I think it's looking good now, lmk what you think

gregfurman (approved):
LGTM!

PR #981, opened 2026-08-18
Expose additional kafka franz opts

jem-davies:


internal/impl/kafka/input_kafka_franz.go
```diff
@@ -879,6 +908,7 @@ func (f *franzKafkaReader) Connect(ctx context.Context) error {
 		kgo.MetadataMaxAge(f.metadataMaxAge),
 		kgo.FetchMaxBytes(f.fetchMaxBytes),
 		kgo.FetchMaxPartitionBytes(f.fetchMaxPartitionBytes),
+		kgo.BrokerMaxReadBytes(f.brokerReadMaxBytes),
```
gregfurman, line 911: Q: Does franz-go set this to a default value since now we'll be defaulting to `100MiB`

internal/impl/kafka/input_kafka_franz.go
```diff
@@ -414,12 +426,29 @@ func newFranzKafkaReaderFromConfig(conf *service.ParsedConfig, res *service.Reso
...
+		return nil, err
+	}
+
+	brokerReadMaxBytes, err := humanize.ParseBytes(brokerReadMaxBytesStr)
+	if err != nil {
+		return nil, err
+	}
+	if brokerReadMaxBytes > uint64(math.MaxInt32) {
+		return nil, fmt.Errorf("invalid broker_read_max_bytes, must not exceed %v", math.MaxInt32)
+	}
```
gregfurman, line 449: nit: Seems like good justification for parsing bytes in humanized form via bloblang. Then it could be a LineRule. Probably a decent follow-up.

gregfurman (approved):

internal/impl/kafka/input_kafka_franz.go
jem-davies: Yeah that is right - I have set the default for the values to be the default of the kafka franz package

PR #983, opened 2026-08-18
aws_sqs: refresh in-flight visibility using the queue's timeout

### LINKED ISSUE #878: aws_sqs: in-flight visibility refresh hardcoded to 30s, ignores queue VisibilityTimeout (ReceiptHandleIsInvalid livelock under backlog)
**Bento version:** 1.18.0

## Summary

The `aws_sqs` input's in-flight visibility refresh (`update_visibility`, default `true`) re-sets every in-flight message's visibility to a hardcoded **30 s**, ignoring the queue's configured `VisibilityTimeout`. It tries to honor the queue value by reading a per-message `VisibilityTimeout` attribute, but SQS doesn't return that on `ReceiveMessage` (it's a queue-level attribute), so the lookup always misses and the 30 s default stands. Under a backlog large enough that the refresher can't cycle all in-flight handles within 30 s, messages lapse → are redelivered → the cached receipt handles go stale → `ChangeMessageVisibilityBatch` / `DeleteMessageBatch` fail with `ReceiptHandleIsInvalid`, and the queue never drains.

## Where

`internal/impl/aws/input_sqs.go`, `sqsInFlightTracker.AddNew`:

```go
handle := sqsInFlightHandle{
    timeoutSeconds: 30,   // hardcoded default
    ...
}
if timeoutStr, exists := m.Attributes[sqsiAttributeNameVisibilityTimeout]; exists {
    // never true — ReceiveMessage doesn't return VisibilityTimeout as a message attribute
    handle.timeoutSeconds = ...
}
```

`ackLoop → refreshCurrentHandles → updateVisibilityMessages` then issues `ChangeMessageVisibilityBatch` with that 30 s value on a 1 s ticker.

## Reproduce

1. Queue with `VisibilityTimeout=600` and a sizeable backlog (~15k+ messages).
2. `aws_sqs` input with defaults (`update_visibility: true`).
3. The queue won't drain: `ApproximateNumberOfMessagesNotVisible` stays high, `input_received` climbs to many× the message count (redelivery), the log fills with `Failed to update consumed SQS message '…' visibility, response code: ReceiptHandleIsInvalid`, and `output_error` stays `0` (the pipeline/output are healthy — it's purely the input's SQS lifecycle).

Raising the queue's `VisibilityTimeout` has no effect, because the input overrides it to 30 s.

## Impact

A correctly-configured queue (long `VisibilityTimeout` to cover processing) can't be drained under backlog. It's invisible at low volume (the 1 s refresher keeps up), so it only manifests under load.

## Workaround

`update_visibility: false` — the input then leaves visibility alone, messages keep the queue's `VisibilityTimeout`, deletes use valid handles, and the queue drains.

## Suggested fix

Fetch the queue's `VisibilityTimeout` once via `GetQueueAttributes` at `Connect` and use it as the in-flight refresh value (instead of the per-message attribute that's never present), falling back to the current 30 s if the fetch fails / `sqs:GetQueueAttributes` is not permitted. This realizes the existing intent (the code comment says *"keep the queue timeout setting refreshed"*), fixes both the ignored-setting and the backlog livelock, and is backward-compatible. A config override field could be layered on top if preferred.

I'm happy to open a PR along these lines — just let me know whether you'd prefer the auto-detect, a config field, or both.

ReguiguiMohamed:
Fixes #878.

The in-flight tracker stamped every handle with a hardcoded 30s. It tried to read the queue's value from `m.Attributes[VisibilityTimeout]`, but `ReceiveMessage` only returns message system attributes and `VisibilityTimeout` is a queue level one, so that branch never ran. A queue set to a longer timeout had it silently overridden, and under a backlog big enough that the refresher could not cycle every handle within 30s, messages lapsed, got redelivered, and the cached receipt handles went stale with `ReceiptHandleIsInvalid`.

`Connect` now reads the queue's `VisibilityTimeout` once through `GetQueueAttributes` and stamps handles with it, falling back to 30s when the call fails or `sqs:GetQueueAttributes` is not granted. That is the auto-detect option from the issue, so it stays backward compatible. Glad to add a config override instead, or on top, if you'd prefer.

Two supporting changes, both needed for the above rather than separate fixes:

- `GetQueueAttributes` added to the internal `sqsAPI` interface, with a passthrough on `customHeaderSQSClient` so the new call still carries `custom_request_headers`. Without it the call would skip them and fall back to 30s for anyone using that field.
- `mockSqsInput.GetQueueAttributes` already existed and already returned `queueTimeout`, but with a pre-SDK-v2 signature that never satisfied the interface. Corrected, so the existing tests now exercise the queue value.

The dead per-message lookup is gone.

### Testing

- `go vet ./...` clean, and `golangci-lint run -c .golangci.yml ./internal/impl/aws/...` clean.
- `go build ./...`, `bento template lint`, and `bento test ./config/test/...` all pass.
- `internal/impl/aws` passes in full, over several repeat runs.
- Reverting just the `AddNew` line back to the hardcoded 30 turns the new test red with `expected: 600, actual: 30`.
- No docs regeneration, since no config field was added.
- I did not run the integration suite.

Worth flagging: now that the queue value is honored, `TestSQSInput` refreshes at its 10s `queueTimeout` rather than a fixed 30s, so its tolerance for a stalled refresher is narrower. I hit one failure there that I could not reproduce across six later runs, and the mock's `panic("nope")`, which fires when an ack races a visibility refresh, looks like the likely trigger.

Separately, and unrelated to this change, `internal/impl/sql` fails three tests on my Windows machine (`TestBufferSQLiteCloseWithPending`, `TestBufferSQLiteCloseAfterNack`, `TestConnSettingsInitFiles`). They fail identically on main.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ All notable changes to this project will be documented in this file.
 ### Fixed
 
 - `parse_big_decimal` rejects scales above 16383 to avoid unbounded big-integer work @aratz-lasa
+- `aws_sqs` input refreshes in-flight message visibility with the queue's `VisibilityTimeout` instead of a hardcoded 30s, which could let messages lapse and be redelivered under a backlog @ReguiguiMohamed
 
 ### Security
 
--- a/internal/impl/aws/client_factory.go
+++ b/internal/impl/aws/client_factory.go
@@ -73,3 +73,13 @@ func (c *customHeaderSQSClient) SendMessageBatch(
 	allOptFns := append(c.getAPIOptions(), optFns...)
 	return c.sqsAPI.SendMessageBatch(ctx, params, allOptFns...)
 }
+
+// GetQueueAttributes wraps the standard GetQueueAttributes call and adds custom headers
+func (c *customHeaderSQSClient) GetQueueAttributes(
+	ctx context.Context,
+	params *sqs.GetQueueAttributesInput,
+	optFns ...func(*sqs.Options),
+) (*sqs.GetQueueAttributesOutput, error) {
+	allOptFns := append(c.getAPIOptions(), optFns...)
+	return c.sqsAPI.GetQueueAttributes(ctx, params, allOptFns...)
+}
--- a/internal/impl/aws/input_sqs.go
+++ b/internal/impl/aws/input_sqs.go
@@ -154,6 +154,7 @@ type sqsAPI interface {
 	DeleteMessageBatch(context.Context, *sqs.DeleteMessageBatchInput, ...func(*sqs.Options)) (*sqs.DeleteMessageBatchOutput, error)
 	ChangeMessageVisibilityBatch(context.Context, *sqs.ChangeMessageVisibilityBatchInput, ...func(*sqs.Options)) (*sqs.ChangeMessageVisibilityBatchOutput, error)
 	SendMessageBatch(context.Context, *sqs.SendMessageBatchInput, ...func(*sqs.Options)) (*sqs.SendMessageBatchOutput, error)
+	GetQueueAttributes(context.Context, *sqs.GetQueueAttributesInput, ...func(*sqs.Options)) (*sqs.GetQueueAttributesOutput, error)
 }
 
 type awsSQSReader struct {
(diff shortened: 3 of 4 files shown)


ReguiguiMohamed:
Closing in favour of a cleaner branch, reopening with the same fix.

PR #984, opened 2026-08-18
aws_sqs: refresh in-flight visibility using the queue's timeout

### LINKED ISSUE #878: aws_sqs: in-flight visibility refresh hardcoded to 30s, ignores queue VisibilityTimeout (ReceiptHandleIsInvalid livelock under backlog)

ReguiguiMohamed:
Fixes #878.

`AddNew` stamped every in-flight handle with 30s. The lookup below it read `m.Attributes[VisibilityTimeout]`, which `ReceiveMessage` never returns, so it never ran and a queue configured with a longer timeout had its value ignored.

`Connect` now reads `VisibilityTimeout` once through `GetQueueAttributes` and stamps handles with that, falling back to 30s when the call fails. This adds `GetQueueAttributes` to the `sqsAPI` interface and a passthrough on `customHeaderSQSClient` so the call still carries `custom_request_headers`, and drops the dead lookup.

`mockSqsInput.GetQueueAttributes` already returned `queueTimeout`, but its pre-SDK-v2 signature never satisfied the interface. Correcting it means the existing tests now cover the queue value.

`go vet ./...`, `golangci-lint run -c .golangci.yml ./internal/impl/aws/...`, `go build ./...`, `internal/impl/aws`, `bento template lint` and `bento test ./config/test/...` all pass. Putting the hardcoded 30 back turns the new test red with `expected: 600, actual: 30`.

### DIFF
--- a/internal/impl/aws/input_sqs.go
+++ b/internal/impl/aws/input_sqs.go
@@ -27,6 +27,7 @@ const (
 	sqsiFieldDeleteMessage        = "delete_message"
 	sqsiFieldResetVisibility      = "reset_visibility"
 	sqsiFieldUpdateVisibility     = "update_visibility"
+	sqsiFieldVisibilityTimeout    = "visibility_timeout"
 	sqsiFieldMaxNumberOfMessages  = "max_number_of_messages"
 	sqsiFieldCustomRequestHeaders = "custom_request_headers"
 
@@ -39,6 +40,7 @@ type sqsiConfig struct {
 	DeleteMessage        bool
 	ResetVisibility      bool
 	UpdateVisibility     bool
+	VisibilityTimeout    time.Duration
 	MaxNumberOfMessages  int
 	CustomRequestHeaders map[string]string
 }
@@ -57,6 +59,9 @@ func sqsiConfigFromParsed(pConf *service.ParsedConfig) (conf sqsiConfig, err err
 		return
 	}
 	if conf.UpdateVisibility, err = pConf.FieldBool(sqsiFieldUpdateVisibility); err != nil {
+		return
+	}
+	if conf.VisibilityTimeout, err = pConf.FieldDuration(sqsiFieldVisibilityTimeout); err != nil {
 		return
 	}
 	if conf.MaxNumberOfMessages, err = pConf.FieldInt(sqsiFieldMaxNumberOfMessages); err != nil {
@@ -109,6 +114,11 @@ You can access these metadata fields using
 				Description("Whether to periodically refresh the visibility timeout of in-flight messages to prevent more-than-once delivery while still processing.").
 				Version("1.6.0").
 				Default(true).
+				Advanced(),
+			service.NewDurationField(sqsiFieldVisibilityTimeout).
+				Description("The visibility timeout to request for consumed messages, and the value in-flight messages are refreshed with. Zero uses whatever the queue itself is configured with. SQS counts this in whole seconds, so anything finer is rounded down.").
+				Version("1.21.0").
+				Default("0s").
 				Advanced(),
 			service.NewIntField(sqsiFieldMaxNumberOfMessages).
 				Description("The maximum number of messages to return on one poll. Valid values: 1 to 10.").
@@ -196,7 +206,7 @@ func (a *awsSQSReader) Connect(ctx context.Context) error {
 
 	ift := &sqsInFlightTracker{
 		handles:                  map[string]sqsInFlightHandle{},
-		visibilityTimeoutSeconds: a.queueVisibilityTimeout(ctx),
+		visibilityTimeoutSeconds: a.visibilityTimeoutSeconds(ctx),
 	}
 
 	var wg sync.WaitGroup
@@ -214,10 +224,18 @@ func (a *awsSQSReader) Connect(ctx context.Context) error {
 // sqs:GetQueueAttributes is not granted.
 const defaultVisibilityTimeoutSeconds = 30
 
-// queueVisibilityTimeout reads the queue's configured VisibilityTimeout, which
-// is the value in-flight messages should be refreshed with. It is a queue level
-// attribute, so ReceiveMessage never reports it per message.
-func (a *awsSQSReader) queueVisibilityTimeout(ctx context.Context) int {
+// visibilityTimeoutSeconds is the value in-flight messages are refreshed with.
+// Without a configured timeout the queue has to be asked for its own, since
+// VisibilityTimeout is a queue level attribute ReceiveMessage never reports.
+func (a *awsSQSReader) visibilityTimeoutSeconds(ctx context.Context) int {
+	if a.conf.VisibilityTimeout > 0 {
+		return int(a.conf.VisibilityTimeout.Seconds())
+	}
+	if !a.conf.UpdateVisibility {
+		// Nothing refreshes, so nothing needs the queue's value.
+		return 0
+	}
+
 	res, err := a.sqs.GetQueueAttributes(ctx, &sqs.GetQueueAttributesInput{
 		QueueUrl:       aws.String(a.conf.URL),
 		AttributeNames: []types.QueueAttributeName{types.QueueAttributeNameVisibilityTimeout},
@@ -402,6 +420,7 @@ func (a *awsSQSReader) readLoop(wg *sync.WaitGroup, inFlightTracker *sqsInFlight
 			QueueUrl:              aws.String(a.conf.URL),
 			MaxNumberOfMessages:   int32(a.conf.MaxNumberOfMessages),
 			WaitTimeSeconds:       int32(a.conf.WaitTimeSeconds),
+			VisibilityTimeout:     int32(a.conf.VisibilityTimeout.Seconds()),
 			AttributeNames:        []types.QueueAttributeName{types.QueueAttributeNameAll},
 			MessageAttributeNames: []string{"All"},
 		})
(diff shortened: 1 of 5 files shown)


ReguiguiMohamed:
hey @gregfurman hope you're having a great day 
i'll be here waiting for any remark or fixes to update the PR if needed!

internal/impl/aws/input_sqs.go
```diff
@@ -208,15 +210,40 @@ func (a *awsSQSReader) Connect(ctx context.Context) error {
 	return nil
 }
 
+// Used when the queue's own VisibilityTimeout cannot be read, for example when
+// sqs:GetQueueAttributes is not granted.
+const defaultVisibilityTimeoutSeconds = 30
```
gregfurman, line 225: Wdyt about including this is a configurable parameter in the bento config? e.g

```yml
aws_sqs:
  url: foo
  reset_visibility: true
  update_visibility: true
  visibility_timeout: 30s # value of 0 means use sqs queue default
```

That way, even if the actual queue's visibility timeout is `30s`, we could optionally increase this value. Otherwise, if set to `0`, then we could retrieve the value using your new function. Would require that the `ReceiveMessage` request include the new parameter:

https://github.com/warpstreamlabs/bento/blob/c4b89ea14a7423a7fb17f533fe0fa7a5f1388b38/internal/impl/aws/input_sqs.go#L382-L388

internal/impl/aws/input_sqs.go
```diff
@@ -208,15 +210,40 @@ func (a *awsSQSReader) Connect(ctx context.Context) error {
...
+
+// queueVisibilityTimeout reads the queue's configured VisibilityTimeout, which
+// is the value in-flight messages should be refreshed with. It is a queue level
+// attribute, so ReceiveMessage never reports it per message.
+func (a *awsSQSReader) queueVisibilityTimeout(ctx context.Context) int {
```
gregfurman: Can we make the `VisibilityTimeout` an attribute of the `sqsiConfig`?

internal/impl/aws/input_sqs.go
```diff
@@ -194,7 +195,8 @@ func (a *awsSQSReader) Connect(ctx context.Context) error {
 	}
 
 	ift := &sqsInFlightTracker{
-		handles: map[string]sqsInFlightHandle{},
+		handles:                  map[string]sqsInFlightHandle{},
+		visibilityTimeoutSeconds: a.queueVisibilityTimeout(ctx),
```
gregfurman: If `update_visibility` is `false`, then this request should probably be a no-op?

gregfurman (changes requested):
Thanks for the PR! Have left some comments. Would be curious to hear your thoughts on allowing the visibility timeout value to be a configurable field of the input.

internal/impl/aws/input_sqs.go
ReguiguiMohamed: yes , nothing refreshes when `update_visibility` is false, so that was a wasted request on every connect.. the test asserts `GetQueueAttributes` is not called

internal/impl/aws/input_sqs.go
ReguiguiMohamed: Yes and done , its  `sqsiConfig.VisibilityTimeout`  now , and renamed the helper to visibilityTimeoutSeconds since it only asks the queue when the config leaves it unset

internal/impl/aws/input_sqs.go
ReguiguiMohamed: actually that makes sense , i added `visibility_timeout ` as a duration field with default 0s, and it now goes into `ReceiveMessage` so now when resolving the order is config, then the queue, then the 30s fallback, which I kept out of the receive request so a denied GetQueueAttributes cannot shorten a longer queue timeout. Rounds to whole seconds.

internal/impl/aws/input_sqs.go
```diff
@@ -110,6 +115,11 @@ You can access these metadata fields using
 				Version("1.6.0").
 				Default(true).
 				Advanced(),
+			service.NewDurationField(sqsiFieldVisibilityTimeout).
+				Description("The visibility timeout to request for consumed messages, and the value in-flight messages are refreshed with. Zero uses whatever the queue itself is configured with. SQS counts this in whole seconds, so anything finer is rounded down.").
+				Version("1.21.0").
+				Default("0s").
```
gregfurman: Perhaps we leave this at `30s` for now, since changing the default will alter how existing pipelines behave? Like, if I know that Bento sets this to `30s` and my queue has a `VisibilityTimeout` of 12 hours, then this'll mean things only get processed again after 12 hours.

internal/impl/aws/input_sqs.go
```diff
@@ -110,6 +115,11 @@ You can access these metadata fields using
 				Version("1.6.0").
 				Default(true).
 				Advanced(),
+			service.NewDurationField(sqsiFieldVisibilityTimeout).
+				Description("The visibility timeout to request for consumed messages, and the value in-flight messages are refreshed with. Zero uses whatever the queue itself is configured with. SQS counts this in whole seconds, so anything finer is rounded down.").
```
gregfurman:
```suggestion
				Description("Visibility timeout (truncated to whole seconds) requested when retrieving messages, and used when refreshing in-flight messages. A value of `0` defers to the SQS queue's configured timeout.").
```

internal/impl/aws/input_sqs.go
ReguiguiMohamed: yes fair , it should default to 30s..
but my question is here is if we have `update_visibility: false` . those pipelines get the queue's full 12h today because bento never touches visibility for them, but the 30s default would still go out on `ReceiveMessage `and drop them to 30s , so for now the field only applies when `update_visibility `is on... should it apply always?

internal/impl/aws/input_sqs.go
```diff
@@ -110,6 +115,11 @@ You can access these metadata fields using
 				Version("1.6.0").
 				Default(true).
 				Advanced(),
+			service.NewDurationField(sqsiFieldVisibilityTimeout).
+				Description("Visibility timeout (truncated to whole seconds) requested when retrieving messages, and used when refreshing in-flight messages. A value of `0` defers to the SQS queue's configured timeout, as does disabling `update_visibility`.").
+				Version("1.21.0").
```
gregfurman:
```suggestion
				Version("1.22.0").
```

gregfurman (approved):
LGTM! Once merge conflicts are resolved and the docs updated, this looks good to merge 👍 

Thanks for the neat contribution here @ReguiguiMohamed !

ReguiguiMohamed:
Wonderful, thank you!

PR #988, opened 2026-08-21
feat(parquet): add DATE and TIMESTAMP_MILLIS logical type support

triddell:
## Summary
- Adds `DATE` and `TIMESTAMP_MILLIS` as selectable column types in the `parquet_encode` processor's schema config, mapping to parquet-go's `date`/`timestamp` logical type struct tags.
- Fixes a panic when a `DATE`/`TIMESTAMP_MILLIS` field is marked `optional` with `optionalAsPtrs` enabled: parquet-go's `date`/`timestamp` struct tags only accept a pointer when the pointee is `time.Time`, so pointer-wrapping a plain `int32`/`int64` for these types triggered `throwInvalidTag`. Optionality for these two types is now carried by the `optional` tag component alone rather than by wrapping the Go type in a pointer.

## Test plan
- [x] `go test ./internal/impl/parquet/...` passes
- [x] Added `TestOptionalDateTimestampDoesNotPanic` reproducing the pointer-wrap panic
- [x] Extended existing schema-generation table test with `DATE`/`TIMESTAMP_MILLIS` cases

### DIFF
--- a/internal/impl/parquet/schema.go
+++ b/internal/impl/parquet/schema.go
@@ -304,12 +304,33 @@ func wrapType(
 				return nil, fmt.Errorf("getting optional flag: %w", err)
 			}
 			if optional {
+				// parquet-go's "date"/"timestamp" struct tags only accept a pointer for
+				// *time.Time, not *int32/*int64 (see parquet-go schema.go, case "date"/
+				// "timestamp": a Ptr whose Elem isn't time.Time panics via throwInvalidTag).
+				// Our DATE/TIMESTAMP_MILLIS tokens keep the underlying value as a plain
+				// int32/int64, so optionality has to be carried by the "optional" tag
+				// component alone (already appended above) rather than by wrapping in a
+				// pointer here.
+				if isDateOrTimestampType(field) {
+					return baseType, nil
+				}
 				return reflect.PointerTo(baseType), nil
 			}
 		}
 	}
 
 	return baseType, nil
+}
+
+func isDateOrTimestampType(field *service.ParsedConfig) bool {
+	if !field.Contains("type") {
+		return false
+	}
+	typeStr, err := field.FieldString("type")
+	if err != nil {
+		return false
+	}
+	return typeStr == "DATE" || typeStr == "TIMESTAMP_MILLIS"
 }
 
 func isDeltaLengthByteArrayEncodable(typeStr string) bool {
--- a/internal/impl/parquet/schema.go
+++ b/internal/impl/parquet/schema.go
@@ -84,6 +84,14 @@ func generateStructTypeFromFields(
 
 				components = append(components, fmt.Sprintf("decimal(%d:%d)", scale, precision))
 			}
+
+			if typeStr == "DATE" {
+				components = append(components, "date")
+			}
+
+			if typeStr == "TIMESTAMP_MILLIS" {
+				components = append(components, "timestamp")
+			}
 		}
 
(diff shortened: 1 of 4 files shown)


triddell:
@gregfurman @jem-davies Can you take a look at this PR? It's pretty basic and I have another PR that I'd like to submit but it relies on this being accepted first. Thanks.

gregfurman:
@triddell hey! Sorry I missed this. Will give a priority review tomorrow.

internal/impl/parquet/schema.go
```diff
@@ -292,6 +304,16 @@ func wrapType(
...
 			}
 			if optional {
+				// parquet-go's "date"/"timestamp" struct tags only accept a pointer for
+				// *time.Time, not *int32/*int64 (see parquet-go schema.go, case "date"/
+				// "timestamp": a Ptr whose Elem isn't time.Time panics via throwInvalidTag).
+				// Our DATE/TIMESTAMP_MILLIS tokens keep the underlying value as a plain
+				// int32/int64, so optionality has to be carried by the "optional" tag
+				// component alone (already appended above) rather than by wrapping in a
+				// pointer here.
+				if isDateOrTimestampType(field) {
```
gregfurman: I'm not sure this is correct. It basically means that we can never have an optional `DATE` or `TIMESTAMP_MILLIS` in our schema -- where instead of a `null` we'll end up with a `0`.

Like, what is the scenario where we expect this to actually fail?

gregfurman (changes requested):
Unsure about the latest changes on cfc18f7 -- why would we want to prevent optional `DATE` and `TIMESTAMP_MILLIS`?

gregfurman:
@triddell I've opened a PR against your branch that adds some tests + removes the optional stuff. 

FYI, I'm testing this with the following config and all is looking good:

```yml
input:
  generate:
    count: 1
    mapping: |
      root.date_val = timestamp_unix()
      root.time_val = timestamp_unix_milli()

pipeline:
  processors:
    - parquet_encode:
        schema:
          - name: date_val
            type: DATE
          - name: time_val
            type: TIMESTAMP_MILLIS
    - catch:
      - log:
          message: "Processor ${!error_source_label()} failed: ${!error()}"
    - parquet_decode: {}
    - mapping: |
        root.date = this.date_val.ts_format("Mon Jan 2 15:04:05 -0700 MST 2006", "UTC")
        root.time = this.time_val

```

triddell:
@gregfurman thanks for catching this — you're right, and I've merged your fix in (https://github.com/triddell/bento/pull/4). Good catch on the null-vs-zero issue: skipping the pointer-wrap meant a missing optional `DATE`/`TIMESTAMP_MILLIS` would silently encode as epoch-zero instead of `NULL`, which defeats the point of marking it optional at all.

For context on why I added that skip originally: `parquet-go`'s `"date"`/`"timestamp"` tag case only accepts a `*time.Time` pointer, not `*int32`/`*int64`, and I hit a real `throwInvalidTag` panic against a production batch. I can't reproduce that panic with your revert against the current `parquet-go` (v0.29.0) — your round-trip test (encode → real bytes → decode) passes clean, including a genuinely-missing optional field correctly coming back as `null` rather than `0`. Not 100% sure what changed to make it stop reproducing, but your version is correct either way and I trust the test over my memory of the original crash.

Ready for another look whenever you get a chance.

gregfurman (approved):
LGTM!

PR #989, opened 2026-08-21
feat(statsd metrics): expose go-statsd send queue/buffer pool/send loop tuning

triddell:
## Summary
The `metrics.statsd` exporter only ever passes `FlushInterval`, `Logger`, and `TagStyle` to the underlying `go-statsd` client, leaving its send queue hardcoded at the library's default of 10 packets. Under sustained high-throughput, CPU-constrained workloads this queue can overflow and metrics are silently dropped (`[STATSD] N packets lost (overflow)`), even though `go-statsd` already exposes `SendQueueCapacity`, `BufPoolCapacity`, and `SendLoopCount` as functional options to fix exactly this.

- Adds three new `.Advanced()`, optional config fields: `send_queue_capacity`, `buf_pool_capacity`, `send_loop_count`, wired through to the corresponding `go-statsd` options.
- All three default to `go-statsd`'s own defaults (10 / 20 / 1), so behavior is unchanged unless explicitly configured.

## Test plan
- [x] Added `metrics_statsd_test.go` (package previously had no tests): field defaults, field overrides, and an end-to-end check (via reflection into the vendored client's transport, since `go-statsd` doesn't expose these values post-construction) that the options actually reach `statsd.NewClient`.
- [x] `go test ./internal/impl/statsd/...`
- [x] `make docs` — regenerated `statsd.md`

### DIFF
--- a/internal/impl/statsd/metrics_statsd_test.go
+++ b/internal/impl/statsd/metrics_statsd_test.go
@@ -1,0 +1,88 @@
+package statsd
+
+import (
+	"context"
+	"fmt"
+	"reflect"
+	"testing"
+
+	statsd "github.com/smira/go-statsd"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/warpstreamlabs/bento/public/service"
+)
+
+func parseStatsdYAML(t testing.TB, conf string, args ...any) *service.ParsedConfig {
+	t.Helper()
+
+	pConf, err := statsdSpec().ParseYAML(fmt.Sprintf(conf, args...), nil)
+	require.NoError(t, err)
+	return pConf
+}
+
+func TestStatsdTuningFieldDefaults(t *testing.T) {
+	pConf := parseStatsdYAML(t, `address: localhost:8125`)
+
+	sendQueueCapacity, err := pConf.FieldInt(smFieldSendQueueCapacity)
+	require.NoError(t, err)
+	assert.Equal(t, statsd.DefaultSendQueueCapacity, sendQueueCapacity)
+
+	bufPoolCapacity, err := pConf.FieldInt(smFieldBufPoolCapacity)
+	require.NoError(t, err)
+	assert.Equal(t, statsd.DefaultBufPoolCapacity, bufPoolCapacity)
+
+	sendLoopCount, err := pConf.FieldInt(smFieldSendLoopCount)
+	require.NoError(t, err)
+	assert.Equal(t, statsd.DefaultSendLoopCount, sendLoopCount)
+}
+
+func TestStatsdTuningFieldOverrides(t *testing.T) {
+	pConf := parseStatsdYAML(t, `
+address: localhost:8125
+send_queue_capacity: 200
+buf_pool_capacity: 40
+send_loop_count: 4
+`)
+
+	sendQueueCapacity, err := pConf.FieldInt(smFieldSendQueueCapacity)
+	require.NoError(t, err)
+	assert.Equal(t, 200, sendQueueCapacity)
+
+	bufPoolCapacity, err := pConf.FieldInt(smFieldBufPoolCapacity)
+	require.NoError(t, err)
+	assert.Equal(t, 40, bufPoolCapacity)
+
+	sendLoopCount, err := pConf.FieldInt(smFieldSendLoopCount)
+	require.NoError(t, err)
+	assert.Equal(t, 4, sendLoopCount)
+}
+
+// channelCapacity reads the capacity of an unexported chan field on the vendored
+// statsd.Client via reflection. go-statsd doesn't expose SendQueueCapacity/BufPoolCapacity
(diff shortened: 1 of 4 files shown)


internal/impl/statsd/metrics_statsd_test.go
```diff
@@ -0,0 +1,88 @@
...
+// channelCapacity reads the capacity of an unexported chan field on the vendored
+// statsd.Client via reflection. go-statsd doesn't expose SendQueueCapacity/BufPoolCapacity
+// once a Client is built, so this is the only way to assert the options passed to
+// NewClient actually reached the transport, rather than just asserting config parsing.
+func channelCapacity(t testing.TB, c *statsd.Client, fieldName string) int {
+	t.Helper()
+
+	trans := reflect.ValueOf(c).Elem().FieldByName("trans").Elem()
+	return trans.FieldByName(fieldName).Cap()
+}
```
gregfurman: Can we not create a simple UDP listener, set its address to the `address` field of the statsd config, and inspect the output?

The `smira/go-statsd` has a neat example of a UDP listener test at https://github.com/smira/go-statsd/blob/cee3b7c031bcb785a7217e9f98fc810aee82ec9d/client_test.go#L39-L64 (see snippet below):

```golang
func setupListener(t *testing.T) (*net.UDPConn, chan []byte) {
	inSocket, err := net.ListenUDP("udp4", &net.UDPAddr{
		IP: net.IPv4(127, 0, 0, 1),
	})
	if err != nil {
		t.Error(err)
	}
// ...
}

```

Then we could wire it up similar to how we unit test the prometheus component

https://github.com/warpstreamlabs/bento/blob/80189d9dbbb19d53b7819fd6260cb316dbdf540f/internal/impl/prometheus/metrics_prometheus_test.go#L36-L59

gregfurman (commented):
Thanks for the contribution! IMO the reflection approach in the tests should be replaced with a simple mock server, with incoming data being inspected to verify all works. Have left a comment on how I'd approach this.

Otherwise, if you could just include a section in the changelog for this that'd be great 👍

triddell:
Thanks for the review, Greg! Both addressed:

* Swapped the reflection-based check for a real UDP listener (following the pattern from go-statsd's client_test.go you linked, wired up the same way as the prometheus test). The test now spins up a net.ListenUDP socket, points the exporter's address at it, fires a counter increment through newStatsdFromParsed, and asserts on the actual packet bytes received (test.counter:1|c) — so it's exercising the real send path end-to-end rather than poking at unexported fields.
* Added a changelog entry under ## Unreleased in CHANGELOG.md.

Let me know if you'd like anything else adjusted.

triddell:
Also rebased onto the latest main to pick up the 1.21.1 changelog changes and resolve the resulting conflict there. One thing I can't self-serve: golangci-lint is failing on a transient module-proxy error unrelated to this diff (could not import ... aws-sdk-go-v2/service/sns ... stream error ... INTERNAL_ERROR fetching from proxy.golang.org) — none of the changed files touch AWS imports. Would appreciate a re-run when someone has a chance.

internal/impl/statsd/metrics_statsd_test.go
```diff
@@ -0,0 +1,115 @@
...
+
+func TestNewStatsdFromParsedWiresTuningOptions(t *testing.T) {
+	conn, out := setupListener(t)
+
+	pConf := parseStatsdYAML(t, `
+address: %v
+flush_period: 1ms
+send_queue_capacity: 123
+buf_pool_capacity: 45
```
gregfurman, line 97: Would we expect any of these new fields being reflected in the output?

triddell:
Good question — no, none of these three fields would change the output content. `send_queue_capacity`, `buf_pool_capacity`, and `send_loop_count` only affect internal buffering/concurrency inside `go-statsd` (queue depth before packets get dropped, buffer pool size, and number of draining goroutines) — the wire format of a given metric call is identical regardless of what they're set to. Checked the `go-statsd` source (`client.go`/`buffers.go`/`loops.go`) to confirm.

So this test isn't proving those specific values reached `NewClient` — it's really just a smoke test that the client still constructs and sends correctly when these fields are set to non-default values, on top of `TestStatsdTuningFieldDefaults`/`TestStatsdTuningFieldOverrides` which cover the config-parsing side.

Proving the values actually reach the transport without reflection would mean forcing a real queue overflow (e.g. tiny `send_queue_capacity` + enough load to overflow it, asserting a `[STATSD] N packets lost` log line, vs. no overflow at a larger capacity under identical load) — doable, but it'd hinge on relative timing between the test goroutine and the UDP send loop, which is a good way to get an intermittently flaky test in CI for not much benefit, since `send_loop_count` in particular has no log-visible effect at all.

Given that, I'd lean toward keeping the current scope (config parsing + smoke-tested wiring) unless you feel strongly that the overflow-based check is worth the flakiness risk — happy to add it if so.

gregfurman (approved):
Thanks for adding that test and addressing my feedback!

LGTM

Issue #353, opened 2025-06-01
Contains seems to return true even if the field isn't set


jem-davies:
https://github.com/warpstreamlabs/bento/pull/346#discussion_r2119397025

PR #349, opened 2025-05-29
feat(input_csv): add expected_headers and expected_number_of_fields params

etolbakov:
## What's changed and what's your intention?

As per the title

## Related Issue link
https://github.com/warpstreamlabs/bento/issues/281

PR #991, opened 2026-08-22
Add mqtt_v5 input and output

### LINKED ISSUE #985: MQTT 5 support: new `mqtt_v5` input and output
## The gap

Bento's `mqtt` input and output speak MQTT 3.1.1 only. `internal/impl/mqtt/client.go` imports `github.com/eclipse/paho.mqtt.golang` v1.5.1, and the config surface reflects that: `clean_session` rather than `clean_start` plus a session expiry, and no way to reach any of the properties MQTT 5 added.

The practical consequences for a Bento pipeline:

- **No user properties**, so a message carries no headers. This is the big one — a producer's own metadata cannot reach the pipeline, and metadata set in a pipeline cannot be published. `integration_test.go` records this: `integration.StreamTestMetadata()` sits commented out with a `// TODO`, because 3.1.1 has nothing for it to test.
- **No content type, response topic, correlation data, message expiry or payload format indicator** — the properties that make an MQTT message self-describing.
- **No session expiry interval**, so a durable subscription's lifetime cannot be stated; `clean_session: false` is all there is.
- **No reason codes or reason strings** on CONNACK, SUBACK, PUBACK or DISCONNECT, so a refusal arrives as a generic error rather than as the reason the broker gave.
- **Shared subscriptions** (`$share/<group>/<filter>`), which MQTT 5 standardises, cannot be relied on over a 3.1.1 connection.

MQTT 5.0 has been the current version of the protocol since 2019 and is what brokers ship today. Upstream `paho.mqtt.golang` has had MQTT 5 support as a long-open request rather than a plan, so this is not something a dependency bump will resolve.

## Proposal

Add an `mqtt_v5` input and an `mqtt_v5` output that speak MQTT 5 **only** — no fallback, no version negotiation, and no change to the existing `mqtt` components.

**Shape.** New files in the existing `internal/impl/mqtt` package: `client_v5.go`, `input_v5.go`, `output_v5.go`, `integration_v5_test.go`. `public/components/mqtt/package.go` already imports that package, so `public/components/all` needs no change. This follows the precedent already in the tree — `internal/impl/elasticsearch` keeps `output.go` beside `output_v2.go`, and `integration_test.go` beside `integration_v2_test.go`, with the component named `elasticsearch_v2`.

**Config.** Keep the existing field names wherever the concept survives (`urls`, `client_id`, `dynamic_client_id_suffix`, `connect_timeout`, `keepalive`, `user`, `password`, `tls`, `will`), so moving a config across changes as little as possible. Add only what MQTT 5 actually introduces: `clean_start`, `session_expiry_interval`, `receive_maximum`, `maximum_packet_size`, `will.delay_interval`; on the input, `no_local`, `retain_as_published` and `retain_handling`; on the output, `content_type`, `response_topic`, `correlation_data`, `message_expiry_interval` and `payload_format_indicator`.

**Metadata.** The input sets the same `mqtt_topic` / `mqtt_qos` / `mqtt_retained` / `mqtt_duplicate` / `mqtt_message_id` keys as the 3.1.1 input so pipelines port across unchanged, adds the MQTT 5 properties under `mqtt_*` names, and maps user properties to metadata under their own keys. The output turns metadata into user properties through a `service.NewMetadataExcludeFilterField`, the same way `internal/impl/amqp09/output.go` does. `StreamTestMetadata()` can then be turned on rather than left as a TODO.

**Acknowledgement.** Manual acknowledgement, so the broker is not told a message was handled until the pipeline's own ack fires, alongside the standard `auto_replay_nacks` toggle.

## The dependency

`github.com/eclipse/paho.golang` (`autopaho` + `paho`), currently v0.23.0. It is the Eclipse Paho project's Go MQTT 5 client — the same project as the 3.1.1 client already in the tree, under the same dual licence, EPL-2.0 **or** EDL-1.0, so it raises no licensing question that `paho.mqtt.golang` has not already raised.

The obvious objection is that it is pre-1.0. For what it is worth, `go.mod` here already requires around forty direct v0.x modules — `apache/pulsar-client-go` v0.17.0, `getsentry/sentry-go` v0.47.0, `parquet-go` v0.29.0, `sourcegraph/conc` v0.3.0 and `dop251/goja` among them — so a pinned v0.x does not look out of place. There is no second maintained MQTT 5 client in Go, so if this is a blocker it is worth saying now.

## Not in scope

Topic aliases, request/response helpers, enhanced authentication (AUTH packets), per-filter subscription options, and any MQTT 5 to 3.1.1 fallback. Each is a reasonable follow-up; including them would make the first review much harder.

## Questions before any code

1. **Separate `mqtt_v5` components, or a version flag on the existing `mqtt`?** I have proposed separate, on the `elasticsearch_v2` precedent: the config surfaces genuinely differ, the client libraries differ, and configuration is validated per component, so a flag would leave roughly half the fields invalid in each mode with nothing able to say so. Happy to build it either way — a flag changes the config spec and registration but not the client code. If you would rather it were named `mqtt_5`, or something else entirely, that is a rename.
2. **Is `eclipse/paho.golang` at v0.x acceptable as a dependency?**
3. **Which broker image would you like the integration test to run against?** I have assumed `eclipse-mosquitto:2`, since it speaks MQTT 5 and needs only a small config to enable an anonymous listener.

I am happy to do the work — roughly 1,200 lines including tests — and would rather agree the shape here first than send a PR that has to be restructured. Thanks for taking a look.

ifnesi:
Closes #985.

Bento's `mqtt` components speak MQTT 3.1.1 only, so nothing MQTT 5 added is reachable: no user properties, no shared subscriptions, no session expiry, and no reason codes. This adds an `mqtt_v5` input and output speaking MQTT 5 only — no fallback, no negotiation, and no change to the existing components.

## The shortest argument for it

`internal/impl/mqtt/integration_test.go` carries this line, commented out with a TODO, because MQTT 3.1.1 has nowhere to put metadata:

```go
// integration.StreamTestMetadata(), TODO
```

In `integration_v5_test.go` it is switched on and passes, along with `StreamTestMetadataFilter()`. That is the whole feature in one line.

## Answering the questions from #985

Thanks for the quick answers — all three shaped this directly.

- **Separate components rather than a version flag**, on the `elasticsearch_v2` precedent. The config surfaces genuinely differ and Bento validates per component, so a flag would leave half the fields invalid in each mode with nothing able to say so.
- **`github.com/eclipse/paho.golang` v0.23.0** — same Eclipse project as the 3.1.1 client already in the tree, same dual licence (EPL-2.0 **or** EDL-1.0, which is BSD-3-Clause under another name), so no new licence question.
- **Experimental status**, following the note that a component resting on a v0 dependency is sometimes marked so. These specs simply do not call `.Stable()`, which is already the default for a plugin spec.
- **`eclipse-mosquitto:2`** for the integration tests, via dockertest.

## What the components do

**Input** — `topics` (a shared subscription is just a `$share/<group>/<filter>` filter, so it needs no field of its own), `qos`, per-connection subscription options, and manual acknowledgement so the server is not told a message was handled until the pipeline finished with it.

MQTT 5 properties become metadata under `mqtt_*`. **User properties become metadata under their own keys**, written before the `mqtt_*` fields so a publisher cannot forge `mqtt_topic` by sending a user property of that name.

**Output** — the 3.1.1 fields plus the MQTT 5 publish properties (`content_type`, `response_topic`, `correlation_data`, `message_expiry_interval`, `payload_format_indicator`), each interpolated, and message metadata sent as user properties through the standard exclude filter.

## Reason codes, which is most of why this exists

A refusal in MQTT 5 says why, and the client library's error text does not carry it — `Subscribe` reports "at least one requested subscription failed" and drops the code. These components read the code off the packet instead, and decode CONNACK, SUBACK, PUBACK/PUBREC and DISCONNECT reasons.

That matters because **a refused subscription does not close the connection**: without handling, the input sits connected, healthy, and receiving nothing. `on_subscribe_refused` chooses between retrying, failing the input, or continuing with the filters that were granted; `on_connect_refused` does the same for CONNACK.

There is deliberately **no** setting for a refused publication, because Bento already has a better one: the error carries the reason code, a `fallback` output copies it into `fallback_error`, and a `switch` routes on it — a different destination per reason code rather than one blanket behaviour. That pattern is in the output's docs.

One trap worth flagging for review: **at QoS 2 the library returns a refused publication with a nil error** — `publishQoS12` hands back a PUBREC it has itself logged as "must have errored". The output checks the reason code rather than trusting the error return, so a rejected message is not reported to the pipeline as delivered. There is a test for it.

## Two decisions a reviewer should see were made on purpose

**The output holds back the `mqtt_` prefix by default.** That namespace is what the input writes to describe a delivery — topic, QoS, delivery flags. Forwarding it means every server-to-server hop adds a handful of properties describing the previous hop, which a server that limits properties will refuse. The setting is an ordinary exclude filter, so `exclude_prefixes: []` sends them and any other list holds back whatever a pipeline names.

**An interpolated optional property that resolves to nothing leaves the property unset.** Reading an absent metadata field yields the text `null`, and most messages carry only some of these properties, so treating that as a value failed every message whose source had no expiry. This matches what `kafka_franz` already does for a record key. A value that is present but *malformed* is still an error.

## Tests

- **Unit** — config parsing, bounds, CONNACK/DISCONNECT decoding, metadata mapping.
- **A stub MQTT 5 server, no container** — every refusal path, including the QoS 2 case no real broker will produce on demand. mosquitto answers `SUBACK 0` for a topic its own ACL denies, so an ACL fixture cannot test this.
- **Integration** (`eclipse-mosquitto:2`, dockertest) — the stock stream suite, plus the cases it cannot express: a durable session receiving what was published while the input was away, two inputs sharing one group splitting a stream without duplicates, a rejected message redelivered and an acknowledged one not, and the properties surviving output → input.

`make fmt`, `make lint` and `make test` are clean, and `make docs` regenerates the two pages with no diff.

Run the integration tests selectively as the Makefile advises:

```sh
go test -run '^TestIntegrationMQTTv5$' ./internal/impl/mqtt/...
```

## Deliberately out of scope

Topic aliases, request/response helpers, enhanced authentication, per-filter subscription options, any MQTT 5 → 3.1.1 fallback, and any change to the existing `mqtt` components.

Written with [Claude Code](https://claude.com/claude-code) (Opus 5), matching the `Co-authored-by` trailer on the commits.


ifnesi:
Thanks for running the checks, @jem-davies — they found two real problems and both are now fixed.

**The failures were mine, and in the documentation rather than in the components.** `make docs` lints every fenced block tagged `` ```yaml `` as a whole configuration and skips every one tagged `` ```yml ``, and two of my blocks wore the linted tag without being configurations:

- the input's dead-letter example left `- kafka_franz: { }` as a placeholder, so the lint asked for the `topic` that output requires. It now names a broker and a topic, so it stays linted and a reader can paste it.
- the output's `metadata` description shows `exclude_prefixes` twice to contrast the two settings side by side, which is a duplicate mapping key. Since the contrast is the point of the block, it is tagged `` ```yml `` now, as the other partial snippets in the tree are.

The other two `` ```yaml `` blocks in these components were already valid configurations and are unchanged.

I also rebased onto `main`: the branch had drifted and `CHANGELOG.md` conflicted, so the `mqtt_v5` entry now sits above the 1.21.1 section.

Verified before pushing — `make test`, `TestIntegrationMQTTv5` and every subtest, `golangci-lint run` and `go vet` clean, and `make docs` leaving a clean tree. I also ran the same `bento lint` binary against the two documents as they stood before the fix and got the three CI errors back exactly, then against the fixed pair and got nothing, so the check is capable of failing.

One note on the earlier run that is not about this branch: the `golangci-lint` job failed at `golangci-lint config verify`, fetching its JSON schema from `golangci-lint.run` over the network. That step passes locally, so it looks like a transient timeout rather than anything in the config.

The new run is sitting at `action_required` — would you mind kicking it off when you have a moment? Thanks again for the review time.

ifnesi:
Hi @jem-davies @gregfurman — nudging this one gently. All four checks pass. The only conflict is the changelog's `Unreleased` entry; I've left it rather than re-resolving it each time something lands upstream, and I'll clear it as soon as you're ready to review.

Happy to change anything you'd like.

internal/impl/mqtt/client_v5.go
```diff
@@ -0,0 +1,687 @@
...
+		service.NewDurationField(mv5FieldSessionExpiryInterval).
+			Description("How long the server keeps this client's session after the connection closes. Zero means the session ends the moment the connection does. Rounded down to whole seconds, which is the resolution MQTT 5 carries it at.").
+			Default("0s").
+			Examples("1h", "24h"),
+		service.NewIntField(mv5FieldReceiveMaximum).
+			Description("The number of QoS 1 and QoS 2 messages this client is willing to have in flight at once — in effect, how far ahead of the pipeline the server is allowed to read. Left unset, the server chooses.").
+			Optional().
+			Advanced(),
+		service.NewIntField(mv5FieldMaximumPacketSize).
+			Description("The largest packet, in bytes, this client is willing to accept. Left unset, no limit is sent and the server may deliver a packet of any size it supports.").
```
jem-davies, line 106: Confusing as written when this appears on the output component because as it is an output it shouldn't be receiving any messages.

EDIT - apparently it can receive the nack/ack message which this field will apply - but currently as written it seems confusing.

internal/impl/mqtt/input_v5.go
```diff
@@ -0,0 +1,643 @@
...
+func mv5SubackReason(code byte) string {
+	if name, ok := mv5SubackReasons[code]; ok {
+		return name
+	}
+	return "unrecognised reason code"
+}
+
+func inputConfigSpecV5() *service.ConfigSpec {
+	// Deliberately not marked Stable: components resting on a v0 dependency are
+	// documented as experimental, which is what a plugin spec defaults to.
```
jem-davies, line 55: nit: Could add this to the actual documentation, like we have for the [parquet input](https://warpstreamlabs.github.io/bento/docs/components/inputs/parquet/#:~:text=This%20input%20uses%20https%3A//github.com/parquet%2Dgo/parquet%2Dgo%2C%20which%20is%20itself%20experimental.)

internal/impl/mqtt/output_v5.go
```diff
@@ -0,0 +1,457 @@
...
+
+func mv5PubackReason(code byte) string {
+	if name, ok := mv5PubackReasons[code]; ok {
+		return name
+	}
+	return "unrecognised reason code"
+}
+
+func outputConfigSpecV5() *service.ConfigSpec {
+	// Deliberately not marked Stable: components resting on a v0 dependency are
```
jem-davies, line 53: nit: Could add this to the actual documentation, like we have for the [parquet input](https://warpstreamlabs.github.io/bento/docs/components/inputs/parquet/#:~:text=This%20input%20uses%20https%3A//github.com/parquet%2Dgo/parquet%2Dgo%2C%20which%20is%20itself%20experimental.).

internal/impl/mqtt/output_v5.go
```diff
@@ -0,0 +1,457 @@
...
+	retained := m.retained
+	if m.retainedInterp != nil {
+		retainedStr, parseErr := m.retainedInterp.TryString(msg)
+		switch {
+		case parseErr != nil:
+			m.log.Errorf("Retained interpolation error: %v", parseErr)
+		case propertyUnset(retainedStr):
+			// The field it reads is absent, so the configured retained value
+			// stands. Reporting that as a parse failure described a malformed
+			// value when nothing was malformed, and did so once per message.
```
jem-davies:
```suggestion
			// stands. 
```

internal/impl/mqtt/output_v5.go
```diff
@@ -0,0 +1,457 @@
...
+// nothing worth sending.
+//
+// A metadata key that does not exist interpolates to the string "null", not to
+// an empty one, and that is the ordinary case rather than a mistake: bridging
+// one server to another means naming each property to carry, and most messages
+// carry only some of them. Treating it as a value made a bridge fail every
+// message whose source had no message expiry — permanently, because the output
+// retries and the input then holds the acknowledgement behind it. A value that
+// is present but malformed is still an error, because that is a configuration
+// mistake worth stopping for.
```
jem-davies:
```suggestion
```

internal/impl/mqtt/client_v5.go
```diff
@@ -0,0 +1,687 @@
...
+			c.down = make(chan struct{})
+		default:
+			// Still open, which means this is the first connection and nothing
+			// has replaced the channel built with the connection. A reader can
+			// already be holding it, because autopaho releases AwaitConnection
+			// before calling this — so it has to stay, or it would be orphaned
+			// open and that reader would never learn the link had dropped.
+		}
+		c.mu.Unlock()
+		c.log.Info("Connection established.")
```
jem-davies: Bento will already log `INFO Output type mqtt_v5 is now active` - when the Connect function returns without error.

internal/impl/mqtt/client_v5.go
```diff
@@ -0,0 +1,687 @@
...
+			}
+			return pkt, nil
+		}
+	}
+}
+
+// mv5ConnackReasons names the CONNACK reason codes MQTT 5 defines. The client
+// library's error text does not carry the code, so a refusal logged from the
+// error alone says only that the connection failed, never why.
+var mv5ConnackReasons = map[byte]string{
```
jem-davies, line 408: I see that the library doesn't return anything in X.Properties.ReasonString / ServerReference - and that we are hardcoding them here - makes sense. 

Is it worth raising an issue on the dependency to add these reasons in? 

Not blocking.

internal/impl/mqtt/client_v5.go
```diff
@@ -0,0 +1,687 @@
...
+func mv5DisconnectReason(code byte) string {
+	if name, ok := mv5DisconnectReasons[code]; ok {
+		return name
+	}
+	return "unrecognised reason code"
+}
+
+// errConnectionRefused marks a connection the server answered and turned down,
+// as opposed to one that never got an answer at all. Components test for it
+// with errors.Is to honour on_connect_refused.
```
jem-davies: nit: self-explanatory comment really

internal/impl/mqtt/client_v5.go
```diff
@@ -0,0 +1,687 @@
...
+		service.NewStringField(mv5FieldClientID).
+			Description("An identifier for the client connection. Under MQTT 5 this is the key the server stores a session against, so a client that wants to resume a session must present the same identifier every time it connects.").
+			Default(""),
+		service.NewStringAnnotatedEnumField(mv5FieldDynClientIDSuffix, map[string]string{
+			"nanoid": "append a nanoid of length 21 characters",
+		}).
+			Description("Append a dynamically generated suffix to the specified `client_id` on each run of the pipeline. This can be useful when clustering Bento producers. Note that a generated suffix produces a new client identifier on every run, and therefore a new session: combining it with `clean_start: false` will not resume anything.").
+			Optional().
+			Advanced().
+			LintRule(`root = []`), // Disable linting for now
```
jem-davies:
```suggestion
```

I see that it is also on `mqtt`.

internal/impl/mqtt/client_v5.go
```diff
@@ -0,0 +1,687 @@
...
+		return
+	}
+	if c.backoffMax, err = backoff.FieldDuration(mv5FieldReconnectBackoffMax); err != nil {
+		return
+	}
+	// autopaho.NewExponentialBackoff panics rather than returning an error, so
+	// its constraints are checked here where a bad value is still a
+	// configuration error the user can read.
+	if c.backoffMin <= 0 {
+		err = errors.New("reconnect_backoff.min must be greater than zero")
```
jem-davies, line 293: These are the kind of things that you could check earlier in Bento's runtime lifecycle - by using a LintRule. 

Not blocking but worth pointing out.

internal/impl/mqtt/client_v5.go
```diff
@@ -0,0 +1,687 @@
...
+		c.mu.Unlock()
+		c.log.Info("Connection established.")
+		if onUp != nil {
+			onUp(cm, connack)
+		}
+	}
+	cfg.OnConnectionDown = func() bool {
+		c.markDown()
+		c.log.Warn("Connection lost, reconnecting.")
+		return true
```
jem-davies: I think with this we are handing off the re-connection logic to autopaho - which makes sense to do. 

But the problem I am seeing is that in the event of a disconnect & subsequent reconnect via this library - Bento's observability metrics read wrong - the event of a disconnect & reconnect will read as a message send failure. 

I think that semantically & for the purpose of metrics/logs it would be better if we had reconnect via throwing Bento's service.ErrNotConnected.

internal/impl/mqtt/input_v5.go
```diff
@@ -0,0 +1,643 @@
...
+// is not where the common case shows up. Under the default auto_replay_nacks
+// the pipeline's error never reaches this component at all: Bento re-queues the
+// message and retries it internally for ever, and the acknowledgement function
+// here is only ever called on eventual success. A message that can never
+// succeed therefore holds its acknowledgement — and every acknowledgement
+// behind it — with nothing in this component seeing an error to report. Once
+// the receive window is full the server stops delivering and the pipeline sits
+// healthy and idle. Watching the counters catches that; watching for rejections
+// does not.
+func (m *mqttReaderV5) watchProgress() {
```
jem-davies: Yeah I see the motivation behind doing this.

>  Bento re-queues the message and retries it internally for ever, and the acknowledgement function here is only ever called on eventual success. 

I do think that can surprise people - it seems like Bento doesn't really know what to do with the message so it replays it. 

I think that the idea behind it is that eventually it should be noted by Bento's operator that it isn't delivering messages (via observability) - and ultimately because the message has been replayed - it cannot deliver the message according to the stream config. 

This would mean that there must be some intervention required , i.e. updating the stream config / checking the health of the upstream / downstream services etc. 

But I think that this should be removed because it is too ad-hoc with this particular input - that it should be a separate issue to either fix generally or improve the understanding of this behaviour via documentation.

internal/impl/mqtt/output_v5.go
```diff
@@ -0,0 +1,457 @@
...
+		}
+		return m.publishError(resp, err)
+	}
+
+	// The library reports a refusal as an error at QoS 1, and as a success at
+	// QoS 2: publishQoS12 returns a PUBREC it has itself logged as "must have
+	// errored" with a nil error, and a PUBCOMP carrying a failure code the same
+	// way. Reading the reason code here rather than trusting the error return
+	// is what stops a publication the server rejected being reported to the
+	// pipeline as delivered.
```
jem-davies, line 341: is this a problem with library - should it be reported as an issue on the dependency? 

Or is this expected behaviour with MQTT?

internal/impl/mqtt/output_v5.go
```diff
@@ -0,0 +1,457 @@
...
+				Advanced().
+				Optional(),
+			service.NewInterpolatedStringField(mov5FieldPayloadFormatIndicator).
+				Description("Whether the payload is UTF-8 text: `1` if it is, `0` if it is unspecified bytes. Must resolve to `0` or `1`.").
+				Advanced().
+				Optional(),
+			service.NewObjectField(mov5FieldMetadata,
+				service.NewStringListField("exclude_prefixes").
+					Description("Metadata keys beginning with any of these are not sent.").
+					Default([]any{"mqtt_"}),
```
jem-davies: Don't think we should set that as a default - without the exclude_prefixes in the config - I think people would assume it was empty.

jem-davies (commented):
Thanks for this! - look like we are filling a big gap with this PR & MQTT v5 

I have spent time reading and checking things and it is looking really good. 

I think the two biggest things so far are: 

 - reconnections aren't being captured by metrics (at least on the output, presumably on the input as well) 
 - the watchProgress() go-routine 

And then more generally the verbosity of the documentation, assumptions it makes regarding it's use as a bridge for MQTT <--> MQTT, I would want to rewrite it such it reads more terse and to the point, but not necessarily remove it. 

 ---

ifnesi:
Thanks @jem-davies — this was a genuinely useful review, and both of the big ones were right. Pushed, rebased on `main`, and the changelog conflict is resolved (our entry now sits alongside the `statsd` one in `Unreleased` rather than replacing it).

## Reconnection and the metrics

Conceded, and done the way you suggested: `OnConnectionDown` now returns `false`, so a dropped link ends that manager and the components report `service.ErrNotConnected`. Bento applies its own backoff and calls `Connect`, which builds another. autopaho still owns the dial — inside one `Connect` it retries on `reconnect_backoff` until the link is up.

Three things fell out of it that were not obvious from the outside:

- **A rebuilt manager must never clean-start.** `clean_start` speaks about the first connection of the component's life; honouring it again on a reconnection would discard the very session it exists to resume. Forced off after the first successful connection.
- **The down signal is renewed by `connect` rather than by the connection-up handler.** autopaho releases `AwaitConnection` *before* calling that handler, so a signal renewed there can still be the closed one at the moment `connect` returns — and the first read afterwards then reports a live connection as down. Moving it is only safe now that every connection begins with a `Connect` call.
- **A manager that has stopped has to be released, and this one is worth flagging.** Returning `false` breaks autopaho's `mainLoop`, but its deferred shutdown does `queueWg.Wait()` before `close(c.done)`, and that goroutine (`managePublishQueue`) only exits when the manager's context is cancelled. So `Done()` never closes on this path and every reconnection would leak a pair of goroutines plus its queue. Raised upstream as eclipse-paho/paho.golang#349 with a reproduction and a one-line fix; here the next `Connect` disconnects the retired manager, which cancels before it waits, and that stays correct against every released version whatever upstream decides.

There is a new integration test for the whole cycle. It counts SUBACKs rather than timing anything: one before the drop, still one while it is down — so nothing reconnects unasked — and two after Bento's reconnection, on a session the server keeps nothing of, where re-subscribing is the only way the last message can arrive.

Measured on a live stack (two pipelines, broker restarted underneath them), Bento's own metrics now read `connection_lost=1`, `connection_up=2`, `connection_failed=1` on both the input and the output side, with `output_error=0` and no `input_error` series at all. That was the point of the change, so it seemed worth checking rather than asserting.

## `watchProgress`

Removed, with its tests and the three paragraphs of documentation it needed. Your framing was right: every input that acknowledges in order can reach that state, so answering it in this one component was the wrong shape. Happy to open a separate issue with the reasoning if it is useful — the condition is real, it is just not this input's to detect.

The warning logged the first time a message is rejected stays, since that one is about this component's own acknowledgement ordering.

## The metadata default

Conceded — `exclude_prefixes` now defaults to `[]`. You are right that a field nobody configured should not be holding something back. What the default was protecting is now stated where somebody bridging will read it: the field's own docs, the bridging example, and the input's metadata section each name the prefix to exclude.

## Documentation

Trimmed both components: the acknowledgement, refusal and metadata sections, and the bridge framing you flagged — MQTT-to-MQTT is now one example rather than a section that assumes it is the destination. Please still reshape it however you like; better that you start from something shorter than from a blank page.

## The nits

All taken. Both components name the experimental dependency in their rendered docs, as the parquet components do. `receive_maximum` no longer talks about receiving on an output. The duplicate "Connection established." log is gone, since Bento already logs the component active. The `LintRule(root = [])` suppression and the two self-explanatory comments are deleted. And `reconnect_backoff` is now checked by a `LintRule` on the field so a bad value is caught at config time — I kept the Go checks behind it, because `autopaho.NewExponentialBackoff` panics rather than returning an error and a lint can be suppressed.

On your two "should this go upstream?" questions: yes to both, I think. The empty `ReasonString`/`ServerReference` is the same class of thing as #349 and I am happy to file it; the QoS 2 refusal arriving as a success is the sharper one, since it means a rejected publication can look delivered, and that deserves its own report with a reproduction.

jem-davies:
OK great - I am currently traveling and will try to get a second look when I can. 

> On your two "should this go upstream?" questions: yes to both, I think. The empty `ReasonString`/`ServerReference` is the same class of thing as #349 and I am happy to file it; the QoS 2 refusal arriving as a success is the sharper one, since it means a rejected publication can look delivered, and that deserves its own report with a reproduction.

In that case - could you file those issues and reference them in code comments at appropriate places? 

We will be merging it in as experimental so that shouldn't block. 

In the first review I did focus more on the `output` / general Bento idioms etc. - will want to get to look at the input when I can. 

Let me know if that works for you @ifnesi, thanks!

ifnesi:
That works, thanks, and enjoy the travelling.

Both are filed: eclipse-paho/paho.golang#352 for the QoS 2 refusal and #353 for the CONNACK properties. I'll push the code comments referencing them shortly.

Two corrections from writing them up. The QoS 2 one is worse than I described: PUBREC returns the refusal with a nil error *and* PUBCOMP has no reason-code check at all, so both paths can report a rejected publication as delivered. And the CONNACK one is narrower than it looked. `autopaho.ConnackError` does carry the reason code, so what we hardcode is only the reason *text*; what the library actually drops is `ServerReference`, which is the address a `0x9C Use another server` refusal is pointing at.

One option on the comments, take it or leave it: rather than a comment alone, I can pin each workaround with a test that asserts the current upstream behaviour, so it fails when paho fixes it. Matt turned #349 around in a day, so a plain comment may go stale before Bento's next release without anyone noticing.

No rush at all on the input, happy for that feedback after it merges as experimental.

jem-davies (approved):
OK - let's get this in - we can continue to work on it - as it will be experimental whilst the dependency is v0.

PR #993, opened 2026-08-26
fix kafka franz output: broker_write_max_bytes misconfigured

### LINKED ISSUE #992: bug(kafka_franz output): broker_write_max_bytes always 0 — wrong variable assigned in PR #981
## Description

  `kafka_franz` output in v1.21.0 fails to connect with:

  Failed to connect to kafka_franz: max broker write bytes 0 is less than allowed 1024

  Setting `broker_write_max_bytes` explicitly in config has no effect.

  ## Root Cause

  In `internal/impl/kafka/output_kafka_franz.go` (introduced in PR #981), the parsed `brokerWriteMaxBytes` value is never assigned to the struct field. Instead, `maxBufferedBytes` is used by mistake:

  ```go
  f.brokerWriteMaxBytes = int32(maxBufferedBytes)  // BUG: should be brokerWriteMaxBytes
  ```

  Since max_buffered_bytes defaults to "0", the struct field is always 0 regardless of what the user sets for broker_write_max_bytes. The franz-go library then rejects 0 at connect time.

  Fix

  ```go
  f.brokerWriteMaxBytes = int32(brokerWriteMaxBytes)
  ```

  Reproduction

  Any kafka_franz output in v1.21.0 will fail. Minimal config:

  output:
    kafka_franz:
      seed_brokers: ["localhost:9092"]
      topic: test

  Workaround

  Pin image to v1.20.0.

  Environment

  - Bento v1.21.0 (ghcr.io/warpstreamlabs/bento:1.21.0)
  - Kafka (Strimzi, SCRAM-SHA-512, TLS)
 
> Note: PR #981 shipped with no test coverage for the new fields. The existing output_kafka_franz_test.go only tests partitioner and seed broker validation. A unit test calling newFranzKafkaWriterFromConfig with default config and asserting f.brokerWriteMaxBytes != 0
   would have caught this.

jem-davies:
fixes #992

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,12 @@ Changelog
 
 All notable changes to this project will be documented in this file.
 
+## 1.21.1 - 2026-08-21
+
+### Fixed
+
+- `kafka_franz` output field `broker_write_max_bytes` misconfigured (introduced in version 1.21.0) @jem-davies
+
 ## 1.21.0 - 2026-08-21
 
 ### Added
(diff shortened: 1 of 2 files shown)


gregfurman (approved):

gregfurman:
Let's add a changelog entry then cut a release

PR #995, opened 2026-08-26
chore(deps/sql): Migrate mssql driver to official microsoft/go-mssqldb

cmelanson:
## Overview

This PR changes the driver Bento is using for connecting to Microsoft SQL Server from `denisenkom/go-mssqldb` to the offically supported fork from Microsoft `microsoft/go-mssqldb`. The old driver is no longer maintained, with the last commit being April 30, 2023 and the [readme](https://github.com/denisenkom/go-mssqldb#a-pure-go-mssql-driver-for-gos-databasesql-package) for the project directs users to the `microsoft/go-mssqldb` fork for updates.

Also, Microsoft is starting the process of deprecating NTLM authentication in Windows in the second half of 2026 ([blog post](https://techcommunity.microsoft.com/blog/windows-itpro-blog/advancing-windows-security-disabling-ntlm-by-default/4489526)) with Kerberos being the more secure alternative. The `denisenkom/go-mssqldb` driver does not support Kerberos authentication.

## Change

In [`public/components/sql/package.go`](public/components/sql/package.go):

```diff
- _ "github.com/denisenkom/go-mssqldb"
+ _ "github.com/microsoft/go-mssqldb"
+ _ "github.com/microsoft/go-mssqldb/integratedauth/krb5"
```

Users can then opt-in to Kerberos authentication in the DSN, for example using username, password, domain and a custom krb5.conf:
```
sqlserver://user:password@host:1433?authenticator=krb5&krb5-realm=DOMAIN.COM&krb5-configfile=/path/to/krb5.conf&database=dbname
```

### Compatibility

- Both the old and new drivers register the same `mssql` driver name
- Existing DSNs should still work, as this new driver is a fork
- Kerberos support is opt-in and requires the user to add `authenticator=krb5` to the DSN
- The `integratedauth/krb5` package has a dependency on the `jcmturner/gokrb5` library, which is already being used in the Kafka library and is the same version (`v8.4.4`).

### Testing

I have run the following commands and everything completed without errors:
- `make test`
- `go test ./internal/impl/sql/ -run TestIntegrationMSSQL -v`
- `make fmt`
- `make lint`

I have also manually tested using Kerberos username and password authentication against an AD joined MSSQL 2025 server. This was successful but I haven't added a test case for this scenario, as the setup for it is complicated.


public/components/sql/package.go
```diff
@@ -10,10 +10,11 @@ import (
 
 	// Import all (supported) sql drivers.
 	_ "github.com/ClickHouse/clickhouse-go/v2"
-	_ "github.com/denisenkom/go-mssqldb"
 	_ "github.com/go-sql-driver/mysql"
 	_ "github.com/googleapis/go-sql-spanner"
 	_ "github.com/lib/pq"
+	_ "github.com/microsoft/go-mssqldb"
+	_ "github.com/microsoft/go-mssqldb/integratedauth/krb5"
```
gregfurman, line 17: Wow that is looking much better!

gregfurman (approved):
LGTM! Thanks for the contribution. Could you change the title to something along the lines of:

> chore(deps/sql): Migrate mssql driver to official microsoft/go-mssqldb

Also, how is kerberos auth being enabled? Does it happen via the DSN?

public/components/sql/package.go
cmelanson: Of course, title has been updated.

Yes, Kerberos auth is enabled by adding `authenticator=krb5` to the DSN. Would you like me to add a blurb to the docs that this is an option, with a direct link to the [docs](https://github.com/microsoft/go-mssqldb#kerberos-parameters) that lists other optional related parameters?

gregfurman: Adding a small section on the mssql server DSN section would be super helpful! 

FYI you'll have to run `make docs` on your local afterwards which will end up touching quite a few files, since the DSN field stuff is shared by all SQL components.

Just ping me when that's done and I'll give another look 🙂 \

cmelanson: I've added a small section on usage with Kerberos to the documentation with a minimal example. Please let me know if you wish to see any changes to it (wording, format, etc.) :slightly_smiling_face:

PR #998, opened 2026-08-28
align YAML merge key (<<:) handling between lint & run subcommands

### LINKED ISSUE #872: lint accepts YAML merge keys (<<) but config init ignores them — config lints clean, fails to start

jem-davies:
fixes #872 

It seems odd how the `lint` subcommand is able to handle YAML merge keys but running a stream that has merge keys will fail to init components.

This PR will alter the `lint` subcommand such it will behave the same as `run`. Issue #872 describes a situation where `lint` was called in CI on a config, and reported no errors, but then subsequently `run` failed. 

Related to #969 - which enables Bento to work with <<: merge commands.

### DIFF
--- a/internal/docs/format_yaml.go
+++ b/internal/docs/format_yaml.go
@@ -681,8 +681,7 @@ func (f FieldSpecs) LintYAML(ctx LintContext, node *yaml.Node) []Lint {
 		specNamesAll[field.Name] = field
 	}
 
-	var walkNodeContent func(*yaml.Node)
-	walkNodeContent = func(walkNode *yaml.Node) {
+	walkNodeContent := func(walkNode *yaml.Node) {
 		for i := 0; i < len(walkNode.Content)-1; i += 2 {
 			if walkNode.Content[i].Tag == "!!merge" {
 				continue
--- a/internal/docs/format_yaml.go
+++ b/internal/docs/format_yaml.go
@@ -684,8 +684,7 @@ func (f FieldSpecs) LintYAML(ctx LintContext, node *yaml.Node) []Lint {
 	var walkNodeContent func(*yaml.Node)
 	walkNodeContent = func(walkNode *yaml.Node) {
 		for i := 0; i < len(walkNode.Content)-1; i += 2 {
-			if walkNode.Content[i].Tag == "!!merge" && walkNode.Content[i+1].Alias != nil {
-				walkNodeContent(walkNode.Content[i+1].Alias)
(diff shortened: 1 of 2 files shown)


internal/docs/format_yaml.go
```diff
@@ -681,11 +681,9 @@ func (f FieldSpecs) LintYAML(ctx LintContext, node *yaml.Node) []Lint {
 		specNamesAll[field.Name] = field
 	}
 
-	var walkNodeContent func(*yaml.Node)
-	walkNodeContent = func(walkNode *yaml.Node) {
+	walkNodeContent := func(walkNode *yaml.Node) {
 		for i := 0; i < len(walkNode.Content)-1; i += 2 {
-			if walkNode.Content[i].Tag == "!!merge" && walkNode.Content[i+1].Alias != nil {
-				walkNodeContent(walkNode.Content[i+1].Alias)
```
gregfurman, line 688: Instead of removing the behaviour from `lint`, how come we don't extend `run` to handle this case?

PR #1000, opened 2026-08-31
feat(gcp): add service account impersonation to GCP components

### LINKED ISSUE #996: GCP service account impersonation not supported
Hi, we are one of the heavy users of Bento. Thanks for the tool.

### Version

main/v1.20.x

### Problem 
Currently, we have encountered an issue regarding **GCP impersonation**. We are running **Bento multi-streams** in Cloud Run service, but due to least privilege principle each stream must have a dedicated Service Account with a different set of permissions. 

We have checked carefully and see that there is no way that this could be done. 

Is there any workaround solution? If not, we would love to open a PR for this feature.

NguyenThanhDat741862:
Adds an advanced `credentials` object to every GCP component (gcp_pubsub input/output, gcp_cloud_storage input/output/cache, gcp_bigquery, gcp_bigquery_write_api, gcp_bigquery_select input/processor, gcp_spanner_cdc, gcp_bigtable, gcp_cloudtrace) with:

- `impersonate_service_account`: email of a service account to impersonate using the base Application Default Credentials.
- `impersonate_delegates`: optional delegation chain.

When unset, behaviour is unchanged (plain ADC). This allows a single Bento instance (e.g. streams mode on Cloud Run) to run each stream as its own least-privileged service account.

Closes #996


internal/impl/gcp/credentials.go
```diff
@@ -0,0 +1,74 @@
...
+			Example("target@my-project.iam.gserviceaccount.com").
+			Default(""),
+		service.NewStringListField(gcpFieldImpersonateDelegates).
+			Description("An optional chain of service account email addresses to delegate through when impersonating. Each service account must be granted `roles/iam.serviceAccountTokenCreator` on the next service account in the chain, with the final one granted on `"+gcpFieldImpersonateServiceAccount+"`.").
+			Default([]any{}),
+	).
+		Description("Optional manual configuration of GCP credentials to use. More information can be found [in this document](/docs/guides/cloud/gcp).").
+		Advanced().
+		Version("1.22.0").
+		LintRule(`root = if this.impersonate_delegates.length() > 0 && this.impersonate_service_account == "" { "impersonate_delegates requires impersonate_service_account to be set" }`)
```
gregfurman, line 36: praise: Nice use of `LintRule`!

internal/impl/gcp/credentials.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+
+	delegates, err := cConf.FieldStringList(gcpFieldImpersonateDelegates)
+	if err != nil {
+		return nil, err
+	}
+
+	ts, err := impersonate.CredentialsTokenSource(context.Background(), impersonate.CredentialsConfig{
+		TargetPrincipal: target,
+		Delegates:       delegates,
```
gregfurman, line 66: Q: I know we have a `LintRule` preventing this from occurring, but would just like to confirm that this request will fail if `target` is empty and `delegates` is defined?

internal/impl/gcp/output_bigquery_storage.go
```diff
@@ -198,7 +206,7 @@ type bigQueryStorageWriterConfig struct {
 
 func (bq *bigQueryStorageWriter) Connect(ctx context.Context) error {
 
-	var bqClientOpts []option.ClientOption
+	bqClientOpts := slices.Clone(bq.conf.clientOptions)
```
gregfurman, line 209: nit: FYI you can just set the value of `bqClientOpts` directly to the slice at `bq.conf.clientOptions` since doing an `append` on `bqClientOpts` wont affect the original

internal/impl/gcp/credentials_test.go
```diff
@@ -0,0 +1,99 @@
...
+`, nil)
+	require.NoError(t, err)
+
+	opts, err := gcpClientOptionsFromParsed(pConf)
+	require.NoError(t, err)
+	require.Len(t, opts, 1)
+}
+
+func TestGCPCredentialsTargetOnlyLintHappy(t *testing.T) {
+	err := service.NewStreamBuilder().SetYAML(`
```
gregfurman: Think it'd be cleaner to go via GCP specific configuration specs as opposed to the public builder for this check e.g for the pubsub input, you can create a new spec:

https://github.com/warpstreamlabs/bento/blob/80189d9dbbb19d53b7819fd6260cb316dbdf540f/internal/impl/gcp/input_pubsub.go#L70

and then use it to parse yaml/config:

```golang
pbiSpec().ParseYAML(pbSpec)
gcpCloudStorageCacheConfig().ParseYAML(csSpec) // or for cloud storage...
```

Also, the current approach requires that `pure` be imported (allowing for that `drop` output to resolve) which is unnecessary here.

internal/impl/gcp/credentials_test.go
```diff
@@ -0,0 +1,99 @@
...
+	})
+	require.NoError(t, err)
+
+	path := filepath.Join(t.TempDir(), "sa.json")
+	require.NoError(t, os.WriteFile(path, saJSON, 0o600))
+	t.Setenv("GOOGLE_APPLICATION_CREDENTIALS", path)
+}
+
+func TestGCPCredentialsDefaultNoOptions(t *testing.T) {
+	pConf, err := testCredentialsSpec().ParseYAML(`{}`, nil)
```
gregfurman: suggestion: I think a small test table here with valid and invalid configurations would be useful, since these tests don't really make behaviour of the fields too clear

website/docs/guides/cloud/gcp.md
```diff
@@ -17,3 +17,38 @@ as a JSON file. Then all you need to do set the path to this JSON file in the `G
...
+Delegated impersonation chains can be expressed with `impersonate_delegates`, where each service account must be
+granted `roles/iam.serviceAccountTokenCreator` on the next one in the chain, and the final delegate on the target:
+
+```yaml
+credentials:
+  impersonate_service_account: target@my-project.iam.gserviceaccount.com
+  impersonate_delegates:
+    - hop-one@my-project.iam.gserviceaccount.com
+    - hop-two@my-project.iam.gserviceaccount.com
+```
```
gregfurman, line 55: This will fail CI since the docs linter cannot interpret the detached `credentials` section as valid Bento config.

https://github.com/warpstreamlabs/bento/actions/runs/33353870005/job/99450546843?pr=1000

Either attach it to a component (like you did above with `input.gcp_pubsub`) or change the `yaml` to `yml`, which will disable the component linter in favour of just linting the yaml.

gregfurman (commented):
This is looking great! Have left a couple of comments. 

Think adding an integration test of some kind would be useful here just to verify API behaviour but am unsure how much extra work that would entail. Perhaps even a mock testing server https://pkg.go.dev/net/http/httptest that we redirect these credentials requests to could help since I'm unclear on how components will behave when these credential requests fail.

In addition, the current tests could be consolidated a bit, with a test table that exhaustively checks valid + invalid config probably being the best indicator. Since we never actually test the impersonation behaviour (and only ever fake it) so I'd be curious to see how Bento behaves when these auth requests fail. As mentioned, perhaps an integration test or test server could help clear this up. Wdyt?

Otherwise, CI is failing because of our whacky internal document linter, have left a comment on how to reconcile this.

internal/impl/gcp/credentials.go
NguyenThanhDat741862: No request was made at all. Lint catches it in files, but env-interpolated configs bypass linting, so I've added a runtime guard returning an error for any delegates without target.

internal/impl/gcp/credentials_test.go
NguyenThanhDat741862: Done: 5-case table for the options behaviour, 3-case for lint

internal/impl/gcp/credentials_test.go
NguyenThanhDat741862: Done: I have dropped the StreamBuilder/pure dependency

internal/impl/gcp/output_bigquery_storage.go
NguyenThanhDat741862: The clone is deliberate here, two slices are appended from the same base, and if the base ever has spare capacity the second append would overwrite the first through the shared backing array. Currently, cap==len so direct assignment happens to be safe, but I would rather not depend on that invariant at a distance

NguyenThanhDat741862:
> This is looking great! Have left a couple of comments.
> 
> Think adding an integration test of some kind would be useful here just to verify API behaviour but am unsure how much extra work that would entail. Perhaps even a mock testing server https://pkg.go.dev/net/http/httptest that we redirect these credentials requests to could help since I'm unclear on how components will behave when these credential requests fail.
> 
> In addition, the current tests could be consolidated a bit, with a test table that exhaustively checks valid + invalid config probably being the best indicator. Since we never actually test the impersonation behaviour (and only ever fake it) so I'd be curious to see how Bento behaves when these auth requests fail. As mentioned, perhaps an integration test or test server could help clear this up. Wdyt?
> 
> Otherwise, CI is failing because of our whacky internal document linter, have left a comment on how to reconcile this.

I've added coverage at both levels:

-  Live integration test: `integration_impersonation_test.go` is an opt-in test (gated by -run + env vars, so CI is unaffected) that runs against real GCP, because there is no emulator for the IAM Credentials API. It covers direct impersonation, delegated impersonation through an intermediate SA, and a split configuration where input and output impersonate different SAs each holding only their own side's permissions, so a pass proves each component used its own configured identity. `resources/impersonation_e2e.sh` provisions everything (within GCP free tier) and prints the `go test` command. The base ADC is deliberately a powerless SA so success is only attributable to impersonation. All subtests pass against a live project.
Screenshot 2026-09-01 at 18 16 24


- Failure behaviour via httptest: `TestGCPCredentialsImpersonationFailure` does what you have suggested. A local server plays the IAM Credentials API and denies `generateAccessToken`. It pins the answer to your question, component construction succeeds (tokens are minted lazily), and the first operation fails with an error carrying the impersonation context and status code. One wrinkle worth knowing is that `impersonate.CredentialsTokenSource` hardcodes the `iamcredentials.googleapis.com` URL (`impersonate.go:177` in v0.259.0), so `option.WithEndpoint` is silently ignored, the redirect has to go through `option.WithHTTPClient` with a rewriting RoundTripper, which is what the test does. `gcpClientOptionsFromParsed` gained a variadic `...option.ClientOption` as the injection seam.

website/docs/guides/cloud/gcp.md
NguyenThanhDat741862: I have fixed by attaching the snippet to a gcp_pubsub input

internal/impl/gcp/integration_impersonation_test.go
gregfurman, line 1: Tbh I think the `credentials_test.go` you've written should suffice. The integration test will fail when run in CI since we don't have access to a live environment. Think we can remove this (and the script) for now, but I'll use it for a manual verification in anycase 👍

internal/impl/gcp/credentials_test.go
```diff
@@ -44,56 +50,185 @@ func setFakeADC(t *testing.T) {
...
+			}
+			require.NoError(t, err)
+			require.Len(t, opts, test.wantOpts)
+		})
+	}
 }
 
-func TestGCPCredentialsImpersonation(t *testing.T) {
+// redirectTransport rewrites every request to the given test server host,
+// since the impersonate library hardcodes the IAM Credentials API URL.
```
gregfurman: Can we remove these extra comments? They're kinda polluting the diff a bit 🧹

gregfurman (changes requested):
Thanks for addressing my feedback! Think we can drop the integration test for now in favour of just keeping that extra impersonation unit test

internal/impl/gcp/integration_impersonation_test.go
NguyenThanhDat741862: One note for the record. It could not actually fail CI,  `make test` never matches it due to `integration.CheckSkip`, and `make test-integration` skips it without the `GCP_E2E_*` env vars. However, I agreed that it is reasonable to keep it out of the tree for now

internal/impl/gcp/credentials_test.go
NguyenThanhDat741862: Done, I have trimmed these redundant comments

npsables:
Hi @gregfurman, I believe the PR is ready? Kindly re-review it when you have time.

gregfurman:
@NguyenThanhDat741862 Thanks for the followups + ping! Am giving this another look 👀

gregfurman (approved):
LGTM! Once CI is green, we can ship this bad-boy. Thanks for the contribution :shipit:

PR #1004, opened 2026-09-05
fix(fsevent): reconnect on watcher error instead of stalling

henrikschristensen:
The fsevent input had no recovery path when the underlying fsnotify watch died. On Windows, when a watched path is a symlink to an SMB share and the SMB session is torn down (typically overnight), the async read completes with ERROR_NETNAME_DELETED ("The specified network name is no longer available"). This surfaced two dead ends, both requiring a manual process restart:

  - the watcher error was only logged while the read loop kept running against a dead watch, so Read blocked forever (silent stall); and
  - the broken-watch path returned ErrEndOfInput, which the enclosing `sequence` input treats as graceful termination and shuts down.

On a watcher error we now tear down the read loop, and the closed-watch path in Read returns ErrNotConnected (closing the dead watcher) instead of ErrEndOfInput. The framework's async reader then re-runs Connect with backoff, rebuilding the watch and resuming file pickup with no restart.

Adds a white-box regression test that injects a watcher error and asserts Read returns ErrNotConnected and that new files are picked up after reconnect. Verified failing before the fix (Read blocks) and passing after.

### DIFF
--- a/internal/impl/io/input_fsevent.go
+++ b/internal/impl/io/input_fsevent.go
@@ -205,7 +205,15 @@ func (f *fsEventWatcher) Connect(ctx context.Context) error {
 				if !ok {
 					return
 				}
-				f.log.Errorf("error: %v", err)
+				// A watcher error (for example a watched path backed by an SMB
+				// share reporting "The specified network name is no longer
+				// available" once the session is torn down) leaves the
+				// underlying fsnotify watch dead with no way to recover it in
+				// place. Stop the read loop so that Read surfaces a disconnect
+				// and the framework rebuilds the watcher via Connect, instead of
+				// silently blocking until the process is restarted.
+				f.log.Errorf("filesystem watcher error, reconnecting: %v", err)
+				return
 			}
 		}
 	}()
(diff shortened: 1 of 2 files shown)


internal/impl/io/input_fsevent.go
```diff
@@ -230,8 +238,15 @@ func (f *fsEventWatcher) Read(ctx context.Context) (*service.Message, service.Ac
...
-			f.watcher = nil
-			return nil, nil, service.ErrEndOfInput
+			if f.watcher != nil {
+				_ = f.watcher.Close()
+				f.watcher = nil
+			}
+			// The watch loop stopped (the watcher errored, or was closed
+			// because nothing remained to watch). Report a disconnect rather
+			// than end-of-input so the framework reconnects and rebuilds the
+			// watch, instead of permanently terminating the input.
```
gregfurman, line 248:
```suggestion
```

This the above comment is implied by the error type, so we can remove this IMO

internal/impl/io/input_fsevent.go
```diff
@@ -230,8 +238,15 @@ func (f *fsEventWatcher) Read(ctx context.Context) (*service.Message, service.Ac
...
+			return nil, nil, service.ErrNotConnected
```
gregfurman, line 249: Hmm... I feel like there are certain cases where we want the pipeline to finish running. Like, what if I'm watching a single directory that is then deleted.

I suppose this case could trigger a reconnect, which could then in turn fail with `service.ErrEndOfInput`.

gregfurman (approved):
LGTM! Nothing blocking, so feel free to resolve. Else, ping me when ready and I can merge 👍

henrikschristensen:
@gregfurman should be ready for merge

PR #1007, opened 2026-09-06
feat(bloblang): add precision and rounding style params to round method

### LINKED ISSUE #1005: Extend the `round` Bloblang method with precision and rounding style params
Bloblang currently has no way to round a number to a given number of decimal places, and no way to choose a rounding mode. round() only rounds to whole numbers (half away from zero), and the usual workarounds are both flawed: (this.value * 100).round() / 100 suffers float double-rounding (2.675 * 100 is not exactly 267.5 in general), and this.value.format("%.2f").number() is a string round-trip locked to one rounding mode.

Use case: we process financial (FIX protocol) messages where prices and quantities must be rounded to tick-size precision under specific, contractually mandated rounding modes — banker's rounding (half_even), commercial rounding (half_up), or truncation depending on the venue and jurisdiction.

Proposal: extend round with two optional parameters, fully backwards compatible (no-arg behaviour unchanged):

root.price = this.price.round(precision: 2, style: "half_even")
root.lot = this.qty.round(precision: -2)   # round to hundreds

precision (int, default 0): decimal places; negative values round to tens/hundreds/etc.
style (string, default "half_up"): one of half_up (ties away from zero — the current behaviour), half_even (banker's rounding), truncate (towards zero). Names follow the Java BigDecimal / Python decimal vocabulary.
Rounding is performed on the shortest decimal representation of the value, so 2.675.round(precision: 2) is exactly 2.68 with no float-shift artefacts. Result typing follows the existing Numbers-category contract: whole results that fit in a 64-bit integer are returned as integers.

Happy to implement this as a PR (two commits: precision, then styles, so they can be reviewed/merged independently), with tests and generated docs.

henrikschristensen:
Closes #1005                                                                                                                                                                                                  
                                                                                                                                                                                                                   
     Extends the `round` Bloblang method with two optional parameters, fully backwards compatible — no-arg behaviour is unchanged.                                                                                 
                                                                                                                                                                                                                   
     ## Changes                                                                                                                                                                                                    
                                                                                                                                                                                                                   
     **Commit 1 — `precision` (int, default 0):**                                                                                                                                                                  
     - Number of decimal places to round to; negative values round to tens/hundreds/etc. (`round(2.5, -2)` → `0`, `round(250, -2)` → `200`... etc.)                                                                
     - Rounding is performed on the shortest decimal representation of the value, so `2.675.round(precision: 2)` is exactly `2.68` — no float-shift artefacts from the `(x * 100).round() / 100` workaround.       
                                                                                                                                                                                                                   
     **Commit 2 — `style` (string, default `"half_up"`):**                                                                                                                                                         
     - `half_up` — ties away from zero (the current behaviour)                                                                                                                                                     
     - `half_even` — banker's rounding                                                                                                                                                                             
     - `truncate` — towards zero                                                                                                                                                                                   
     - Names follow the Java BigDecimal / Python decimal vocabulary.                                                                                                                                               
                                                                                                                                                                                                                   
     The commits are split so precision and styles can be reviewed/merged independently.                                                                                                                           
                                                                                                                                                                                                                   
     ```coffee                                                                                                                                                                                                     
     root.price = this.price.round(precision: 2, style: "half_even")                                                                                                                                               
     root.lot = this.qty.round(precision: -2)   # round to hundreds                                                                                                                                                
   ```                                                                                                                                                                                                             
                                                                                                                                                                                                                   
   Result typing follows the existing Numbers category contract: whole results that fit in a 64-bit integer are returned as integers.                                                                              
                                                                                                                                                                                                                   
   Testing                                                                                                                                                                                                         
                                                                                                                                                                                                                   
   • Unit tests added in internal/bloblang/query/methods_test.go covering precision (incl. negative), all three styles, tie cases, and backwards compatibility of the no-arg form.                                 
   • Generated docs updated (website/docs/guides/bloblang/methods.md).

### DIFF
--- a/internal/bloblang/query/methods_numbers.go
+++ b/internal/bloblang/query/methods_numbers.go
@@ -207,7 +207,7 @@ var _ = registerSimpleMethod(
 
 var _ = registerSimpleMethod(
 	NewMethodSpec(
-		"round", "Rounds numbers to the nearest value with a given number of decimal places, defaulting to the nearest integer, rounding half away from zero. If the resulting value fits within a 64-bit integer then that is returned, otherwise a new floating point number is returned.",
+		"round", "Rounds numbers to the nearest value with a given number of decimal places, defaulting to the nearest integer. The rounding style for ties can be specified and defaults to half away from zero. If the resulting value fits within a 64-bit integer then that is returned, otherwise a new floating point number is returned.",
 	).InCategory(
 		MethodCategoryNumbers,
 		"",
@@ -223,10 +223,20 @@ var _ = registerSimpleMethod(
 			`{"value":2.675}`,
 			`{"new_value":2.68}`,
 		),
+		NewExampleSpec("",
+			`root.new_value = this.value.round(precision: 2, style: "half_even")`,
+			`{"value":0.125}`,
+			`{"new_value":0.12}`,
+		),
 	).
-		Param(ParamInt64("precision", "The number of decimal places to round to. Negative values round to tens, hundreds, and so on.").Optional()),
+		Param(ParamInt64("precision", "The number of decimal places to round to. Negative values round to tens, hundreds, and so on.").Optional()).
+		Param(ParamString("style", "The rounding style to use when the value lies exactly halfway between two candidates: `half_up` rounds ties away from zero, `half_even` rounds ties to the nearest even digit (banker's rounding), and `truncate` always rounds towards zero.").Default("half_up")),
 	func(args *ParsedParams) (simpleMethod, error) {
 		precision, err := args.FieldOptionalInt64("precision")
+		if err != nil {
+			return nil, err
+		}
+		style, err := args.FieldString("style")
 		if err != nil {
 			return nil, err
 		}
@@ -234,8 +244,11 @@ var _ = registerSimpleMethod(
 		if precision != nil {
 			p = *precision
 		}
+		if style != "half_up" && style != "half_even" && style != "truncate" {
+			return nil, fmt.Errorf("unknown rounding style %q: must be one of half_up, half_even, truncate", style)
+		}
 		roundAndCoerce := func(v float64) (any, error) {
-			rounded := roundToPrecision(v, p)
+			rounded := roundToPrecision(v, p, style)
 			if i, err := value.IToInt(rounded); err == nil {
 				return i, nil
 			}
@@ -259,11 +272,11 @@ var _ = registerSimpleMethod(
 	},
 )
 
-// roundToPrecision rounds v to the given number of decimal places, with ties
-// rounding away from zero. The value is rounded as its shortest decimal
-// representation, so 2.675 rounds to 2.68 at a precision of 2, free of float
+// roundToPrecision rounds v to the given number of decimal places using the
+// given style. The value is rounded as its shortest decimal representation,
+// so 2.675 rounds to 2.68 at a precision of 2 under half_up, free of float
 // shift artefacts.
-func roundToPrecision(v float64, precision int64) float64 {
+func roundToPrecision(v float64, precision int64, style string) float64 {
 	if math.IsNaN(v) || math.IsInf(v, 0) {
 		return v
 	}
@@ -295,7 +308,18 @@ func roundToPrecision(v float64, precision int64) float64 {
 	}
 
 	kept, discarded := digits[:point], digits[point:]
-	roundUp := discarded[0] >= '5'
+	roundUp := false
+	switch style {
+	case "half_up":
+		roundUp = discarded[0] >= '5'
+	case "half_even":
+		switch {
+		case discarded[0] > '5':
+			roundUp = true
+		case discarded[0] == '5':
+			roundUp = strings.TrimLeft(discarded[1:], "0") != "" || (kept != "" && kept[len(kept)-1]%2 == 1)
+		}
+	}
 
 	if roundUp {
 		b := []byte(kept)
--- a/internal/bloblang/query/methods_numbers.go
+++ b/internal/bloblang/query/methods_numbers.go
@@ -4,6 +4,8 @@ import (
 	"errors"
 	"fmt"
 	"math"
+	"strconv"
+	"strings"
 
 	"github.com/warpstreamlabs/bento/internal/value"
 )
@@ -205,7 +207,7 @@ var _ = registerSimpleMethod(
 
 var _ = registerSimpleMethod(
 	NewMethodSpec(
-		"round", "Rounds numbers to the nearest integer, rounding half away from zero. If the resulting value fits within a 64-bit integer then that is returned, otherwise a new floating point number is returned.",
+		"round", "Rounds numbers to the nearest value with a given number of decimal places, defaulting to the nearest integer, rounding half away from zero. If the resulting value fits within a 64-bit integer then that is returned, otherwise a new floating point number is returned.",
 	).InCategory(
 		MethodCategoryNumbers,
 		"",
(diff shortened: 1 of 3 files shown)


internal/bloblang/query/methods_numbers.go
```diff
@@ -216,20 +218,141 @@ var _ = registerSimpleMethod(
...
+// so 2.675 rounds to 2.68 at a precision of 2 under half_up, free of float
+// shift artefacts.
+func roundToPrecision(v float64, precision int64, style string) float64 {
+	if math.IsNaN(v) || math.IsInf(v, 0) {
+		return v
+	}
+	// Beyond these bounds rounding is determined by float64's range alone:
+	// no finite float64 has more than ~330 significant decimal digits, and
+	// every finite float64 rounds to zero at a precision of -309 or below.
+	if precision > 400 {
```
gregfurman, line 286: Should we be using `math.Big` here?

internal/bloblang/query/methods_numbers.go
```diff
@@ -216,20 +218,141 @@ var _ = registerSimpleMethod(
...
+			return roundAndCoerce(float64(*ui))
 		}), nil
 	},
 )
+
+// roundToPrecision rounds v to the given number of decimal places using the
+// given style. The value is rounded as its shortest decimal representation,
+// so 2.675 rounds to 2.68 at a precision of 2 under half_up, free of float
+// shift artefacts.
+func roundToPrecision(v float64, precision int64, style string) float64 {
```
gregfurman, line 279: Can we split these into separate functions? Perhaps:

e.g

```go
func roundUp(val float64, precision uint) float64 {
	ratio := math.Pow(10, float64(precision))
	return math.Ceil(val*ratio) / ratio
}

func roundDown(val float64, precision uint) float64 {
	ratio := math.Pow(10, float64(precision))
	return math.Floor(val*ratio) / ratio
}

func roundTrunc(val float64, precision uint) float64 {
	ratio := math.Pow(10, float64(precision))
	return math.Trunc(val*ratio) / ratio
}
```

internal/bloblang/query/methods_numbers.go
```diff
@@ -216,20 +218,141 @@ var _ = registerSimpleMethod(
...
+			`{"new_value":2.68}`,
+		),
+		NewExampleSpec("",
+			`root.new_value = this.value.round(precision: 2, style: "half_even")`,
+			`{"value":0.125}`,
+			`{"new_value":0.12}`,
+		),
+	).
+		Param(ParamInt64("precision", "The number of decimal places to round to. Negative values round to tens, hundreds, and so on.").Optional()).
+		Param(ParamString("style", "The rounding style to use when the value lies exactly halfway between two candidates: `half_up` rounds ties away from zero, `half_even` rounds ties to the nearest even digit (banker's rounding), and `truncate` always rounds towards zero.").Default("half_up")),
```
gregfurman, line 233: I'm unsure about adding this parameter tbh

For these `half_up`/`half_down` cases could we not just do the following (for a precision of 5):

```coffee
root.up = (this * 10.pow(5)).ceil() / 10.pow(5) # always rounds up
root.down = (this * 10.pow(5)).floor() / 10.pow(5) # always rounds down
```

gregfurman (changes requested):
Hey. Thanks for the PR!

Am thinking this could be simplified to just add support for precision on round, defaulting to `0` as we currently do.

Unsure if we want to also be adding support for the round-down + round-up cases since they seem pretty easy to just do yourself.

Lmk 🙂

henrikschristensen:
> Hey. Thanks for the PR!
> 
> Am thinking this could be simplified to just add support for precision on round, defaulting to `0` as we currently do.
> 
> Unsure if we want to also be adding support for the round-down + round-up cases since they seem pretty easy to just do yourself.
> 
> Lmk 🙂

Good call — happy to split this. I'll pull style out into a follow-up so this PR is just precision defaulting to 0, no-arg behaviour untouched.                                                               
                                                                                                                                                                                                               
On the "easy to do yourself" bit though, a heads-up since it changes the maths: the shift-and-round trick mis-rounds half-way cases. Tried it with bento blobl on this branch:                                
                                                                                                                                                                                                               
```coffeescript                                                                                                                                                                                               
   # Input: {"value": 1.005}                                                                                                                                                                                   
   root.diy_round = (this.value * 10.pow(2)).round() / 10.pow(2)                                                                                                                                               
   root.pr_round  = this.value.round(2)                                                                                                                                                                        ```                                                                                                                                                                                                           
                                                                                                                                                                                                               
```                                                                                                                                                                                                           
  {"diy_round": 1, "pr_round": 1.01}   # 1.005 * 10.pow(2) comes out as 100.49999999999999                                                                                                                    
```                                                                                                                                                                                                           
                                                                                                                                                                                                               
Same for 1.255 → 1.25 (should be 1.26) and 35.855 → 35.85 (should be 35.86). My use case here is FX pricing, where a half-pip going the wrong way is real money — so "close enough" float rounding isn't an   
option for me, and that's why roundToPrecision works off the shortest decimal repr instead of shifting. If bento as a project is fine with approximate rounding I can collapse it to math.Round(v*pow)/pow,   
but then 1.005.round(2) gives 1, which feels like a trap waiting for the first person rounding prices 😄

File: internal/bloblang/query/methods_numbers.go (hunk 1 above)
Comment (henrikschristensen) [reply internal/bloblang/query/methods_numbers.go:286]: I looked at that and I don't think it buys us much — big.Float is binary floating point too (just wider), so it has the same decimal representation problem. Doing it properly would mean big.Rat scaled by  powers of 10, which is a fair bit more code and allocations for the same answer. The strconv.FormatFloat(v, 'f', -1, 64) shortest-repr trick already gives decimal-exact results (1.005.round(2) → 1.01 etc.  are in the tests), which is what I need for the FX pricing use case behind this PR. The 400/−309 bounds are just float64's range limits, not a precision cap on the algorithm 🙂

File: internal/bloblang/query/methods_numbers.go (hunk 2 above)
Comment (henrikschristensen) [reply internal/bloblang/query/methods_numbers.go:279]: Afraid those implementations bring back the exact artefact this PR is trying to fix — math.Round(1.005*100)/100 gives 1 instead of 1.01, because the product lands on 100.49999999999999. And the directed      
   versions aren't safe either: math.Floor(1.005*1e5)/1e5 gives 1.00499 instead of 1.005, and math.Ceil(0.55*100)/100 gives 0.56 instead of 0.55 (the product is 55.00000000000001, so ceil rounds up off an       
   already-exact value). Also, precision uint would lose negative precision (rounding to tens/hundreds), which this adds.                                                                                          
                                                                                                                                                                                                                   
   Worth noting too: ceil/floor/trunc are directed operations, but two of the three styles (half_up, half_even) are tie-break rules — half_even especially can't be built out of ceil/floor/trunc at all. So the   
   split doesn't quite map onto what the code does. That said, if we drop style like you suggested up top, the switch goes away and it's one simple function again anyway 👍

File: internal/bloblang/query/methods_numbers.go (hunk 3 above)
Comment (henrikschristensen) [reply internal/bloblang/query/methods_numbers.go:233]: Gave your snippet a spin on this branch (bento blobl, input 1.005):                                                                                                                                           
                                                                                                                                                                                                               
```coffeescript                                                                                                                                                                                               
  root.up   = (this * 10.pow(5)).ceil() / 10.pow(5)                                                                                                                                                           
  root.down = (this * 10.pow(5)).floor() / 10.pow(5)                                                                                                                                                          
```                                                                                                                                                                                                                                                                                                                                                                                                                          
```                                                                                                                                                                                                           
  {"down": 1.00499, "up": 1.005}                                                                                                                                                                              
```                                                                                                                                                                                                                                                                                                                                                                                                                          
down comes out wrong — 1.005 only has three decimals, so rounding it down at precision 5 should stay 1.005. But 1.005 * 10.pow(5) evaluates to 100499.99999999999 and floor knocks it under the true value. So
the manual version struggles even for plain directed rounding.                                                                                                                                                
                                                                                                                                                                                                               
There's also a subtle mismatch: ceil/floor round every value, not just ties. 1.001.round(2) is 1 with half_up, but the ceil version gives 1.01. And half_even has no one-liner at all:                        
                                                                                                                                                                                                               
```coffeescript                                                                                                                                                                                               
  # Input: {"value": 0.125} — exact tie at 2 places, even candidate is 0.12                                                                                                                                   
  root.pr_half_even = this.value.round(2, "half_even")                 # 0.12                                                                                                                                 
  root.best_diy     = (this.value * 10.pow(2)).round() / 10.pow(2)     # 0.13                                                                                                                                 
```                                                                                                                                                                                                                                                                                                                                                                                                                          
For context: the motivation behind this PR is FX pricing, where half_even (banker's rounding) is the standard convention for exactly this reason — ties distributed evenly instead of always drifting up. It's the one style users genuinely can't write themselves, so I'd love to keep it on the roadmap even if it lands as a follow-up. Happy to drop style from this PR in the meantime and keep just precision 🙂

PR #1012, opened 2026-09-09
fix(aws/kinesis): stop duplicate EFO shard consumers accumulating under backpressure

matus-tomlein:
## Problem

An `aws_kinesis` input using `enhanced_fan_out` accumulates duplicate shard consumers when the pipeline applies sustained backpressure. Memory grows with every surplus consumer until the process is killed, and because each restart resumes from an older checkpoint the cycle repeats.

We hit this in production on a 64-shard stream: consumer goroutines grew from 120 to **287 for 64 shards** in about four minutes, with 96% of the heap held by decoded `SubscribeToShard` events in the AWS SDK's event-stream reader (5–9 MB per subscription).

Three things combine:

1. **A blocked consumer cannot renew its lease.** The periodic checkpoint that doubles as lease renewal lives in the `case <-(*commitCtx).Done():` branch of `efoSubscribeAndProcess`'s outer `select`. `flushBatchedMessage` blocks on `k.msgChan <- *pendingMsg` with only `ctx.Done()` as an alternative, so while the pipeline is full that branch is unreachable and the lease ages past `2 * lease_period`.

2. **The balancer then claims the shard from itself.** `runBalancedShards` marks any claim older than `leasePeriod*2` as unclaimed regardless of owner, and `Claim` uses `ConditionExpression: "ClientID = :old_client_id"`, which succeeds when the client ID is its own. A second consumer starts for a shard already being consumed.

3. **The original consumer never exits.** Kinesis terminates the older `SubscribeToShard` subscription, but that goroutine is still blocked sending to the pipeline, so it never reads the closed stream and never notices. It keeps its decoded batch alive indefinitely.

The polling consumer does not have this problem: it keeps `<-commitCtx.Done()` and `nextFlushChan <- pendingMsg` as sibling cases of one `select`, so a blocked send can never starve lease renewal. The EFO rewrite in #819 moved the send into a helper whose only other case is context cancellation, which is where that property was lost.

## Fix

**Renew the lease while waiting on the pipeline.** `renewLease` is extracted from the commit branch, and every pipeline send now goes through `sendToPipeline`, which selects on the send, the commit timer, and `ctx` — restoring the invariant the polling path already has. A lease genuinely lost while blocked now ends the subscription and yields the shard (`errLeaseLost`) instead of blocking indefinitely.

**Never start a second consumer for a shard we already run.** `kinesisReader` tracks the shards it has live consumers for; `startConsumer` refuses to start a duplicate and the balancer skips such shards before attempting a claim. Consumers deregister in their existing deferred cleanup, so a genuinely dead consumer's shard is still reclaimable.

**Drop the unsent batch when the lease is lost.** Retaining it left the consumer's exit drain blocked on the same full pipeline that cost the lease: the outer loop stops once the state is yielding, and the tail drain in `runEFOConsumer` then waits on `msgChan` until the whole input shuts down, so the deferred cleanup never yields the checkpoint or deregisters the shard. This is safe for at-least-once delivery because `Yield` checkpoints with `recordBatcher.GetSequence()`, which returns `ackedSequence` and only advances when a batch is acked downstream — the yielded position sits before anything unsent, so the next owner reprocesses those records. Both lease-loss exits are covered: the one inside `sendToPipeline` and the one in the subscribe loop's commit branch, which can also lose the lease holding a flushed-but-unsent batch.

This also fixes a third blocking send I found while working through it — the inline `k.msgChan <- *pendingMsg` in the event-handling case, which had the same defect as the two inside `flushBatchedMessage`.

Behaviour when the pipeline is keeping up is unchanged, and the polling path is untouched.

## Testing

`internal/impl/aws/input_kinesis_efo_lease_test.go` covers lease renewal while blocked on the pipeline, giving up when the lease is genuinely lost, the shutdown path, the running-shard registry including its concurrent case, and the balancer guard.

Each test was verified to fail against the previous behaviour: reverting the renewal reports `got 0 renewals`, reverting the balancer guard starts a second consumer, and reverting the lease-loss drop blocks the exit drain until timeout.

- `go test ./internal/impl/aws/...` passes, including with `-race`
- `golangci-lint run -c .golangci.yml ./internal/impl/aws/...` reports 0 issues
- `gofmt -s` clean, `go vet ./...` clean

To make renewal testable without a DynamoDB client, the periodic checkpoint call goes through a `checkpointFn` field that is nil in production. Happy to swap this for an interface on the checkpointer if you'd prefer — it seemed the smaller change.

## Validation

Deployed the fix on the same 64-shard EFO stream that previously OOMed, at higher load than the run that crashed:

| | Before | After |
|---|---|---|
| Consumer goroutines for 64 shards | 120 → 287, climbing | **64, steady** |
| Container restarts | 7 (OOMKilled) | **0** |
| Peak memory | 6.5 GiB → killed | ~2.5 GiB, released after the run |
| Throughput | collapsed to ~0 | ~1,300 events/s sustained |

One consumer per owned shard, with memory that tracks shard count rather than elapsed time.

One operational note for anyone else running EFO at high shard counts: because all shards now renew on schedule rather than only the unblocked ones, steady-state writes to the DynamoDB lease table go up. Ours went from ~3 to ~13 writes/second on a 64-shard stream — that is one write per shard per `commit_period`, as intended, but it is worth checking the table's provisioned capacity before deploying.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ All notable changes to this project will be documented in this file.
 
 - `send_queue_capacity`, `buf_pool_capacity` & `send_loop_count` fields on the `statsd` metrics exporter to tune the underlying `go-statsd` client and avoid packet loss under high throughput @triddell
 
+### Fixed
+
+- `aws_kinesis` input with `enhanced_fan_out` no longer accumulates duplicate shard consumers when the pipeline applies sustained backpressure, which previously grew memory usage until the process was killed @matus-tomlein
+
 ## 1.21.1 - 2026-08-26
 
 ### Fixed
--- a/internal/impl/aws/input_kinesis.go
+++ b/internal/impl/aws/input_kinesis.go
@@ -247,6 +247,18 @@ type kinesisReader struct {
 	cMut    sync.Mutex
 	msgChan chan asyncMessage
 
+	// runningShards tracks the shards this client currently has a consumer
+	// goroutine for, keyed by streamID+shardID. The balancer consults it before
+	// claiming a shard so that a lease which looks stale — most often because
+	// the consumer is blocked waiting for a busy pipeline — cannot cause a
+	// second consumer to be started for a shard we are already consuming.
+	runningMut    sync.Mutex
+	runningShards map[string]struct{}
+
+	// checkpointFn overrides the periodic checkpoint call in tests. Nil in
+	// production, where the real checkpointer is used.
+	checkpointFn func(ctx context.Context, streamID, shardID, sequence string) (bool, error)
+
 	ctx  context.Context
 	done func()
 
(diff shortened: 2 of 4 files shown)


gregfurman:
Hey @matus-tomlein. Apologies for the delay here and thanks for opening! Will give a look sometime this week.

PR #1023, opened 2026-09-13
test: poll for the reloaded state in the config watcher tests

### LINKED ISSUE #1022: TestReaderStream*Watching tests are flaky on CI
Two of the last three `main` test runs failed in `internal/config/watcher_test.go`, on different tests and different lines:

- https://github.com/warpstreamlabs/bento/actions/runs/34759905165 — `TestReaderStreamDirectWatching`, [watcher_test.go:229](https://github.com/warpstreamlabs/bento/blob/main/internal/config/watcher_test.go#L229): `b` still holds `b2` after the `os.Remove`.
- https://github.com/warpstreamlabs/bento/actions/runs/34761799121 — `TestReaderStreamWildcardWatching`, [watcher_test.go:293](https://github.com/warpstreamlabs/bento/blob/main/internal/config/watcher_test.go#L293): `c` was never reloaded in the first phase.

`TestReaderStreamDirectWatching`, `TestReaderStreamWildcardWatching` and `TestReaderStreamDirWatching` each write or remove three files, receive exactly three times from an unbuffered channel fed by the subscription callback, and then assert on all three files. The watcher does not promise one callback per file change: with `newDummyReader`'s 1 ms flush and delay periods a single file can trigger the callback more than once, the count of three is reached early, and the assertion runs before the last file was processed. Both failures above are that shape. `TestReaderResourceWildcardWatching` in the same file already polls with `require.Eventually` instead.

*This issue was created with AI assistance.*

slachiewicz:
Fixes #1022. Replaces the count-three loops in the three stream-watching tests with one `requireStreamLabels` helper that polls for the final state under `confsMut`, matching `TestReaderResourceWildcardWatching`, and drops the unbuffered `changeChan` send from the callbacks. The skipped `TestReaderWatcherRace` is untouched.

Verified: `go test -count=5 -race -run 'TestReaderStream|TestReaderResource|TestReaderFile' ./internal/config/` → ok


slachiewicz:
@jem-davies shoudl fix master branch

gregfurman (approved):
Seems like a reasonable solution to minimise flakiness on these tests 👍 


Looking at the original test cases though, am not the biggest fan of the polling approach since that seems like another flake waiting to happen. While I'd prefer these scenarios be event-driven (e.g using channels within that update closure to signal we're ready to check the collected map), the repeated failures on `main` are very annoying so I'm happy to merge in these changes and look at a rework in a follow-up.

slachiewicz:
Thank you @gregfurman I'll take a look and offer PR with cleanups

PR #1046, opened 2026-09-17
test: stop the slowest unit tests waiting out fixed deadlines

slachiewicz:
Follow-up to #1045. These were the three longest tests in the CGO job's test run (per-package times from run 35264672313): `TestBatcherEarlyTermination` 30 s, `TestGrpcClientOutputUnableToFindMethodErr` and `TestGrpcClientOutputBrokenProtoFile` 20 s each, `TestHugotConfigParsing` 20.8 s.

**grpc_client** (`internal/impl/grpc/output_grpc_client_test.go`): both tests ran a whole stream under a 20 s context and only read the log once the deadline had fired, to observe an error that `Connect` returns directly. They now parse the output config and call `Connect`, asserting on its error, like `TestGrpcClientOutputHealthCheck` beside them. 40 s → 0.05 s. A `fallback` output cannot serve here: the async writer only starts reading transactions once `Connect` has succeeded, so the message never reaches the second output.

**batcher** (`internal/component/output/batcher/batcher_test.go`): the second `WaitForClose` proves that nothing has triggered a stop, so the call must run until its context expires. That holds at 100 ms as well as at 30 s; the assertion now also checks the error is `context.DeadlineExceeded`.

**hugot** (`internal/impl/huggingface/processor.go`): the `path` field documents that it must be a directory when `enable_download` is set, but nothing checked it. `hugot.DownloadModel` contacts the hub before touching the destination, so the "download with file path should fail" case passed only because the download of the nonexistent `foo/bar` repository failed after its retries, about 20 s later, and any environment without network access would pass it for yet another reason. The processor now rejects an existing path that cannot hold the model (a regular file, a symlink to one, or a dangling symlink) before downloading, and the test points at real temporary paths and asserts on those messages. The same up-front check is proposed in hugot itself in knights-analytics/hugot#144. `make docs` produces no change.

Verified: `go test -count=10 -race` on the three packages → pass; `golangci-lint run` → no issues.


internal/impl/huggingface/processor.go
```diff
@@ -133,6 +134,10 @@ func newPipelineProcessor(conf *service.ParsedConfig, mgr *service.Resources) (*
 	}
 
 	if shouldDownload {
+		if st, err := os.Stat(p.modelPath); err == nil && !st.IsDir() {
```
gregfurman: To be honest... I don't think we should actually be supporting model downloading. IMO seems kinda out-of-scope for a stream processing tool.

I believe that @sananguliyev is using this component, so would be curious to hear their thoughts on this.

internal/impl/grpc/output_grpc_client_test.go
```diff
@@ -450,57 +452,89 @@ health_check:
...
+	l.mut.Lock()
+	n, err := l.buf.Write(p)
+	l.mut.Unlock()
 
-output:
-  grpc_client:
-    address: localhost:%v
-    service: helloworld.Greeter
-    method: DoesNotExist
-    reflection: true
```
gregfurman, line 470: I think this logging approach is adding a ton of complexity. Can we not rather change this YAML to include a fallback of some kind? Perhaps one that writes to a test server or something that we can block on.

https://warpstreamlabs.github.io/bento/docs/components/outputs/fallback/

e.g

```yaml
output:
  fallback:
    - grpc_client: {}
    - http_client: 
         url: foo # URL of local test server
```

gregfurman (changes requested):

internal/impl/huggingface/processor.go
slachiewicz: I've touched this part only becouse it was affected by similar validation problem. Test was failing not becouse of failed download but becouse dir was missing.

internal/impl/grpc/output_grpc_client_test.go
slachiewicz: Tried `fallback` first: it can't reach the second output here, because the async writer only starts reading transactions once `Connect` succeeds, and both of these errors come from `Connect`. `fallback: [grpc_client, drop]` sat until the context deadline.

Since `Connect` is what's under test, the tests now call it directly and assert on its error, following `TestGrpcClientOutputHealthCheck` beside them. The log recorder is gone.

*This comment was created with AI assistance.*

internal/impl/huggingface/processor.go
sananguliyev: thanks @gregfurman for mentioning. Yes, I am actively using it, and it's part of my project. One of the most used case is creating knowledge base (RAG).  Here is an example: https://qaynaq.io/playbooks/rag-knowledge-base/

gregfurman: @sananguliyev OK nice thanks for the context!

Am curious if you're using the model downloading functionality though?

internal/impl/grpc/output_grpc_client_test.go
gregfurman: OK this is looking much nicer 👍

internal/impl/huggingface/processor.go
gregfurman: @slachiewicz Will the `DownloadModel` call not fail if the path is not a directory? 

```go
hugot.DownloadModel(context.Background(), modelRepo, p.modelPath, opts)
```

slachiewicz: It does, but only after the download: `DownloadModel` resolves the repo (5 retries × 5 s, which is where the old test's 20 s went), downloads every file into the HF cache, and only then copies them to `<path>/<repo>/…`, which fails with `mkdir <path>: not a directory`. I've proposed the check upstream in knights-analytics/hugot#144; until bento is on a release with it, this one rejects the config before any of that, with the constraint the `path` docs already state, and lets the test run without network access. It now also catches a dangling symlink, which `os.Stat` alone reports as non-existent.

*This comment was created with AI assistance.*

PR #1049, opened 2026-09-18
test(kafka_franz): Fix topic re-creation flow

gregfurman:
## Context

Fixes both `TestIntegrationFranzInputDetectUnknownTopicError` and `TestIntegrationFranzInputReconnectUnknownTopicError` integration tests.

## Changes

- `TestIntegrationFranzInputDetectUnknownTopicError` now sets the cluster's `KAFKA_OFFSETS_TOPIC_REPLICATION_FACTOR=1`. Where previously. we were creating a single node cluster with a (default) replication factor of 3 -- hence writes failed and the required unknown topic behaviour wasn't triggering.
- `TestIntegrationFranzInputReconnectUnknownTopicError` also has the above extra config set, in addition to an entirely separate `kgo` client being used in the final step of the `Create topic -> Delete topic -> Re-create topic` flow. This was failing as the recreation step used the _original client_ which cached information on the topic we were trying to re-create, causing the request to fail.


internal/impl/kafka/input_kafka_franz_test.go
```diff
@@ -309,21 +311,21 @@ func TestIntegrationFranzInputReconnectUnknownTopicError(t *testing.T) {
...
+			"KAFKA_OFFSETS_TOPIC_REPLICATION_FACTOR=1",
+			"KAFKA_GROUP_INITIAL_REBALANCE_DELAY_MS=0",
 		}),
 		dockertest.WithoutReuse(),
 	)
 
 	brokerAddress := "localhost:" + kafkaPortStr
-	cl, err := kgo.NewClient(kgo.SeedBrokers(brokerAddress))
-	require.NoError(t, err)
-	defer cl.Close()
```
gregfurman, line 319: This isn't actually used in the test...?

PR #1054, opened 2026-09-22
fix(gcp): correct byte_size lint threshold for BigQuery Storage Write API output

SJ1397:
## Summary
- The `batching.byte_size` lint rule for the `gcp_bigquery_write_api` output triggered at 1,000,000 bytes (1 MB) while the lint message stated the limit was 10 MB, which is BigQuery's documented `AppendRows` request size limit.
- This mismatch caused the lint to incorrectly flag valid batching configurations that were well within the actual API limit (e.g. a 5 MB `byte_size` setting), needlessly forcing smaller batches and more RPC round-trips.
- Updates the threshold to 10,000,000 bytes so the lint rule matches the limit it claims to enforce.

## Test plan
- [x] `go build ./internal/impl/gcp/...` passes
- [x] Verified via a standalone lint test against `schema.NewStreamConfigLinter()`: configs with `byte_size` at 5 MB and just under 10 MB no longer trigger the lint (previously they incorrectly did), while configs at or above 10 MB still correctly trigger it
- [x] `TestGCPBigQueryStorageOutputConnectOk`, `TestGCPBigQueryStorageOutputWriteOK`, `TestGCPBigQueryStorageOutputProtobufFormat`, `TestGCPBigQueryStorageOutputProtobufUnmarshalError` (Docker-backed integration tests for this output) all pass


internal/impl/gcp/output_bigquery_storage.go
```diff
@@ -66,7 +66,7 @@ sets the type of stream this write client is managing.`).Default(string(managedw
 			"json":     "Messages are in JSON format (default)",
 			"protobuf": "Messages are in protobuf format",
 		}).Description("Format of incoming messages").Default("json")).
-		Field(service.NewBatchPolicyField("batching").Advanced().LintRule(`root = if this.byte_size >= 1000000 { "the amount of bytes in a batch cannot exceed 10 MB" }`)).
+		Field(service.NewBatchPolicyField("batching").Advanced().LintRule(`root = if this.byte_size >= 10000000 { "the amount of bytes in a batch cannot exceed 10 MB" }`)).
```
gregfurman, line 69: praise: Nice catch! Makes me think that we should definitely be supporting humanised byte format in bloblang -- that way specifying the literal `10MiB` would've resolved correctly as opposed to relying on correctly counting the zeros in 10 million 😅

gregfurman (approved):
LGTM! Could you add a small line to the `CHANGELOG.md` under the `Fixed` header for this?

SJ1397:
Done — added a CHANGELOG.md entry under Fixed.

gregfurman:
> Done — added a CHANGELOG.md entry under Fixed.

Brilliant. Will merge when CI is 🟢

internal/impl/gcp/output_bigquery_storage.go
SJ1397: Agreed — definitely worth exploring. It'd help keep lint rules like this one readable and less error-prone (no more manually counting zeros). Given go-humanize's ParseBytes is already used elsewhere in the codebase (kafka_franz's broker_write_max_bytes, etc.), a parse_bytes()-style Bloblang function might be a lighter-weight first step than a native literal in the grammar. Happy to open a follow-up issue for it if that's useful.

gregfurman: Agree that using the pre-existing go-humanize package would be easiest here. Having something in the parsing (or perhaps string manipulation) section seems like a good fit

https://warpstreamlabs.github.io/bento/docs/guides/bloblang/methods#parsing

If you could open an issue for this that'd be much appreciated 🙏

PR #1058, opened 2026-09-23
fix(mongodb/cache): Error no documents & non-bytes return type

gregfurman:
## Changes

- Ensures that `Exists` does not return an error when the error response is `ErrNoDocuments` -- flagged by the integration test failure in https://github.com/warpstreamlabs/bento/actions/runs/35805699301/job/107005919433#step:6:131
- If a `cache.Get` returned a non-string value, the `value.StringValue` would panic. Instead, add a check for `bson.TypeString` or `TypeBinary` and return an error if neither.


jem-davies (approved):
good spot

